Issue #614, opened 2014-02-15
Run go build on a directory instead of specific file


jkbbwr:
Go build is able to run against a directory so the plugin should be able to do this as well.

The interface should be changed to accommodate for this.
## Old description

With two files like so

```
package main

import (
    "fmt"
)

func main() {
    fmt.Printf("Hello world! %d", test())
}
```

And 

```
package main

func test() int {
        return 2+2;
}
```

I get the following errors 

```
Could't build project C:\projects\CronOverlord\bin/CronOverlord.exe
C:/Go/bin/go.exe build  -o C:\projects\CronOverlord\bin/CronOverlord.exe     C:/projects/CronOverlord/src/github.com/jkbbwr/cronoverlord/CronOverlord.go

Could't build project C:\projects\CronOverlord\bin/CronOverlord.exe
# command-line-arguments
   C:\projects\CronOverlord\src\github.com\jkbbwr\cronoverlord\CronOverlord.go:8: undefined: test
```

My output for `go env`

```
set GOARCH=amd64
set GOBIN=
set GOCHAR=6
set GOEXE=.exe
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOOS=windows
set GOPATH=C:\projects\go;C:\projects\CronOverlord
set GORACE=
set GOROOT=C:\Go
set GOTOOLDIR=C:\Go\pkg\tool\windows_amd64
set CC=gcc
set GOGCCFLAGS=-g -O2 -m64 -mthreads
set CGO_ENABLED=1
```

My project is structured as so

```
src\github.com\jkbbwr\cronoverlord\
    - CronOverlord.go
    - test.go
```

In the command line I can build it by going into the directory and running `go build`
I can even run it by doing `go run CronOverlord.go test.go`

I currently have `build before run` turned on

It tries to build, fails because it can't find `test.go` And then tries to run the single file `CronOverlord.go` which fails for the same reason. 

Is there a fix? Because this makes using this plugin impossible

mageddo:
The follow command works great for me 

    $ go build -o build/hello hello.go 
	.
	├── build
	│   └── hello
	└── hello.go

Issue #555, opened 2014-01-20
can not run multi file project


tombrus:
Can not run a func in another file.

When I have one go file I can run fine.
When I have two go files (main using a func in the other), I get an "undefined:" error.

repro:
- make a new go project
- create one file:

``` go
package main
import "fmt"
func main() {
    fmt.Printf("Hello world!")
    xxx()
}
```
- create another file:

``` go
package main
import "fmt"
func xxx() {
    fmt.Printf("Hello world again!")
}
```
- run the first file and get:

```
# command-line-arguments
src/gotry.go:7: undefined: xxx

Process finished with exit code 2
```

My 'internals':

```
Project dir -> /Users/tom/projects/gotry
GO_GOROOT_PATH -> /usr/local/go
GO_BIN_PATH -> /usr/local/go/bin/go
GO_GOPATH_PATH -> /Users/tom/gocode
TARGET_OS -> Darwin
TARGET_ARCH -> _amd64
VERSION_MAJOR -> go1.1.2 darwin/amd64
VERSION_MINOR -> 
Extended Go Env -> 
SHELL=/bin/bash
TMPDIR=/var/folders/96/rgpc19kj6tx1q67p8mqbr7gr0000gn/T/
__CF_USER_TEXT_ENCODING=0x1F5:0:2
PATH=/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/go/bin:/Users/tom/gocode/bin:/Users/tom/projects/gotry/bin
GOROOT=/usr/local/go
DISPLAY=/tmp/launch-KuXA1u/org.macosforge.xquartz:0
GOPATH=/Users/tom/gocode:/Users/tom/projects/gotry
USER=tom
COM_GOOGLE_CHROME_FRAMEWORK_SERVICE_PROCESS/USERS/TOM/LIBRARY/APPLICATION_SUPPORT/GOOGLE/CHROME_SOCKET=/tmp/launch-FBuB9z/ServiceProcessSocket
HOME=/Users/tom
LOGNAME=tom
Apple_PubSub_Socket_Render=/tmp/launch-SVekGf/Render
__CHECKFIX1436934=1
SSH_AUTH_SOCK=/tmp/launch-hUIPDS/Listeners
```

```
Project dir -> /Users/tom/projects/gotry
/usr/local/go/bin/go env
GOARCH="amd64"
GOBIN=""
GOCHAR="6"
GOEXE=""
[...]

PR #666, opened 2014-03-26
Build package instead of single file

d3xter:
Creating source files under src/ is not allowed with go build,
so in new projects a "main"-package is created now,
and the new go-file is placed in there.

When "Build before Run is enabled", the package which will be built,
is determined from the Script-File chosen in the RunConfiguration.
This package will be built by "go build".

If all source-files are placed in the project-root, compilation still works.
#614 #555

### DIFF
--- a/src/ro/redeul/google/go/runner/GoRunProfileState.java
+++ b/src/ro/redeul/google/go/runner/GoRunProfileState.java
@@ -92,7 +92,14 @@ public class GoRunProfileState extends CommandLineState {
 
         try {
             String[] goEnv = GoSdkUtil.convertEnvMapToArray(sysEnv);
-            String[] command = GoSdkUtil.computeGoBuildCommand(goExecName, m_configuration.runBuilderArguments, execName, m_configuration.scriptName);
+
+            String pkg = GoSdkUtil.getPackageOfFile(projectDir, m_configuration.scriptName);
+
+            if(pkg == null) {
+                throw new CantRunException("Script is not inside the project root");
+            }
+
+            String[] command = GoSdkUtil.computeGoBuildCommand(goExecName, m_configuration.runBuilderArguments, execName, pkg);
 
             Runtime rt = Runtime.getRuntime();
             Process proc = rt.exec(command, goEnv, new File(projectDir));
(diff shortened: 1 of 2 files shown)


dominicwilliams:
Anyone got an update on state of play here?

I'm using the v1.0.0-alpha0 branch (for lovely refactoring). But...

Cannot find convenient way to run a project with more than 1 source file in same folder. Am I missing something?

Seems important as many projects come packaged in this convenient format. For example:
https://github.com/go-qml/qml/tree/v1/examples/gopher

Having to run "go build" from command line each time is just silly

dominicwilliams:
I'm trying to find a way to conveniently build/run https://github.com/go-qml/qml/tree/v1/examples/gopher

The problem is the application consists of two files in the same folder: gopher.go and wavefront.go

When I run, the plugin *only* builds gopher.go, resulting in missing reference. Would be nice if was way to configure to build using e.g. `go build", which works fine

`/usr/local/go/bin/go build -o "/private/var/folders/jz/ymsf2ky56yjfhb4ygccqd_rm0000gn/T/Gopher Application0go" /Users/dwilliams/go/src/scratch/qt-examples/gopher/gopher.go
/Users/dwilliams/go/src/scratch/qt-examples/gopher/gopher.go:53: undefined: Object`

dlsniper:
Ah, sorry, I've misread the package in the example. That functionality is not working at the moment. Please see https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1520 and the related tickets from comments to understand the issue. Thank you.

DipenModi:
How to set this on visual studio code editor - Run | Edit Configurations | + | Go Application

dlsniper:
@DipenModi I'm not sure what you mean.
This is not meant to be used in Visual Studio Code but in IntelliJ IDEA and other JetBrains IDEs.
You can see how to configure it here: https://youtu.be/a_mFM3YOTIM

dvic:
Maybe worth while mentioning: for me it works when I right click on the directory and select 'Run'. This makes an application run configuration for the package: it builds and runs the one main function. Works for me 👍

PR #909, opened 2014-10-26
Send feedback on crashes via Github API

dlsniper:
A crash report looks like this: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/921


bennetelli:
how did you solve the problem with authentication of the users reporting issues?
Looks like you created a personal access token and use it for every user? How do you pass the respective users access token in the plugin?

dlsniper:
> how did you solve the problem with authentication of the users reporting issues?
> Looks like you created a personal access token and use it for every user? 

That is correct, I've created a new github user to be a bot and used a token for that.

> How do you pass the respective users access token in the plugin?

I haven't. The plugin was posting to an AppEngine running applications which handled the posting to Github on behalf of the user.

bennetelli:
@dlsniper ah. Thank you :-)

PR #1055, opened 2014-11-25
add two failing test cases from go sdk sync

bronze1man:


dlsniper:
I've ported one test case to an issue on the 1.0.0 branch and I'll close this for now as 0.9.16 seems is unlikely to happen before 1.0.0

PR #1093, opened 2014-11-29
[WIP] Fix project creation (fixes #675)

### LINKED ISSUE #675: Recheck project creation / import logic
The logic for creating a new project should be checked again as the plugin now creates a 'main' folder inside 'src' when creating a new project and that's not needed.

Also the import logic should be checked to make sure everything is ok.

dlsniper:
This one replaces #1004


dlsniper:
while I'd like to have something similar to the 1.0 or 1.maybe, I'll close this for now while we wait for a better solution.

PR #1115, opened 2014-12-06
Add go1.4 support

dlsniper:
This adds support for Go 1.4+ SDKs.

I've also corrected the default Linux paths for the SDK (as per [Go documentation](http://golang.org/doc/install#tarball)) and upgraded the suggestions for the SDK default directory.

After the change I've noticed a severe drop in performance due to ``` getPathsToLookup() ``` method being called a great deal of times / file / element ``` src/com/goide/psi/impl/imports/GoImportReferenceHelper.java ```. Example to be used: ``` src/fmt/print.go ``` from Go 1.4 SDK.


dlsniper:
This was implemented by someone else so I"m closing this.

PR #1172, opened 2015-01-13
Fix NoClassDefFoundError: com/goide/jps/model/JpsGoSdkType

### LINKED ISSUE #1169: Unable to configure a go SDK in Idea CE 14
Hi,
I cannot configure a new go sdk in Idea CE 14: when I click on "new" and then select the good directory (actually, the file browser is already on the good one), then validate..nothing happens!

Versions:
java version "1.6.0_65"
Java(TM) SE Runtime Environment (build 1.6.0_65-b14-466.1-11M4716)
Java HotSpot(TM) 64-Bit Server VM (build 20.65-b04-466.1, mixed mode)
golang: 1.4
Idea CE: 14
Os: Mac Os Yosemite

Sorien:
~~fixes #1169 for 1.0.0-alpha~~


dlsniper:
I can't reproduce the issue using f2af14cbd3b8b078704be28614beb54c59eaa7af

Sorien:
without this i can't even try to select sdk dir it throw errors like java.lang.NoClassDefFoundError: com/goide/jps/model/JpsGoSdkType

its something with java-deps.xml and line <compileServer.plugin classpath="go-jps-plugin.jar"/>
i tryed to change it to jps-plugin.jar but it was not working then i renamed jps-plugin module to go-jsp-plugin recompiled and not errors so far ... but there are still some problems wit sdk selection :)

zolotov:
Hi Sorien,

Thank you for investigation. The caf7ee7 introduce the wrong name for jps artifact. If you don't mind, could you instead of renaming module, just rename artifact in 'Project Structure | Artifacts'?

Sorien:
i tried to rename Artifact Go.zip \lib\jps\jps-plugin.jar -> go-jps-plugin.jar but its not working for me, same errors. .. its not something critical I just wanted to that plugin 

PS: when i run or debug plugin there is no such error, i see that error only when i open new idea instance with registered plugin and i try to select Go SDK.

zolotov:
> i tried to rename Artifact Go.zip \lib\jps\jps-plugin.jar -> go-jps-plugin.jar but its not working for me, same errors

That's not exactly that I meant. I've pushed some changes, could you please check whether it works now?

> PS: when i run or debug plugin there is no such error, i see that error only when i open new idea instance with registered plugin and i try to select Go SDK.

I understand it, that's why we didn't experience such problems.

PR #1174, opened 2015-01-14
Introduce isVariadic() in GoParamDefinition, fix type in index expression for variadic parameter

jhsx:


ignatov:
Sorry, I've just pushed several commit. Please rebase yours.

ignatov:
And please provide a highlighting test that will show the fix. If you don't know how to deal with tests just attach a code sample with the original issue and I'll help.

PR #1175, opened 2015-01-14
Introduce isVariadic() in GoParamDefinition, fix type in index expression for variadic parameter

jhsx:
Hi @ignatov , just had time now, i end up by closing the other pull request accidentally, i rebase and i am sending the pull request again, but i am not sure about how to make the test for this.

The case is following.
```go
func Add(users ...*User){
      users[i].Name() // Name was unresolved, now we resolve "users" reference
      // and check if this is a parameter and if is Variadic, then if is true
      // we return the type of the parameter which in this case is *User
}
```

### DIFF
--- a/gen/com/goide/psi/GoParamDefinition.java
+++ b/gen/com/goide/psi/GoParamDefinition.java
@@ -12,4 +12,7 @@ public interface GoParamDefinition extends GoNamedElement, StubBasedPsiElement<G
   @NotNull
   PsiElement getIdentifier();
 
+  @NotNull
+  boolean isVariadic();
+
 }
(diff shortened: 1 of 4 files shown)


ignatov:
Thanks for contribution. Your PR's been merged manually. Please see the real fix for your problem and check out how to create a tests for such kind of things: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/commit/a4bc9b2e2c156c498eeeaa7a51f495b30187a426.

ignatov:
In case of any questions about how to contribute please don't hesitate to ask me.

jhsx:
Ok thanks :)

PR #1179, opened 2015-01-15
Fixing the javac2 class location to the default Idea path

letsjustfixit:
It's nothing really just a path change but can be annoying :)


mtoader:
Please rebase to latest master.

dlsniper:
Also, if you want to build the plugin from the command line, look at how we
do it for TravisCI
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/coverity.sh

letsjustfixit:
@mtoader I'm sorry I'll do so :)
@dlsniper I'm building via the intellij ide with the ant plugin.

dlsniper:
There's no need for that. Just use Build -> Prepare plugin for deployment (last option in the Build menu). And that's it.

dlsniper:
I'll close this one as 1.0.0 doesn't present this issue and it's overriding the 0.9.x branch

PR #1190, opened 2015-01-24
Fix #1189, Call inspection Signatures with unnamed parameter

### LINKED ISSUE #1189: [1.0.0] GoFunctionCallInspection incorrectly flagging error on unnamed parameters
The following code reproduces the error
```go

type CollectionLoaderInterface interface {
    MangoCollection(string) *mgo.Collection
}

func Collection(parent interface{},collection string) *mgo.Collection{
     switch parent:= parent.(type) {
          case CollectionLoaderInterface:
          return parent.MongoCollection(collection) // Inspection error: too many arguments...
     }
     return nil
}

```

jhsx:


ignatov:
Merged manually, thanks for contributing.
Next time please don't hesitate to squash your commits all together.

PR #1196, opened 2015-01-29
replaced /null/nil/ in the go livetemplate

nerdalert:
-quick fix on replacing null with nil in the go.xml template

great work thanks for all the effort, the alpha branch is great.

Signed-off-by: Brent Salisbury <brent@socketplane.io>


ignatov:
Thanks for the contributing.

PR #1253, opened 2015-02-04
Added goimports to code menu. 

zippelmann:
Hello,

goimports is now in code menu. 

It assumes that goimports is already installed. A shortcut is missing. I tested goimports on macos 10.10.2 with go 1.4.1 and intellij 14.0.3.

### DIFF
--- a/jps-plugin/src/com/goide/jps/model/JpsGoSdkType.java
+++ b/jps-plugin/src/com/goide/jps/model/JpsGoSdkType.java
@@ -43,4 +43,9 @@ public class JpsGoSdkType extends JpsSdkType<JpsDummyElement> implements JpsElem
   private static File getExecutable(@NotNull String path, @NotNull String command) {
     return new File(path, getBinaryFileNameForPath(command));
   }
+
+  @NotNull
+  public static File getExecutableFileFromEnv(String command) {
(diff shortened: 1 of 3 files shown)


ignatov:
Thanks for contributing. Some common code's been extracted. Please check that all functionality with goimports are work correctly.

jps-plugin/src/com/goide/jps/model/JpsGoSdkType.java
```diff
@@ -43,4 +43,9 @@ public JpsDummyElement createDefaultProperties() {
   private static File getExecutable(@NotNull String path, @NotNull String command) {
     return new File(path, getBinaryFileNameForPath(command));
   }
+
+  @NotNull
+  public static File getExecutableFileFromEnv(String command) {
```
ignatov: It's not a good idea to annotate with not null the method that can return a null value. Please keep this in mind next time.

zippelmann: You are right. It would crash, if goimports is not install. I could add a notification if the method returns null.

ignatov: I've already add the warning about it. Please find it out.

PR #1256, opened 2015-02-04
Remove removed method use from debugger (fixes #1245)

### LINKED ISSUE #1245: [1.0] Plugin doesn't compile with EAP 14.1 (build 140.2110.5)
Because ```setAutoInitBreakpoints()``` was deprecated in 13.1 and bound to be removed in 14... I guess the time has came for that. Quick refference for it: ```com/goide/debugger/ideagdb/run/GdbRunner.java:76```

dlsniper:
Not sure if it's the best way but at the same time, the debugger is not there yet anyway.

### DIFF
--- a/src/com/goide/debugger/ideagdb/run/GdbRunner.java
+++ b/src/com/goide/debugger/ideagdb/run/GdbRunner.java
@@ -73,7 +73,6 @@ public class GdbRunner extends DefaultProgramRunner {
       @NotNull
       @Override
       public XDebugProcess start(@NotNull XDebugSession session) throws ExecutionException {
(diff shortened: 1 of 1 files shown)


develar:
Please see https://devnet.jetbrains.com/message/5509464#5509464

 XDebugProcess::checkCanInitBreakpoints must be implemented if setAutoInitBreakpoints was useful before.

dlsniper:
Agreed. However gdb integration never really worked fully as the support for gdb waa dropped by the Go team. 

I'm trying to help a different debugger, https://github.com/derekparker/delve to get IDE support and that will work differently. 

Imho, the whole debugging part could / should be dropped for now.

PR #1271, opened 2015-02-07
Improve default path detection for go SDK

dlsniper:
This allows me to have the Browse for sdk folder dialog open directly in the SDK dir.

I've followed the instructions from here: [http://golang.org/doc/install#tarball](http://golang.org/doc/install#tarball)

Also, the tests now finally pass on my computer due to this.


dlsniper:
As an open question, how can I move the ```GoSdkType::suggestHomePath()``` to ```Jps*``` so that this can be used in the tests as well?

Thank you.

PR #1285, opened 2015-02-10
[1.0] Fix parameter info help (fixes #1214)

### LINKED ISSUE #1214: [1.0] Multiple parameters with same type highlighted incorrectly in parameter info
Using 14add20fb5041b75b49c20f9e95fc76766f74db5

```go
package main

func demo(a, b, c string, d bool) (string, int, error) {
	return a, 0, nil
}

func main() {
	demo("a", "b", "c", 1)
}

```

snapshot99

horkhe:
The parameter info help has not worked correctly in cases when parameters where declared like this `a, b, c int` or this `more ...int`. Those cases were fixed and a test case suite was added to confirm that.

### DIFF
--- a/src/com/goide/editor/GoParameterInfoHandler.java
+++ b/src/com/goide/editor/GoParameterInfoHandler.java
@@ -18,6 +18,7 @@ package com.goide.editor;
 
 import com.goide.GoTypes;
 import com.goide.psi.*;
+import com.google.common.collect.Lists;
 import com.intellij.codeInsight.lookup.LookupElement;
 import com.intellij.lang.parameterInfo.*;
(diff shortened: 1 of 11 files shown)


dlsniper:
@horkhe am I correct to assume it fixes #1214?
Also, thank you for your contribution!

horkhe:
@dlsniper it does fix #1214. [This](https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1285/files#diff-18aeb8b5fe620ea6c29d28126ccfde31R44) test case covers that scenario.

dlsniper:
Awesome!

ignatov:
Thanks, please next time prefer ContainerUtil to guava library.

PR #1301, opened 2015-02-12
Added 'gt' live template to add a go unit test

dtanner:


dlsniper:
@dtanner sorry for the late response. If you want this to be in the 1.0 releases, please submit a PR and target the v1.0.0-alpha0 branch. Else it won't make it in the 0.9.16 as I'm not sure it's going to be one before 1.0. Thank you!

PR #1307, opened 2015-02-14
Fix running goimports when on Windows.

jakecoffman:
Running the goimports tool was failing on Windows because it searches for "goimports" and not "goimports.exe."


ignatov:
Thanks

Issue #1257, opened 2015-02-04
[1.0] Type nil has wrong coloring


dmitry-vovk:
Type ```nil``` has the same color as global variable, but actually it is a reserved word, like ```chan``` or ```func``` or ```var```.

Also spotted that ```true``` and ```false``` are classified as constants now, not sure if that's intended. Perhaps it is better to treat them as reserved words too?

IDE: IU 14.0.3
Plugin: v1.0.0-alpha0 (from here: http://kokizzu.blogspot.com/2015/02/how-to-build-latest-go-idea-plugin.html)

PR #1315, opened 2015-02-15
Add highlight support for different visibility levels per type (fixes #1241, #1257, #1319)

### LINKED ISSUE #1241: [1.0] Coloring for various visibility levels
Hi,

Based on the discussion from #1211, I've created this ticket so we can keep track of the various coloring progress:

- package.Var1 -> Var1 is a package Exported variable
- package.var2 -> var2 is a package Private variable
- struct.Var3 -> Var3 is a struct Exported variable
- struct.var4 -> var4 is a struct Private variable
- func(){ var var5, var Var6 } -> var5 and Var6 are Local variables
- package.Const1 -> Const1 is a package Exported constant
- package.const2 -> const2 is a package Private constant
- func(){ const const3, const const4 } -> const4 and Const5 are Local constants
- package.Func1() -> Func1() is a package Exported function
- package.func2() -> func2() is a package Private function
- interface.Func3() -> Func3() is a struct Exported function
- interface.func4() -> func4() is a struct Private function.
-  variable redefined by usage in if statement (which makes it local to the if statement)

As a code sample, we could use this:

```go
package demo

type (
	PublicInterface interface {
		PublicFunc() int
		privateFunc() int
	}

	private interface {
		PublicFunc() int
		privateFunc() int
	}

	PublicStruct struct {
		PublicField  int
		privateField int
	}

	privateStruct struct {
		PublicField  int
		privateField int
	}
)

const (
	PublicConst  = 1
	privateConst = 2
)

var (
	PublicVar  = 1
	privateVar = 2
)

func PublicFunc() int {
	const LocalConst = 1
	localVar := PublicVar
	return localVar
}

func privateFunc() int {
	const localConst = 2
	LocalVar := privateVar
	return LocalVar
}

func (ps PublicStruct) PublicFunc() int {
	return ps.privateField
}

func (ps PublicStruct) privateFunc() {
	return ps.PublicField
}

func (ps privateStruct) PublicFunc() int {
	return ps.privateField
}

func (ps privateStruct) privateFunc() {
	return ps.PublicField
}

func variableFunc() {
	var demo1 int
	demo1 = 3

	if demo1 := privateFunc(); demo1 != 3 {
		_ = demo1
		return
	}

	_ = demo1
}
```

dlsniper:
This adds support for different highlight options per level and type as described in #1241 

- [x] package.Var1 -> Var1 is a package Exported variable
- [x] package.var2 -> var2 is a package Private variable
- [x] struct.Mem1 -> Mem1 is a struct Exported member
- [x] struct.mem2 -> mem2 is a struct Private member
- [x] func(){ var var3, var Var4 } -> var3 and Var4 are Local variables
- [x] package.Const1 -> Const1 is a package Exported constant
- [x] package.const2 -> const2 is a package Private constant
- [x] func(){ const const3, const const4 } -> const4 and Const5 are Local constants
- [x] package.Func1() -> Func1() is a package Exported function
- [x] package.func2() -> func2() is a package Private function
- [x] interface.Func3() -> Func3() is a struct Exported function
- [x] interface.func4() -> func4() is a struct Private function.
- [x] variable redefined by usage in if statement (which makes it local to the if statement)

### DIFF
--- a/src/com/goide/GoAnnotator.java
+++ b/src/com/goide/GoAnnotator.java
@@ -25,6 +25,7 @@ import com.intellij.lang.annotation.Annotator;
 import com.intellij.openapi.editor.colors.EditorColorsManager;
 import com.intellij.openapi.editor.colors.TextAttributesKey;
 import com.intellij.openapi.editor.markup.TextAttributes;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.util.containers.ContainerUtil;
@@ -48,25 +49,29 @@ public class GoAnnotator implements Annotator {
       highlightRefIfNeeded((GoTypeReferenceExpression)o, resolve, holder);
     }
     else if (o instanceof GoTypeSpec) {
-      setHighlighting(((GoTypeSpec)o).getIdentifier(), holder, GoSyntaxHighlightingColors.TYPE_SPECIFICATION);
+      setHighlighting(((GoTypeSpec)o).getIdentifier(), holder, getColor((GoTypeSpec)o));
     }
     else if (o instanceof GoConstDefinition) {
-      setHighlighting(((GoConstDefinition)o).getIdentifier(), holder, GoSyntaxHighlightingColors.CONSTANT);
+      setHighlighting(o, holder, getColor((GoConstDefinition)o));
     }
-    else if (o instanceof GoVarDefinition && isGlobal((GoVarDefinition)o)) {
-      setHighlighting(((GoVarDefinition)o).getIdentifier(), holder, GoSyntaxHighlightingColors.GLOBAL_VARIABLE);
+    else if (o instanceof GoVarDefinition) {
+      setHighlighting(o, holder, getColor((GoVarDefinition)o));
+    }
+    else if (o instanceof GoFieldDefinition) {
+      setHighlighting(o, holder, getColor((GoFieldDefinition)o));
+    }
+    else if (o instanceof GoParamDefinition) {
+      setHighlighting(o, holder, getColor((GoParamDefinition)o));
+    }
+    else if (o instanceof GoReceiver) {
+      setHighlighting(o, holder, getColor((GoReceiver)o));
     }
     else if (o instanceof GoNamedSignatureOwner) {
       PsiElement identifier = ((GoNamedSignatureOwner)o).getIdentifier();
       if (identifier != null) {
-        setHighlighting(identifier, holder, GoSyntaxHighlightingColors.FUNCTION_DECLARATION);
+        setHighlighting(identifier, holder, getColor((GoNamedSignatureOwner)o));
       }
     }
-  }
-
-  private static boolean isGlobal(@NotNull GoVarDefinition o) {
-    GoVarDeclaration declaration = PsiTreeUtil.getParentOfType(o, GoVarDeclaration.class);
-    return declaration != null && declaration.getParent() instanceof GoFile;
   }
 
   private static void highlightRefIfNeeded(@NotNull GoReferenceExpressionBase o,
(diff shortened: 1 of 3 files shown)


dlsniper:
I'll continue adding support for functions and structs.

src/com/goide/GoAnnotator.java
```diff
@@ -64,11 +64,21 @@ else if (o instanceof GoNamedSignatureOwner) {
...
     return declaration != null && declaration.getParent() instanceof GoFile;
   }
 
+  private static boolean isStructContained(@NotNull GoVarDefinition o) {
+    GoVarDeclaration declaration = PsiTreeUtil.getParentOfType(o, GoVarDeclaration.class);
+    return declaration != null && declaration.getParent() instanceof GoStructType;
+  }
+
+  private static boolean isPublic(@NotNull GoVarDefinition o) {
+    String value = o.getNode().getText();
```
ignatov: `o.isPublic()`?

src/com/goide/GoAnnotator.java
```diff
@@ -81,9 +91,34 @@ private static void highlightRefIfNeeded(@NotNull GoReferenceExpressionBase o,
...
+    }
+  }
+
+  private static void highlightVariable(AnnotationHolder holder, GoVarDefinition o) {
+    boolean isPackageWide = isPackageWide(o);
+    boolean isStructContained = isStructContained(o);
+    boolean isPublic = isPublic(o);
+    TextAttributesKey color;
+
+    if (isPackageWide) {
```
ignatov: Please extract a new method getColor() and reduce amount of if statements.

src/com/goide/GoAnnotator.java
```diff
@@ -64,11 +64,21 @@ else if (o instanceof GoNamedSignatureOwner) {
     }
   }
 
-  private static boolean isGlobal(@NotNull GoVarDefinition o) {
+  private static boolean isPackageWide(@NotNull GoVarDefinition o) {
     GoVarDeclaration declaration = PsiTreeUtil.getParentOfType(o, GoVarDeclaration.class);
     return declaration != null && declaration.getParent() instanceof GoFile;
   }
 
+  private static boolean isStructContained(@NotNull GoVarDefinition o) {
```
ignatov: Are you sure that it works correctlly?

dlsniper:
I've added some changes to the PR.

I'm still not sure what you meant with the previous comment about: ```Please extract a new method getColor() and reduce amount of if statements.``` so please help me out a bit on that one.

Do you think it makes sense to have ```isPackageWide()``` moved to ```GoNamedElement``` so that it's present for the others as well (I suspect it could be useful to some other parts of the code as well, maybe in the autocomplete?)

Another thing that I've realized is that we need to change the PSI structure in order to have the struct / interface changed to support the definition / implementation way as the variables / consts / struct members have right now. If you could please help me on either of them, I could do the other one. 

And as a preview, see below. Obviously there's a bit more work to do in order to support colors in the parts there they are actually used, not just defined.

snapshot1

Thank you.

zolotov:
I believe that Sergey meant that invoсations of 'highlightSomething' method could be replaced with setHighlighting(psi, holder, getColorOfSomething()); and isPackageWide and isPublic variables could be inlined.

In this case you'll get pretty simple getColor methods with trivial and short if-return-else-return-statements.

zolotov:
> Do you think it makes sense to have isPackageWide() moved to GoNamedElement so that it's present for the others as well (I suspect it could be useful to some other parts of the code as well, maybe in the autocomplete?)

Don't know, maybe when we'll need the same checking in other places. For now at least we should get rid of method duplications, isPackageWide methods are really the same, parameter type restriction doesn't look good, GoElement or PsiElement would be enough and argument of getParentOfType could be a parameter of isPackageWide method.

zolotov:
> Another thing that I've realized is that we need to change the PSI structure in order to have the struct / interface changed to support the definition / implementation way as the variables / consts / struct members have right now. If you could please help me on either of them, I could do the other one.

Sure, how do you want to change it?

ignatov:
Please squash all your commits.

dlsniper:
I've rebased the PR but it seems the signature of ```getGoType()``` was changed in cd8a3aa7d9382c8c86f2e5a8d99e328a4ea59d2d which breaks the things for now. I have to look into what's changed there.

dlsniper:
@ignatov it's done. Please review :)

PR #1322, opened 2015-02-16
Support for language injections fixed #1321

### LINKED ISSUE #1321: Add support for language injections

stuartcarnie:


ignatov:
Thanks for the contributing.
We need to implement an appropriate escaper for Go language which will respect quoting rules.
At the moment is's a red code:
```go
package main
import "fmt"
func main() {
    fmt.Println("package a; import \"fmt\"; func main() { fmt.Println()}") // inject Go here and get some red code
    fmt.Println(`package a; import "fmt"; func main() { fmt.Println()}`) // but it's ok
}
```
Please see other implementations: [Java](https://github.com/JetBrains/intellij-community/blob/master/java/java-psi-impl/src/com/intellij/psi/impl/source/tree/injected/StringLiteralEscaper.java) and [JavaScript](https://github.com/JetBrains/intellij-community/blob/master/json/src/com/intellij/json/psi/impl/JSStringLiteralEscaper.java) and don't hesitate to ask us any questions.

src/com/goide/editor/GoStringManipulator.java
```diff
@@ -0,0 +1,44 @@
...
+
+import com.goide.psi.impl.GoStringLiteralImpl;
+import com.intellij.openapi.util.TextRange;
+import com.intellij.psi.AbstractElementManipulator;
+import com.intellij.util.IncorrectOperationException;
+
+/**
+ * @author stuartcarnie
+ */
+public class GoStringManipulator extends AbstractElementManipulator<GoStringLiteralImpl> {
```
ignatov: Please use `infer nullity` or just `cleanup` action and annotate your code.

stuartcarnie:
No worries; I'll take care of that

grammars/go.bnf
```diff
@@ -104,7 +104,7 @@ ImportSpec ::= [ '.' | identifier ] ImportString {
   stubClass="com.goide.stubs.GoImportSpecStub"
   methods=[getAlias getLocalPackageName]
 }
-ImportString ::= string {methods=[getReferences resolve getPath getPathTextRange]}
+ImportString ::= string | raw_string {methods=[getReferences resolve getPath getPathTextRange]}
```
zolotov: What is the 'raw_string'? Where is its definition?

src/com/goide/util/GoStringLiteralEscaper.java
```diff
@@ -0,0 +1,320 @@
...
+    super(host);
+  }
+
+  @Override
+  public boolean decode(@NotNull final TextRange rangeInsideHost, @NotNull final StringBuilder outChars) {
+    TextRange.assertProperRange(rangeInsideHost);
+
+    String subText = rangeInsideHost.substring(myHost.getText());
+
+    if (myHost.getRawString() != null) {
```
zolotov: Considering missing raw_string definition, this predicate is always false

src/com/goide/util/GoStringLiteralEscaper.java
```diff
@@ -0,0 +1,320 @@
...
+      if (offset > rangeInsideHost.getEndOffset()) offset = rangeInsideHost.getEndOffset();
+      return offset;
+    }
+
+    int result = offsetInDecoded < outSourceOffsets.length ? outSourceOffsets[offsetInDecoded] : -1;
+    if (result == -1) return -1;
+    return (result <= rangeInsideHost.getLength() ? result : rangeInsideHost.getLength()) + rangeInsideHost.getStartOffset();
+  }
+
+  public static boolean parseStringCharacters(String chars, StringBuilder outChars, Ref<int[]> sourceOffsetsRef) {
```
zolotov: Why public? Why static with Ref passing instead of modifying instance's field? Why returns boolean that nobody uses instead of returning offsets-array? 
I see that the JSON escaper was an example for you but those implementation has sensible answers for all these questions and this one hasn't.
Also if 'parseStringCharacters' is fully duplicate of JSON's then you can just use it without copying it to go-plugin.

zolotov:
Also, please add tests for decode and getOffset methods. You can find examples [here](https://github.com/JetBrains/intellij-community/blob/master/python/testSrc/com/jetbrains/python/PyStringLiteralTest.java). Method that creates string literal you can place to GoElementFactory.

grammars/go.bnf
ignatov: That's ok

src/com/goide/util/GoStringLiteralEscaper.java
stuartcarnie: raw_string is defined in lexer

src/com/goide/util/GoStringLiteralEscaper.java
stuartcarnie: As you mentioned, it was copied from the IntelliJ JSON implementation shared on gitter.im, which is the answer to all your 'why' questions. Easy enough to change visibility and switch to using member variables. It was also changed slightly to suit Go strings.

PR #1349, opened 2015-02-21
Improved support for AppEngine SDK (fixes #1278)

### LINKED ISSUE #1278: [1.0] Support for AppEngine Go SDK
AppEngine SDK is not supported right now and it should be.

dlsniper:
With these changes I can now set the Go AppEngine SDK again.
There's a TODO which should be looked at before doing the merge. If agreed I can simply remove the commented line but I want to have some input first.

Thank you.

### DIFF
--- a/utils/src/com/goide/GoEnvironmentUtil.java
+++ b/utils/src/com/goide/GoEnvironmentUtil.java
@@ -1,6 +1,5 @@
 package com.goide;
 
-import com.intellij.execution.configurations.PathEnvironmentVariableUtil;
 import com.intellij.openapi.application.PathMacros;
 import com.intellij.openapi.util.SystemInfo;
 import com.intellij.openapi.util.io.FileUtil;
@@ -13,15 +12,18 @@ import java.io.File;
 
 public class GoEnvironmentUtil {
   public static final String GO_EXECUTABLE_NAME = "go";
-  
+  public static final String GAE_EXECUTABLE_NAME = "goapp";
+
   private GoEnvironmentUtil() {
   }
 
   @NotNull
   public static File getExecutableForSdk(@NotNull String sdkHome) {
-    File fromSdkPath = getExecutable(new File(sdkHome, "bin").getAbsolutePath(), GO_EXECUTABLE_NAME);
-    File fromEnvironment = PathEnvironmentVariableUtil.findInPath(GO_EXECUTABLE_NAME);
-    return fromSdkPath.canExecute() || fromEnvironment == null ? fromSdkPath : fromEnvironment;
+    File goFromSdkPath = getExecutable(new File(sdkHome, "bin").getAbsolutePath(), GO_EXECUTABLE_NAME);
+    File gaeFromSdkPath = getExecutable(new File(sdkHome, "bin").getAbsolutePath(), GAE_EXECUTABLE_NAME);
+    // TODO Readd this if it's needed but it shouldn't be used in the SDK detection part
+    //File fromEnvironment = PathEnvironmentVariableUtil.findInPath(GO_EXECUTABLE_NAME);
+    return gaeFromSdkPath.canExecute() ? gaeFromSdkPath : goFromSdkPath;
   }
 
   @NotNull


jps-plugin/src/com/goide/jps/builder/GoBuilder.java
```diff
@@ -63,7 +66,14 @@ public void build(@NotNull GoTarget target,
     JpsTypedModule<JpsSimpleElement<JpsGoModuleProperties>> module = jpsModule.asTyped(JpsGoModuleType.INSTANCE);
     assert module != null;
     JpsSdk<JpsDummyElement> sdk = getSdk(context, module);
-    File executable = GoEnvironmentUtil.getExecutableForSdk(sdk.getHomePath());
+    File exec = GoEnvironmentUtil.getExecutableForSdk(sdk.getHomePath());
+    if (exec == null) {
+      Notifications.Bus.notify(
```
zolotov: Many notification duplicates

utils/src/com/goide/GoEnvironmentUtil.java
```diff
@@ -13,15 +12,22 @@
...
 public class GoEnvironmentUtil {
   public static final String GO_EXECUTABLE_NAME = "go";
-  
+  public static final String GAE_EXECUTABLE_NAME = "goapp";
+
   private GoEnvironmentUtil() {
   }
 
-  @NotNull
+  @Nullable
```
zolotov: Why do you need return null? Ok, file is not really exist, but by returning null you make client to check whether it null and handle this situation, whereas most of them don't have to know about it, they just create GeneralCommandLine and hope the everything will be fine.

jps-plugin/src/com/goide/jps/builder/GoBuilder.java
dlsniper: I know that the notification itself is a duplicate. Maybe it could be extracted into a common static method that could receive just the title of the notification, as that's different between them.
However, this way we can cover situations like having an update change the location of the go tool (if the user installs a new version from a package manager not from the official release) and IDEA is still running.

Should I extract the method to ```utils/src/com/goide/GoEnvironmentUtil.java``` or is there a better place for it?

utils/src/com/goide/GoEnvironmentUtil.java
dlsniper: I guess I can change it back to not returning null and if the ```canExecute()``` call is successful then return the path with ```getAbsolutePath()``` already called.

However, I'd prefer for the clients to have a way to check and signal if something is missing from their SDK directory even after the SDK was configured the first time around as we rely on things such as ```go fmt``` or ```go run``` and I think it's not nice not having any feedback for the user in case something went wrong as to why it happened.

jps-plugin/src/com/goide/jps/builder/GoBuilder.java
zolotov: GoEnvironmentUtil is ok for that. 
Also, as I see all clients of getExecutablePath method show these notifications, so you can add parameter to getExecutablePath like (showNotificationOnFail) and move showing notifications there.

utils/src/com/goide/GoEnvironmentUtil.java
zolotov: >  I'd prefer for the clients to have a way to check

The method return File, not path, so clients have plenty ways to check it.

> it's not nice not having any feedback for the user in case something went wrong as to why it happened.

You'll get feedback from GeneralCommandLine. getExecutableForSdk method is just 'getter', it doesn't give any guarantees about correctness of returning file and I believe it shouldn't just because it cannot know how client will use the result.

Anyway, as I see you're solving at least two problems in this commit. The first one is about looking for GAE-executable. And the second one is not really clear for me. Let's separate this commit it two, I'll merge the first one and we'll discuss the second.

dlsniper:
@zolotov I've reverted all the other changes except the change with getting the go path from the environment. I still think that should be done there, especially since the method is called when trying to configure an SDK which might very well be in a different directory than the one that the environment variable is pointing to.

zolotov:
> I still think that shouldn't be done there, especially since the method is called when trying to configure an SDK which might very well be in a different directory than the one that the environment variable is pointing to.

Agreed. And merged.

PR #1354, opened 2015-02-22
Add Github error handling reporter

dlsniper:
I've ported the error handler we've had on 0.9.16 to 1.0. I've seen that previously it was using the internal one but in this case, I can't see the crashes for example. I'm still petitioning Github to add issue deduplication so this doesn't become a problem in case of too many crashes.


dlsniper:
Closing this per talk on Gitter.

PR #1357, opened 2015-02-22
Add a usage notification 

dlsniper:
This adds a notification on first run in order to help the user to setup a new project using the plugin.

PR #1360, opened 2015-02-23
Add a usage notification

dlsniper:
This is the refactoring of #1357

### DIFF
--- a/src/com/goide/ui/ProjectTutorialNotification.java
+++ b/src/com/goide/ui/ProjectTutorialNotification.java
@@ -1,0 +1,65 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Mihai Toader, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.ui;
+
+import com.intellij.ide.util.PropertiesComponent;
+import com.intellij.notification.*;
+import com.intellij.openapi.components.*;
+import com.intellij.util.text.VersionComparatorUtil;
+import org.jetbrains.annotations.NotNull;
+
+import static com.goide.util.GoUtil.PLUGIN_VERSION;
+
+public class ProjectTutorialNotification implements ApplicationComponent {
+
+  private static final String GO_PROJECT_TUTORIAL_NOTIFICATION_SHOWN = "go.tutorial.project.notification.shown";
+
+  @Override
+  public void initComponent() {
+    final PropertiesComponent propertiesComponent = PropertiesComponent.getInstance();
+    boolean shownAlready;
+    //noinspection SynchronizationOnLocalVariableOrMethodParameter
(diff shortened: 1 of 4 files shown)


dlsniper:
@zolotov I've changed the PR as you advised. The only issue that I have right now is persisting the state as I don't understand what I'm missing to enable that one.

zolotov:
@dlsniper as I said use PropertiesComponent.getInstance() for persistent

zolotov:
https://github.com/dlsniper/google-go-lang-idea-plugin/blob/usage--notification/src/com/goide/project/GoModuleLibrariesInitializer.java#L263, but don't pass the project in order to save state on application level

dlsniper:
Done. I didn't understood I should use the "getInstance()" vs implement the methods themselves, hence the confusion. Thanks for helping out.

src/com/goide/ui/ProjectTutorialNotification.java
```diff
@@ -0,0 +1,75 @@
...
+      if (!shownAlready) {
+        propertiesComponent.setValue(GO_PROJECT_TUTORIAL_NOTIFICATION_SHOWN, PLUGIN_VERSION);
+      }
+    }
+    
+    if (shownAlready) {
+      return;
+    }
+
+    NotificationGroup group = NotificationGroup.findRegisteredGroup("Plugins Suggestion");
```
zolotov: Still don't understand why do you need group. Why not just

```
Notifications.Bus.notify(new Notification());
```

Group will be found by its name or created.

'Plugins Suggestion' is not the best name, it's unlikely that user will be able to associate it with Go Plugin. Use 'Go'-group (it's used in quick fixes) or make the new one but with 'Go' word.

src/com/goide/ui/ProjectTutorialNotification.java
```diff
@@ -0,0 +1,75 @@
...
+  public void initComponent() {
+    final PropertiesComponent propertiesComponent = PropertiesComponent.getInstance();
+    boolean shownAlready;
+    //noinspection SynchronizationOnLocalVariableOrMethodParameter
+    synchronized (propertiesComponent) {
+      String storedVersion = propertiesComponent.getValue(GO_PROJECT_TUTORIAL_NOTIFICATION_SHOWN, "0.0");
+      shownAlready = VersionComparatorUtil.compare(storedVersion, PLUGIN_VERSION) == 0;
+
+      if (!shownAlready) {
+        propertiesComponent.setValue(GO_PROJECT_TUTORIAL_NOTIFICATION_SHOWN, PLUGIN_VERSION);
```
zolotov: I believe that you can set value without checking 'shownAlready'

zolotov:
It's ok now. Please simplify notification showing a little.

src/com/goide/ui/ProjectTutorialNotification.java
dlsniper: Unfortunately ```getOrInit()``` doesn't return null if the value didn't existed before.
Wouldn't it be better to write the value there only if it's not there already? If I understand correctly, please correct me if not, the value would be written every time the IDE starts up and I'd rather avoid that.

zolotov: I didn't say anything about getOrInit(), I said that ```if (!showAlready)``` seems to be redundant

dlsniper: Like I've said, wouldn't it create one more disk write on startup every time IDEA starts? I'm happy to remove the check, but I want to understand how things work better before doing so :) Thank you.

zolotov: > Like I've said, wouldn't it create one more disk write on startup every time IDEA starts?

No, you'll just add value to map. PropertiesComponent will make disk write operations no matter you added/replaced value or not.

dlsniper: I see, thanks. Updated and squashed. Thank you.

PR #1363, opened 2015-02-23
[WIP] Formatter (fixes #1215)

### LINKED ISSUE #1215: [1.0] Formatter doesn't follow the same rules as gofmt or goimports
The formatting in IDEA is very different than the one from gofmt (or goimports).

I know this is a pretty generic ticket so let me know how should I best add information to it.
In 0.9.x branch there's a bunch of tests that should cover the source code formatting, and that's a very good start for it and Mihai started working on a new formatting engine a while ago, which can be found here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/tree/formatter_work 

Since the go community is very keen on having all projects respecting gofmt, this should be a must for the initial stable release of the plugin.

Alternatively, integration with gofmt or goimports like we have for 0.9.16 could be done, where we execute gofmt/goimports on save, which could be good enough to start with, even if it's against the way IDEA should handle the issue.

Thanks.

dlsniper:
**Work in progress**

This attempts to make the formatter in order be 1:1 compatible with gofmt.
At a later point it would be nice to have an option to make it compatible with goimports (which does some more work in regards to how it processes the imports section).


src/com/goide/formatter/GoFormattingModelBuilder.java
```diff
@@ -41,6 +41,15 @@
   @NotNull
   @Override
   public FormattingModel createModel(@NotNull PsiElement element, @NotNull CodeStyleSettings settings) {
+    // TODO decide if this is a good thing to do or not.
+    // Given the go docs and the community enforce the rules for tabs
+    // The plugin should try and emulate the gofmt behavior, which replaces spaces with tabs for idents
+    // For more detail see: http://golang.org/doc/effective_go.html#formatting
+    if (settings.getIndentOptions() != null) {
```
zolotov: GoLanguageCodeStyleSettingsProvider should be used for setting default values

dlsniper:
I've added the same tests that Go is using (see [here](https://github.com/golang/go/tree/master/src/cmd/gofmt/testdata)). Now the fun part.

gofmt supports a ``` gofmt -s ``` mode in which it simplifies the source code besides the formatting it does. However, if I understand correctly how IDEA is supposed to work, I think those test shouldn't be part of this and rather be moved to other parts such as quick fixes (or something more appropriate, if any).

Also, any help to understand how the formatting engine in IDEA works would be great. I'd like to know how the processing is done, in what order is the source processed and so on. 

Another issue is that gofmt tool formats comments, which I'm not sure it's possible (am I wrong?) to do today with IDEA?

Thank you!

zolotov:
> gofmt supports a gofmt -s mode in which it simplifies the source code besides the formatting it does. However, if I understand correctly how IDEA is supposed to work, I think those test shouldn't be part of this and rather be moved to other parts such as quick fixes (or something more appropriate, if any).

Depends on particular "simplification". Formatter is able to do some manipulations with code, but many of them also could be represented as quick fixes too. Let's discuss each simplification separately, I believe they are not the part of this PR. 

> Another issue is that gofmt tool formats comments, which I'm not sure it's possible (am I wrong?) to do today with IDEA?

When I can read about gofmt comment?

dlsniper:
> When I can read about gofmt comment?

Please see this [playground example](http://play.golang.org/p/5DHMUVUx_h) . Press the format button and see how the indentation changes for the whole block. Not sure if I explained clear enough.

zolotov:
I meant 'where' of course :)

Well, the feature with comments is quite silly imho. Anyway I'll try to figure out whether it could be done without parsing comments.

zolotov:
There three options to make this work:
1. implement PostFormatProcessor that will calculate indent for each comment string and patch psi
2. reimplement go formatter from PsiBasedFormattingModel to DocumentBasedFormatting that will create formatting blocks for each line in comment
3. split comment tokens into many 'line comments' in lexer, fix parser and build formatting blocks for each line comment

The first option looks easiest

dlsniper:
@zolotov and @ignatov sorry for delay. I've finally got the time to move the tests to another branch. It's called ```formatter-tests``` ```https://github.com/go-lang-plugin-org/go-lang-idea-plugin/tree/formatter-tests```. Hope this helps.

dlsniper:
Also, please let me know if there's any way I could help with the formatter work. I've tried reading more about it from here: https://confluence.jetbrains.com/display/IntelliJIDEA/Formatter but the docs don't cover cases which are encountered in Go for example, or in other language to inspire from them.

PR #1365, opened 2015-02-23
[WIP] Introduce ability to run gofmt or goimports on save (fixes #1324)

### LINKED ISSUE #1324: Execute Go format after saving file
hotkey on macosx: cmd+alt+shift+f (uncomfortably)
please execute command "go fmt" after saving file (great)

thank you

dlsniper:
This is just a port of the functionality we've had in the master branch.

I know it's a bit ugly but I'll rework it where needed.

### DIFF
--- a/src/META-INF/plugin.xml
+++ b/src/META-INF/plugin.xml
@@ -38,7 +38,7 @@
   <idea-version since-build="139.1111"/>
   <depends>com.intellij.modules.lang</depends>
   <depends optional="true" config-file="java-deps.xml">com.intellij.modules.java</depends>
-  
+
   <module-components>
     <component>
       <implementation-class>com.goide.project.GoModuleLibrariesInitializer</implementation-class>
@@ -57,12 +57,15 @@
     <internalFileTemplate name="Go File"/>
 
     <!-- project -->
+    <applicationService
+        serviceImplementation="com.goide.settings.GoApplicationSettings"/>
+
     <applicationService serviceImplementation="com.goide.project.GoApplicationLibrariesService"/>
     <projectService serviceImplementation="com.goide.project.GoProjectLibrariesService"/>
     <moduleService serviceImplementation="com.goide.project.GoModuleLibrariesService"/>
-    <projectConfigurable groupId="language" provider="com.goide.configuration.GoConfigurableProvider" dynamic="true" 
+    <projectConfigurable groupId="language" provider="com.goide.configuration.GoConfigurableProvider" dynamic="true"
                          bundle="do.not.touch.this.attribute" />
-    
+
     <defaultLiveTemplatesProvider implementation="com.goide.template.GoLiveTemplatesProvider"/>
     <liveTemplateContext implementation="com.goide.template.GoLiveTemplateContextType$GoEverywhereContextType"/>
     <liveTemplateContext implementation="com.goide.template.GoLiveTemplateContextType$GoFileContextType"/>
@@ -76,11 +79,11 @@
     <errorHandler implementation="com.intellij.diagnostic.ITNReporter"/>
     <extendWordSelectionHandler implementation="com.goide.editor.GoWordSelectioner"/>
     <annotator language="go" implementationClass="com.goide.GoAnnotator"/>
-    
+
     <lang.importOptimizer language="go" implementationClass="com.goide.codeInsight.imports.GoImportOptimizer"/>
     <psi.fileReferenceHelper implementation="com.goide.psi.impl.imports.GoImportReferenceHelper"/>
     <fileContextProvider implementation="com.goide.psi.impl.imports.GoFileContextProvider"/>
-    
+
     <lang.parserDefinition language="go" implementationClass="com.goide.GoParserDefinition"/>
     <fileTypeFactory implementation="com.goide.GoFileTypeFactory"/>
     <lang.syntaxHighlighterFactory key="go" implementationClass="com.goide.highlighting.GoSyntaxHighlighterFactory"/>
@@ -107,10 +110,11 @@
     <codeInsight.lineMarkerProvider language="go" implementationClass="com.goide.editor.marker.GoMethodSeparatorProvider"/>
 
     <!--formatter-->
+    <fileDocumentManagerListener implementation="com.goide.formatter.GoOnSaveFormatter"/>
     <lang.formatter language="go" implementationClass="com.goide.formatter.GoFormattingModelBuilder"/>
     <codeStyleSettingsProvider implementation="com.goide.formatter.settings.GoCodeStyleSettingsProvider"/>
     <langCodeStyleSettingsProvider implementation="com.goide.formatter.settings.GoLanguageCodeStyleSettingsProvider"/>
-    
+
     <renamePsiElementProcessor implementation="com.goide.refactor.GoAnonymousFieldProcessor"/>
 
     <gotoSymbolContributor implementation="com.goide.go.GoSymbolContributor"/>
@@ -148,9 +152,9 @@
     <programRunner implementation="com.goide.debugger.ideagdb.run.GdbRunner"/>
 
     <checkinHandlerFactory implementation="com.goide.actions.fmt.GoFmtCheckinFactory" order="last"/>
-    
+
     <lang.inspectionSuppressor language="go" implementationClass="com.goide.inspections.suppression.GoInspectionSuppressor"/>
-    
+
     <localInspection language="go" displayName="Unresolved reference inspection"
                      groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.unresolved.GoUnresolvedReferenceInspection"/>
@@ -158,16 +162,16 @@
                      groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoUnusedImportDeclaration"/>
     <localInspection language="go" displayName="Duplicate fields and methods inspection"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoDuplicateFieldsOrMethodsInspection"/>
     <localInspection language="go" displayName="Unused variables inspection"
-                     groupName="Go" enabledByDefault="true" level="ERROR"  
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.unresolved.GoUnusedVariableInspection"/>
     <localInspection language="go" displayName="Unused global variables inspection"
-                     groupName="Go" enabledByDefault="true" level="WARNING" 
+                     groupName="Go" enabledByDefault="true" level="WARNING"
                      implementationClass="com.goide.inspections.unresolved.GoUnusedGlobalVariableInspection"/>
     <localInspection language="go" displayName="Unused function inspection"
-                     groupName="Go" enabledByDefault="true" level="WARNING" 
+                     groupName="Go" enabledByDefault="true" level="WARNING"
                      implementationClass="com.goide.inspections.unresolved.GoUnusedFunctionInspection"/>
     <localInspection language="go" displayName="Assignment to constant"
                      groupName="Go" enabledByDefault="true" level="ERROR"
@@ -176,25 +180,25 @@
                      groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoDuplicateFunctionInspection"/>
     <localInspection language="go" displayName="Duplicate argument"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoDuplicateArgumentInspection"/>
     <localInspection language="go" displayName="Duplicate return argument"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoDuplicateReturnArgumentInspection"/>
     <localInspection language="go" displayName="Incorrect variadic parameter"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoFunctionVariadicParameterInspection"/>
     <localInspection language="go" displayName="Incorrect variable declaration"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoVarDeclarationInspection"/>
     <localInspection language="go" displayName="No new variables on left side of :="
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoNoNewVariablesInspection"/>
     <localInspection language="go" displayName="Missing return at end of function"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoReturnInspection"/>
     <localInspection language="go" displayName="Function call inspection"
-                     groupName="Go" enabledByDefault="true" level="ERROR" 
+                     groupName="Go" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoFunctionCallInspection"/>
     <localInspection language="go" displayName="Defer/go statements check"
                      groupName="Go" enabledByDefault="true" level="ERROR"
@@ -205,7 +209,7 @@
             text="Go File" description="Create new Go file">
       <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFile"/>
     </action>
-    
+
     <action id="GoShowTypeInternalAction"
             class="com.goide.actions.internal.GoShowTypeInternalAction"
             text="Show Go type info">
--- a/src/com/goide/actions/fmt/GoFmtFileAction.java
+++ b/src/com/goide/actions/fmt/GoFmtFileAction.java
@@ -18,10 +18,9 @@ package com.goide.actions.fmt;
 
 import com.goide.GoEnvironmentUtil;
 import com.goide.sdk.GoSdkService;
+import com.goide.util.GoExecutor;
 import com.intellij.execution.ExecutionException;
-import com.intellij.execution.configurations.GeneralCommandLine;
-import com.intellij.execution.process.KillableColoredProcessHandler;
-import com.intellij.execution.process.OSProcessHandler;
+import com.intellij.execution.configurations.PathEnvironmentVariableUtil;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.openapi.application.ApplicationManager;
@@ -49,40 +48,20 @@ public class GoFmtFileAction extends GoExternalToolsAction {
     String filePath = virtualFile.getCanonicalPath();
     assert filePath != null;
 
-    GeneralCommandLine commandLine = new GeneralCommandLine();
+    File executable = GoEnvironmentUtil.getBinaryFilePathFromEnvironment("gofmt");
 
-    String sdkHome = GoSdkService.getInstance(project).getSdkHomePath(ModuleUtilCore.findModuleForPsiElement(file));
-    if (StringUtil.isEmpty(sdkHome)) {
-      warning(project, groupId, "Project sdk is not valid");
+    if (executable == null) {
+      warning(project, groupId, "Looks like you haven't any gofmt executable in your PATH");
       return true;
     }
 
-    File executable = GoEnvironmentUtil.getExecutableForSdk(sdkHome);
-
-    commandLine.setExePath(executable.getAbsolutePath());
-    commandLine.addParameters("fmt", filePath);
-
     FileDocumentManager.getInstance().saveDocument(document);
 
-    String commandLineString = commandLine.getCommandLineString();
-    OSProcessHandler handler = new KillableColoredProcessHandler(commandLine.createProcess(), commandLineString);
-    handler.addProcessListener(new ProcessAdapter() {
-      @Override
-      public void processTerminated(ProcessEvent event) {
-        ApplicationManager.getApplication().invokeLater(new Runnable() {
-          @Override
-          public void run() {
-            ApplicationManager.getApplication().runWriteAction(new Runnable() {
-              @Override
-              public void run() {
-                virtualFile.refresh(false, false);
-              }
-            });
-          }
-        });
-      }
-    });
-    handler.startNotify();
+    GoExecutor.
+      in(project).
+      addParameters("fmt", filePath).
+      showOutputOnError().
+      execute();
     return false;
   }
 
--- a/src/com/goide/actions/fmt/GoImportsFileAction.java
+++ b/src/com/goide/actions/fmt/GoImportsFileAction.java
@@ -18,11 +18,9 @@ package com.goide.actions.fmt;
 
 import com.goide.GoEnvironmentUtil;
 import com.goide.sdk.GoSdkService;
+import com.goide.util.GoExecutor;
 import com.intellij.execution.ExecutionException;
-import com.intellij.execution.configurations.GeneralCommandLine;
 import com.intellij.execution.configurations.PathEnvironmentVariableUtil;
-import com.intellij.execution.process.KillableColoredProcessHandler;
-import com.intellij.execution.process.OSProcessHandler;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
 import com.intellij.openapi.application.ApplicationManager;
@@ -52,8 +50,8 @@ public class GoImportsFileAction extends GoExternalToolsAction {
   protected String getErrorTitle(@NotNull String fileName) {
     return fileName + " optimizing imports with goimports failed. You maybe need to install goimports.";
   }
-  
-  
+
+
   public boolean doSomething(@NotNull PsiFile file,
                              @NotNull Project project,
                              @Nullable final VirtualFile virtualFile,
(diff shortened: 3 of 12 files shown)


src/META-INF/plugin.xml
```diff
@@ -45,6 +45,12 @@
     </component>
   </module-components>
 
+  <project-components>
+    <component>
+      <implementation-class>com.goide.ui.GoToolWindow</implementation-class>
+    </component>
+  </project-components>
```
ignatov: Please use "log only" notification group instead of dedicated tool window.

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    GoProjectSettings.GoProjectSettingsBean settings = GoProjectSettings.getInstance(project).getState();
+
+    final Project p = project;
+    if (settings.goimportsOnSave) {
+      ApplicationManager.getApplication().invokeLater(new Runnable() {
+        @Override
+        public void run() {
+          if (p.isDisposed()) {
```
ignatov: Please invert this if statement.

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,177 @@
...
+          }
+          ProcessFileWithGoImports(p, file);
+        }
+      });
+    }
+    else if (settings.goFmtOnSave) {
+      ApplicationManager.getApplication().invokeLater(new Runnable() {
+        @Override
+        public void run() {
+          if (p.isDisposed()) {
```
ignatov: The same

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,177 @@
...
+            return;
+          }
+          ProcessFileWithGoFmt(p, file);
```
ignatov: We can reuse com.goide.actions.fmt.GoFmtFileAction and com.goide.actions.fmt.GoImportsFileAction

src/META-INF/plugin.xml
```diff
@@ -61,6 +67,14 @@
     <internalFileTemplate name="Go File"/>
 
     <!-- project -->
+    <toolWindow id="Go" anchor="bottom" icon="/icons/go.png" secondary="true" factoryClass="com.goide.ui.GoToolWindowFactory" />
+
+    <projectConfigurable instance="com.goide.ui.GoConfigurable" groupId="tools"/>
+
+    <projectService
```
ignatov: @zolotov Can we use current configurable for storing those two checkboxes?

src/META-INF/plugin.xml
zolotov: Or use ExecutionHelper#showOutput. Dedicated tool window is overkill indeed.

src/META-INF/plugin.xml
zolotov: @ignatov no

@dlsniper put it to "language" group, it's sad to have go-related configurables in different places, I'll made a Go-group later. Also meaningful name for configurable is welcome.

src/com/goide/actions/fmt/GoFmtFileAction.java
```diff
@@ -59,6 +59,7 @@ public boolean doSomething(@NotNull PsiFile file,
 
     commandLine.setExePath(executable.getAbsolutePath());
     commandLine.addParameters("fmt", filePath);
+    commandLine.setPassParentEnvironment(true);
```
zolotov: It's true by default

src/com/goide/sdk/GoSdkUtil.java
```diff
@@ -68,6 +69,12 @@ private static String getSdkVersionString(@NotNull Project project, @Nullable Mo
   }
   
   @Nullable
+  public static Sdk getSdk(@NotNull Project project) {
```
zolotov: There is new GoSdkService. Please use it.

src/com/goide/ui/GoConfigurableForm.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class GoConfigurableForm {
+
+  public JPanel componentPanel;
+
+  private JRadioButton doNothingOnSave;
+  private JRadioButton goFmtOnSave;
+  private JRadioButton goimportsOnSave;
+
+  public GoConfigurableForm() {
```
zolotov: redundant constructor

src/com/goide/ui/GoConsoleFilter.java
```diff
@@ -0,0 +1,72 @@
...
+import com.intellij.execution.filters.Filter;
+import com.intellij.execution.filters.HyperlinkInfo;
+import com.intellij.execution.filters.OpenFileHyperlinkInfo;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.vfs.VirtualFile;
+
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+class GoConsoleFilter implements Filter {
```
zolotov: There is GoTestConsoleFilter with the same pattern, please reuse it.

utils/src/com/goide/GoEnvironmentUtil.java
```diff
@@ -47,4 +49,19 @@ public static String retrieveGoPathFromEnvironment() {
     String path = EnvironmentUtil.getValue(GoConstants.GO_PATH);
     return path != null ? path : PathMacros.getInstance().getValue(GoConstants.GO_PATH);
   }
+  
+  @NotNull
+  public static String[] getEnvironment(@NotNull Project project, @NotNull String sdkHomePath) {
+    String[] env = EnvironmentUtil.getEnvironment();
+    
+    for (int i=0; i < env.length; i++) {
+      if (env[i].toLowerCase().startsWith("path=")) {
```
zolotov: What is it? Why do I need this?

zolotov: And why project base path?

src/com/goide/settings/GoProjectSettings.java
```diff
@@ -0,0 +1,54 @@
...
+import org.jetbrains.annotations.NotNull;
+
+@State(
+  name = "GoProjectSettings",
+  storages = {
+    @Storage(id = "default", file = "$PROJECT_FILE$"),
+    @Storage(id = "dir", file = "$PROJECT_CONFIG_DIR$/go.xml", scheme = StorageScheme.DIRECTORY_BASED)
+  }
+)
+public class GoProjectSettings implements PersistentStateComponent<GoProjectSettings.GoProjectSettingsBean> {
```
zolotov: Are you sure that these are project-level settings? Maybe application-level?

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,177 @@
...
+        }
+      });
+    }
+  }
+
+  private void ProcessFileWithGoImports(Project project, VirtualFile file) {
+    GoToolWindow toolWindow = GoToolWindow.getInstance(project);
```
zolotov: By the way, I think that since goimports/gofmt are run on single particular file, we should get module SDK (see ModuleUtilCore). And use project sdk if there is not module for file. GoFmtAction and GoImportsAction should be changed as well.

Feel free to extend GoSdkService to retrieve sdkHomePath by virtualFile.

utils/src/com/goide/GoEnvironmentUtil.java
dlsniper: > What is it? Why do I need this?
This is needed as without it, on my system, IDEA can't process the correct path. When running ```go fmt``` that will execute ```gofmt``` but since it's calling the executable without a full path, as the ```PATH``` environment variable is not properly passed, this will fail.

There's also a problem that I've commented on [YouTrack](https://youtrack.jetbrains.com/issue/IDEA-118483). Also, from the previous experience with 0.9.x anything that relies on the environment variables is not happily handled in IDEA and it's better to stay away from it.

dlsniper:
Thanks for the feedback, I'll make the changes when I get some time for it and update the PR.

utils/src/com/goide/GoEnvironmentUtil.java
zolotov: It looks like problem of go. So let's invoke ```gofmt``` explicitly with full path instead of ```go fmt```. This magic with path variable looks unreliable.

BTW, it's strange to get map, flattern it and then look for string that starts by key of initial map (see method implementation for details). Also startsWith is case sensitive and can't handle differences between OS. So for such cases use EnvironmentUtil#getValue

dlsniper:
I've did some of the changes as indicated on the feedback.

What remains to be done is: 
- figure out a way to make it run on my system (see: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1365#discussion_r25253981)
- unify the settings entries / windows with the Go Libraries one? I'm thinking of creating a tabbed entry for Go called General and Libraries, would that work?
- get feedback and squash commits

src/META-INF/plugin.xml
```diff
@@ -57,12 +57,18 @@
     <internalFileTemplate name="Go File"/>
 
     <!-- project -->
+    <projectConfigurable instance="com.goide.ui.GoConfigurable" groupId="language"/>
```
zolotov: Please delete configurable definition form plugin.xml and add its constructing to GoConfigurableProvider, I'll clean up this later.

src/META-INF/plugin.xml
```diff
@@ -57,12 +57,18 @@
...
+
+    <applicationService
+        serviceInterface="com.goide.settings.GoApplicationSettings"
```
zolotov: serviceInterface attribute is redundant

src/com/goide/settings/GoApplicationSettings.java
```diff
@@ -0,0 +1,53 @@
...
+    return bean != null ? bean : new GoProjectSettingsBean();
+  }
+
+  @Override
+  public void loadState(GoProjectSettingsBean settingsBean) {
+    this.bean = settingsBean;
+  }
+
+  @NotNull
+  public static GoApplicationSettings getInstance(Project project) {
```
zolotov: You're defined it as application service, try to get project-specific service, store state in application level and use it in project-level configurable :) You should decide what level are needed for all these cases.

src/com/goide/ui/GoConfigurable.java
```diff
@@ -0,0 +1,95 @@
...
+import com.intellij.openapi.options.SearchableConfigurable;
+import com.intellij.openapi.project.Project;
+import org.jetbrains.annotations.Nls;
+import org.jetbrains.annotations.NotNull;
+
+import javax.swing.*;
+
+public class GoConfigurable implements SearchableConfigurable {
+
+  private GoConfigurableForm configurableForm;
```
zolotov: Do you really need separated form? Are you going to reuse it somewhere?

src/com/goide/settings/GoApplicationSettings.java
dlsniper: Ah, I now understand what you mean with the previous comment. Yes, it should be per application not per project, I'll change that. Thank you.

src/com/goide/util/GoExecutor.java
```diff
@@ -70,7 +75,13 @@ public GoExecutor withExePath(@Nullable String exePath) {
     myExePath = exePath;
     return this;
   }
-  
+
+  @NotNull
+  public GoExecutor withCustomExecutable(@NotNull File customExecutable) {
```
zolotov: There is already withExePath, please remove all customExecutable usages

utils/src/com/goide/GoEnvironmentUtil.java
```diff
@@ -45,4 +47,19 @@ public static String retrieveGoPathFromEnvironment() {
     String path = EnvironmentUtil.getValue(GoConstants.GO_PATH);
     return path != null ? path : PathMacros.getInstance().getValue(GoConstants.GO_PATH);
   }
+  
+  @NotNull
+  public static String[] getEnvironment(@NotNull Project project, @NotNull String sdkHomePath) {
```
zolotov: Unused?

src/com/goide/sdk/GoSdkUtil.java
```diff
@@ -72,6 +73,12 @@ public static VirtualFile getSdkSrcDir(@NotNull PsiElement context) {
 
   // todo: caching
   @Nullable
+  public static Sdk getSdk(@NotNull Project project) {
```
zolotov: Unused?

src/com/goide/actions/fmt/GoFmtFileAction.java
```diff
@@ -49,24 +48,27 @@ public boolean doSomething(@NotNull PsiFile file,
...
-    handler.addProcessListener(new ProcessAdapter() {
+    GoExecutor goExecutor = GoExecutor.
+      in(project).
+      addParameters("fmt", filePath).
+      showOutputOnError();
+
+    goExecutor.addProcessListener(new ProcessAdapter() {
       @Override
       public void processTerminated(ProcessEvent event) {
         ApplicationManager.getApplication().invokeLater(new Runnable() {
```
zolotov: GoExecutor refreshes VFS, looks like process listeners aren't needed yet

src/com/goide/actions/fmt/GoFmtFileAction.java
```diff
@@ -49,24 +48,27 @@ public boolean doSomething(@NotNull PsiFile file,
...
-    GeneralCommandLine commandLine = new GeneralCommandLine();
-
     String sdkHome = GoSdkService.getInstance(project).getSdkHomePath(ModuleUtilCore.findModuleForPsiElement(file));
     if (StringUtil.isEmpty(sdkHome)) {
       warning(project, groupId, "Project sdk is not valid");
       return true;
     }
 
-    File executable = GoEnvironmentUtil.getExecutableForSdk(sdkHome);
+    File executable = PathEnvironmentVariableUtil.findInPath(GoEnvironmentUtil.getBinaryFileNameForPath("gofmt"));
```
zolotov: Looks like candidate for GoEnvironmentUtil method

src/com/goide/actions/fmt/GoFmtFileAction.java
```diff
@@ -49,24 +48,27 @@ public boolean doSomething(@NotNull PsiFile file,
...
 
-    String commandLineString = commandLine.getCommandLineString();
-    OSProcessHandler handler = new KillableColoredProcessHandler(commandLine.createProcess(), commandLineString);
-    handler.addProcessListener(new ProcessAdapter() {
+    GoExecutor goExecutor = GoExecutor.
+      in(project).
+      addParameters("fmt", filePath).
+      showOutputOnError();
+
+    goExecutor.addProcessListener(new ProcessAdapter() {
```
zolotov: GoExecutor refreshes VFS, looks like process listeners aren't needed yet

src/com/goide/actions/fmt/GoImportsFileAction.java
```diff
@@ -64,7 +62,6 @@ public boolean doSomething(@NotNull PsiFile file,
     String filePath = virtualFile.getCanonicalPath();
     assert filePath != null;
 
-    GeneralCommandLine commandLine = new GeneralCommandLine();
     String sdkHome = GoSdkService.getInstance(project).getSdkHomePath(ModuleUtilCore.findModuleForPsiElement(file));
     if (StringUtil.isEmpty(sdkHome)) {
```
zolotov: GoExecutor will throw execution error if sdk is not set, don't need check it twice, you can catch the exception or just let executor write the error message in output console

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,109 @@
...
+      return;
+    }
+
+    final VirtualFile file = FileDocumentManager.getInstance().getFile(document);
+    if (file == null || file.getFileType() != GoFileType.INSTANCE) {
+      return;
+    }
+
+    Project project = null;
+    for (Project possibleProject : projects) {
```
zolotov: Don't use source/content roots ever, in go we can hardly manage it and using it leads to problems with different go-project workflow

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,109 @@
...
+    if (project == null) {
+      return;
+    }
+
+    final PsiFile psiFile = PsiManager.getInstance(project).findFile(file);
+    if (psiFile == null) {
+      return;
+    }
+
+    GoApplicationSettings.GoProjectSettingsBean settings = GoApplicationSettings.getInstance(project).getState();
```
zolotov: Actually line 50-73: PsiDocumentManager#getPsiFile

src/com/goide/sdk/GoSdkType.java
```diff
@@ -125,4 +125,8 @@ public void setupSdkPaths(@NotNull Sdk sdk) {
     }
     modificator.commitChanges();
   }
+
+  public static boolean isInstance(@NotNull Sdk sdk) {
```
zolotov: Unused?

src/com/goide/actions/fmt/GoImportsFileAction.java
zolotov: Btw, I think we can extract checking sdk in GoExternalToolsAction. Just check in update method that sdkPath is not empty and make setEnabled(false) if it's empty.

zolotov:
> unify the settings entries / windows with the Go Libraries one? I'm thinking of creating a tabbed entry for Go called General and Libraries, would that work?

If you add your configurable in GoConfigurableProvider there will be parent page Go with children Go Libraries and <Your configurable name>.

src/com/goide/formatter/GoOnSaveFormatter.java
dlsniper: But if I do ```PsiDocumentManager.getInstance()``` I get asked for a project parameter. How should I get the project from the document, which would also solve the above issue with iterating in the opened projects list?

zolotov: Indeed, you're right. `ProjectUtil.guessProjectForContentFile` then.

src/com/goide/ui/GoConfigurable.java
dlsniper: I just think it's easier to visualize what's happening. and how things look like. And I don't see any downside of having it as a separate form. Are there any?

src/com/goide/ui/GoConfigurable.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  public Icon getIcon() {
+    return GoIcons.ICON;
+  }
+
+  @NotNull
+  @Override
+  public String getHelpTopic() {
+    return "reference.settingsdialog.go";
```
zolotov: 1. Doesn't make sense
2. What 'reference' means in this id?

src/com/goide/ui/GoConfigurable.java
zolotov: It's more compact. Extracting some forms to separate class makes sense when you're going to reuse this form or if the form is quite complicated and have some non-trivial logic in initialization/disposing parts.

zolotov: Downside is that the configurable looks redundant, it delegates all its logic-holder methods to another class and doesn't contains any useful parts.

src/com/goide/ui/GoConfigurable.java
```diff
@@ -0,0 +1,89 @@
...
+    configurableForm.reset(getApplicationSettings().getState());
+  }
+
+  @Override
+  public void disposeUIResources() {
+    configurableForm = null;
+  }
+
+  private static GoApplicationSettings getApplicationSettings() {
+    return GoApplicationSettings.getInstance();
```
zolotov: mmm... replace one static method with another, looks awful :(

src/com/goide/ui/GoConfigurable.java
dlsniper: I'll change the implementation to inline the form generation, thanks for the details.

dlsniper:
I've fixed the reminder of feedback. Unfortunately here's where my Java limits show off. Is there any place where I could read about how to build interfaces with Java without using the UI Designer?
Currently the UI looks like this: 
snapshot16

Also, the problem remains that on my computer, I can't run gofmt or goimports to environment problems.
May I suggest adding the gosdk to the path (there's a method for this already in the PR) as well as asking the user where ```goinstall``` should saves the built files? This way we could tell where ```goimports``` is placed under.

zolotov:
> I've fixed the reminder of feedback. Unfortunately here's where my Java limits show off. 

Nevermind, I'll clean it up.

> Is there any place where I could read about how to build interfaces with Java without using the UI Designer?

UI Designer is preferable way to do forms. Anyway some official docs might be useful http://docs.oracle.com/javase/tutorial/uiswing/

> May I suggest adding the gosdk to the path (there's a method for this already in the PR)

Ok. Method is ok, just tidy an implementation.

> asking the user where go install should saves the built files

I didn't understand what this is about.

src/com/goide/util/GoExecutor.java
```diff
@@ -120,6 +124,12 @@ public GoExecutor addParameters(@NotNull String... parameters) {
   }
 
   @NotNull
+  public GoExecutor addProcessListener(final ProcessListener listener) {
```
zolotov: Unused

src/com/goide/configuration/GoSettingsConfigurableProvider.java
```diff
@@ -0,0 +1,98 @@
...
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.Nls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import javax.swing.*;
+import java.awt.*;
+import java.util.List;
+
+public class GoSettingsConfigurableProvider extends ConfigurableProvider {
```
zolotov: Why do you need this?

src/com/goide/configuration/GoSettingsConfigurableProvider.java
```diff
@@ -0,0 +1,98 @@
...
+    return goConfigurable();
+  }
+
+  private static Configurable goConfigurable() {
+    return new CompositeConfigurable<UnnamedConfigurable>() {
+
+      @Nullable
+      @Override
+      public JComponent createComponent() {
+        final List<UnnamedConfigurable> configurables = getConfigurables();
```
zolotov: and this

src/com/goide/configuration/GoSettingsConfigurableProvider.java
```diff
@@ -0,0 +1,98 @@
...
+        final GridLayoutManager layoutManager = new GridLayoutManager(configurables.size() + 1, 1, new Insets(0, 0, 0, 0), -1, -1);
+        final JPanel rootPanel = new JPanel(layoutManager);
+        final Spacer spacer = new Spacer();
+        rootPanel.add(spacer, new GridConstraints(configurables.size(), 0, 1, 1, GridConstraints.ANCHOR_SOUTH,
+                                                  GridConstraints.FILL_HORIZONTAL, GridConstraints.SIZEPOLICY_FIXED,
+                                                  GridConstraints.SIZEPOLICY_FIXED, null, null, null));
+
+        for (int i = 0; i < configurables.size(); i++) {
```
zolotov: and all of this. Completely useless copy-paste

src/com/goide/ui/GoSettingsConfigurable.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Override
+  public void apply() throws ConfigurationException {
+    PropertiesComponent.getInstance().setValue(GO_ONSAVE_FMT_KEY, String.valueOf(goFmtOnSave.isSelected()));
+    PropertiesComponent.getInstance().setValue(GO_ONSAVE_IMPORTS_KEY, String.valueOf(goimportsOnSave.isSelected()));
+  }
+
+  @Override
+  public void reset() {
+    Boolean goFmt = PropertiesComponent.getInstance().getBoolean(GO_ONSAVE_FMT_KEY, true);
```
zolotov: These properties are not some 'flags' for defining IDEA behaviour. It's real user settings. Please bring back the service that has semantic value and can be exportable.

utils/src/com/goide/GoEnvironmentUtil.java
```diff
@@ -35,6 +38,10 @@ public static String getBinaryFileNameForPath(@NotNull String path) {
     return SystemInfo.isWindows ? resultBinaryName + ".exe" : resultBinaryName;
   }
 
+  public static File getBinaryFilePathFromEnvironment(@NotNull String binaryName) {
```
zolotov: @Nullable

src/com/goide/formatter/GoOnSaveFormatter.java
```diff
@@ -0,0 +1,99 @@
...
+    if (psiFile == null) {
+      return;
+    }
+
+    Boolean goFmt = PropertiesComponent.getInstance().getBoolean(GO_ONSAVE_FMT_KEY, true);
+    Boolean goImports = PropertiesComponent.getInstance().getBoolean(GO_ONSAVE_IMPORTS_KEY, false);
+
+    final Project finalProject = project;
+    if (goImports) {
+      ApplicationManager.getApplication().invokeLater(new Runnable() {
```
zolotov: Why do you need invokeLater?

dlsniper:
Closing this as it's too painful do to it right now :) Also, there's a known workaround using File Watcher plugin to achieve the same thing.

dooferlad:
Using File Watcher, it will re-format my code as I insert a couple of blank lines as I start to insert a new function because IDEA is saving my file, it would seem, continuously. Is there a way of stopping that?

zolotov:
@dooferlad that's yet another reason why continuously reformatting is bad idea. Can anybody explain why do it's needed?

dlsniper:
@zolotov that's the behavior from either GoSublime or vim or other editors where people are used to their code to be formatted on save.
The way to fix this would be just to have the IDE formatter work, which it's way above what I know how to do right now, sorry.

zolotov:
@dlsniper I see, my vote for proper IDE formatters then. What should be done for making IDE formatter work? I remember the non-trivial comments formatting issue. What else?

dlsniper:
I've started the PR here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1363 with the test sources from the Go SDK for the formatting part. I think that could be a good base to have the tests for the formatter. Should I maybe just create a branch in the repository with the tests instead of a PR?

dooferlad:
@zolotov @dlsniper I agree that proper IDE formatters are the right thing to do. That said, goimports and gofmt are perfect references for what code should look like. I work on a project where we won't land code that gives gofmt errors, so my needs are quite strict.

zolotov:
> gives gofmt errors, so my needs are quite strict.

As a possible workaround you can run `go fmt` on commit, there is a checkbox in commit dialog.

> Should I maybe just create a branch in the repository with the tests instead of a PR?

Yes, I believe it's more convenient

dlsniper:
@dooferlad see this for the option @zolotov mentioned: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1215#issuecomment-72361111

PR #1367, opened 2015-02-24
Show func and method signature in quick doc (fixes #1234)

### LINKED ISSUE #1234: [1.0] Show function arguments and return value in quick doc
Quick documentation should show the function / method full signature as the first line in the output, followed by the rest of the documentation.

Original:
Currently shown only to the function documentation.

dlsniper:
This works only if the documentation is present at the moment.

### DIFF
--- a/src/com/goide/GoDocumentationProvider.java
+++ b/src/com/goide/GoDocumentationProvider.java
@@ -44,7 +44,9 @@ public class GoDocumentationProvider extends AbstractDocumentationProvider {
       Collection<PsiElement> children = PsiTreeUtil.findChildrenOfType(topLevel, element.getClass());
       boolean alone = children.size() == 1 && children.iterator().next().equals(element);
       List<PsiComment> comments = getPreviousNonWsComment(alone ? topLevel : element);
-      if (!comments.isEmpty()) return getCommentText(comments);
+      String result = getSignature(element);
+      if (!comments.isEmpty()) result += getCommentText(comments);
+      return result.length() > 0 ? "<pre>" + result + "</pre>" : "";
     }
     else if (element instanceof PsiDirectory) {
       String comments = getPackageComment(((PsiDirectory)element).findFile("doc.go"));
@@ -60,7 +62,7 @@ public class GoDocumentationProvider extends AbstractDocumentationProvider {
       // todo: remove after correct stubbing (comments needed in stubs)
       GoPackageClause pack = PsiTreeUtil.findChildOfType(file, GoPackageClause.class);
       List<PsiComment> comments = getPreviousNonWsComment(pack);
-      if (!comments.isEmpty()) return getCommentText(comments);
+      if (!comments.isEmpty()) return "<pre>" + getCommentText(comments) + "</pre>";
     }
     return null;
   }
@@ -87,7 +89,7 @@ public class GoDocumentationProvider extends AbstractDocumentationProvider {
 
   @NotNull
   private static String getCommentText(@NotNull List<PsiComment> comments) {
-    return "<pre>" + StringUtil.join(ContainerUtil.map(comments, new Function<PsiComment, String>() {
+    return StringUtil.join(ContainerUtil.map(comments, new Function<PsiComment, String>() {
       @Override
       public String fun(@NotNull PsiComment c) {
         IElementType type = c.getTokenType();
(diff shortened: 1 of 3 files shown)


src/com/goide/GoDocumentationProvider.java
```diff
@@ -86,8 +86,8 @@ private static String getPackageComment(@Nullable PsiFile file) {
   }
 
   @NotNull
-  private static String getCommentText(@NotNull List<PsiComment> comments) {
```
ignatov: Why it's called getCommentText and can return a signature?

dlsniper:
@ignatov I've updated the PR with proper handling for generating the signature and I've also changed it to retrieve the function signature regardless if other comments are present or not.

dlsniper:
And I've fixed the tests as well for it.

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,25 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
...
-    }), "<br/>") + "</pre>";
+    }), "<br/>");
+  }
+
+  private static String getSignature(PsiElement element) {
+    if (element instanceof GoFunctionDeclaration) {
+      return "func " +
+             ((GoFunctionDeclaration)element).getIdentifier().getText() +
+             ((GoFunctionDeclaration)element).getSignature().getText() +
+             "<br/><br/>";
```
ignatov: Why two `<br/>`?

dlsniper: I was looking at the screenshot in the comment here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1234#issuecomment-72492954 and trying to give it a distinction in relation to the documentation.
Should I rather bold the the function signature?

ignatov: Please attach a screenshot with current situation.

Sergey Ignatov

dlsniper: Please see below:

snapshot6

linniksa:
Maybe add signature to &lt;b> tag and format doc and signature by 2 &lt;pre> tags istead of br:

linniksa:
Also want to point out that the current implementation of parsing comments not the same with godoc in many points. I think need create issue do work on rewrite.
but so far I have too little time for active participation.

ignatov:
@dlsniper @linniksa The different start offset looks unpleasant.

dlsniper:
@linniksa I think this is true for documentation that is provided via ```doc.go``` inside a package.
At least that's where I've seen issues with the documentation. Extracting a function doc from a ```doc.go``` would be very hard I think as it has no structure. The only thing that I can think for that would be to provide a tool window with all the package doc displayed in it.

linniksa:
@ignatov simple way is add space to signature at this time
godoc strip same whitespaces prefix form all lines, and if line contains more whitespace characters wrap it in to block (pre)

dlsniper:
I've changed it to something like this:

snapshot7

ignatov:
What're thinking about multiline declarations?

linniksa:
at this time i can make comment with first tab char, or 2 spaces or with no spaces
and without complete rewrite of comments processing (for example we do not need add <br> of each line and place all comment in pre tag by default) we cannont get "right" result in that cases

linniksa:
@dlsniper 
a this time this nit show signature for interface method

dlsniper:
@ignatov we now have support for multi-line defined functions (maybe not in a nice way, http://9gag.com/gag/aLPD0Xg?ref=fbp) 
snapshot8

src/com/goide/GoDocumentationProvider.java
```diff
@@ -44,7 +44,14 @@ public String generateDoc(PsiElement element, PsiElement originalElement) {
       Collection<PsiElement> children = PsiTreeUtil.findChildrenOfType(topLevel, element.getClass());
       boolean alone = children.size() == 1 && children.iterator().next().equals(element);
       List<PsiComment> comments = getPreviousNonWsComment(alone ? topLevel : element);
-      if (!comments.isEmpty()) return getCommentText(comments);
+      String result = "<pre>" + getSignature(element);
+      result = result.
+        replaceAll("\n", "").
+        replaceAll("\\s+", " ").
+        replaceAll("\\(\\s", "(").
+        replaceAll(",\\s\\)", ")");
```
ignatov: It's not a nice way to create a one-liners. Please use PSI for this. See GoParameterInfoHandler.

dlsniper: Ah, you are right, I've forgot to check where we do that. Thanks.

dlsniper:
I've updated the PR to fix the previous comments and added support for interface signatures as well (good find @linniksa, thank you ;) )

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,64 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
...
+
+      if (signatureParameters.length() > 0) {
+        result += " (" + signatureParameters + ")";
+      }
+    }
+
+    return result + "</b><br/>";
+  }
+  
+  private static String getParametersAsString(GoParameters parameters) {
```
ignatov: Can we reuse some code from GoParameterInfoHandler?

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,64 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
         }
         return text;
       }
-    }), "<br/>") + "</pre>";
+    }), "<br/>");
+  }
+
+  @NotNull
+  private static String getSignature(PsiElement element) {
```
ignatov: Please annotate your code via `Infer nullity` action.

src/com/goide/GoDocumentationProvider.java
dlsniper: The whole building part should be reusable, that's why I extracted it in a separate method.
The issue is that GoParameterInfoHandler::updatePresentation() needs some rework in order to properly signal where the current parameter is, which unfortunately is tightly coupled with building the parameter representation.
I'll look into it.

zolotov: As an example you can use https://github.com/JetBrains/intellij-community/blob/master/java/java-impl/src/com/intellij/codeInsight/hint/api/impls/AnnotationParameterInfoHandler.java, take a look at static `updateUIText` method.

linniksa:
@dlsniper 
result with unnamed parameters not shown:
```go
func Foo() string {

}

func Bar() (string, error) {

}
```

dlsniper:
@linniksa thanks. I'll work on the PR to update it with further tests to cover all scenarios and hopefully it will be done in time for the next alpha release.

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,64 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
...
+    if (element instanceof GoNamedSignatureOwner) {
+      identifier = ((GoNamedSignatureOwner)element).getIdentifier();
+    }
+    GoSignature signature = ((GoSignatureOwner)element).getSignature();
+
+    if (identifier == null && signature == null) {
+      return "";
+    }
+
+    String result = " <b>func " + (identifier != null ? identifier.getText() : "") + "(";
```
ignatov: Looks like you've forgot about methods.

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,64 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
...
+      String signatureParameters = getParametersAsString(signature.getResult().getParameters());
+
+      if (signatureParameters.length() > 0) {
+        result += " (" + signatureParameters + ")";
+      }
+    }
+
+    return result + "</b><br/>";
+  }
+
```
ignatov: @NotNull?

dlsniper: I've used the ```Infer nullity``` action and that's how it annotated the code.

src/com/goide/GoDocumentationProvider.java
dlsniper: I've took the decision to not display the receiver of the function as I don't think it's relevant for the user to know that information. I've tested in practice both with and without it before committing it like this.

dlsniper:
I'd need some help with the tests as I can't figure out what I've done wrong compared with the other examples. Thank you.

zolotov:
Obviously you have no reference at the caret. And it shouldn't be there actually, it's function declaration.

dlsniper:
I've fixed the tests and the rest of the feedback on this one. Unless there's anything that I'm missing, this can be merged as well. Thanks.

zolotov:
I'm going to merge this. But there are three pieces of code that do similar things:
1. GoDocumentationProvider#getParametersAsString
2. GoParameterInfoHandler#updatePresentation
3. GoCompletionUtil#createFunctionOrMethodLookupElement – calculating paramTest

Could you handle it please?

dlsniper:
Sure, I'll create a separate PR to unify those, thanks

src/com/goide/GoDocumentationProvider.java
```diff
@@ -102,6 +104,78 @@ else if (type == GoParserDefinition.MULTILINE_COMMENT) {
...
+        String separator = isVariadic ? " ..." : " ";
+        paramPresentations.add(paramDefinition.getText() + separator + paramDeclaration.getType().getText());
+      }
+      if (paramDeclaration.getParamDefinitionList().size() == 0) {
+        paramPresentations.add(paramDeclaration.getType().getText());
+      }
+    }
+
+    StringBuilder builder = new StringBuilder();
+    for (int i = 0; i < paramPresentations.size(); ++i) {
```
zolotov: StringUtil#join?

PR #1388, opened 2015-02-26
add highlighting for hex & oct literals

jakecoffman:
Hope this is correct!


src/com/goide/GoAnnotator.java
```diff
@@ -40,6 +40,9 @@ public void annotate(@NotNull PsiElement o, @NotNull AnnotationHolder holder) {
         o.putUserData(GoReference.IMPORT_USERS, ContainerUtil.<PsiElement>newArrayListWithCapacity(0));
       }
     }
+    else if (o instanceof GoLiteral) {
+      setHighlighting(o, holder, GoSyntaxHighlightingColors.NUMBER);
+    }
```
ignatov: From the grammar (go.bnf):
```
Literal ::=
    int
  | float
  | floati
  | decimali
  | hex
  | oct
  | imaginary
  | rune
  | StringLiteral
  | char
```
So, you need to check that it's a hex or oct, but not a char or a rune.

jakecoffman:
Is that better @ignatov?

ignatov:
Thanks

zolotov:
Hey, what about tests?;)

jakecoffman:
I see GoHighlightingTest but can't figure out how to test for non-errors like hex highlighting. If you could give me a hint I will write some tests.

zolotov:
@jakecoffman I've just pushed a simple test for that (GoHighlightingAnnotatorTest). TestData is placed at "testData/colorHighlighting".

PR #1404, opened 2015-03-02
Use container instances for TravisCI

dlsniper:
According to [the docs](http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/) it should be faster to build like this.


zolotov:
Build broken, please investigate. I think travis part isn't required to be reviewed

PR #1405, opened 2015-03-02
Revert "Use container instances for TravisCI"

dlsniper:
Reverts go-lang-plugin-org/go-lang-idea-plugin#1404


dlsniper:
@zolotov there's an error that Travis is giving in the previous PR.

Issue #1413, opened 2015-03-04
1.0.0.alpha#164 Cannot Run Go Application configuration


jtseatwave:
*** I removed the full paths from screenshots and output below ***
I am unable to run the go application.  I can run ```go build``` from the project directory and it creates a go executable.  However I get the below output from the run command.  It appears that the plugin is only building the main source file and not the whole directory

![screen shot 2015-03-04 at 13 52 26](https://cloud.githubusercontent.com/assets/6087658/6484678/c267ca88-c275-11e4-8df0-6ae86dee0325.png)

```
# command-line-arguments
# command-line-arguments
.../run_imports.go:14: undefined: VerifyVenueAddresses
.../run_imports.go:18: undefined: Importer
.../run_imports.go:20: undefined: Importer
.../run_imports.go:22: undefined: StartImport
.../run_imports.go:23: undefined: StartImport
.../run_imports.go:24: undefined: StartImport
.../run_imports.go:29: undefined: Importer
.../run_imports.go:32: undefined: WriteImports
.../run_imports.go:37: undefined: Importer```

**Comment by dlsniper** [comment]
Please select the full path to the file not the relative path.

**Comment by jtseatwave** [comment]
I just removed the full path to hide the project name.

**Comment by zolotov** [comment]
> I can run go build from the project directory and it creates a go executable

What happens if you run `go build run_imports.go` in console?

**Comment by jtseatwave** [comment]
```
go build run_imports.go
# command-line-arguments
./run_imports.go:14: undefined: VerifyVenueAddresses
./run_imports.go:18: undefined: Importer
./run_imports.go:20: undefined: Importer
./run_imports.go:22: undefined: StartImport
./run_imports.go:23: undefined: StartImport
./run_imports.go:24: undefined: StartImport
./run_imports.go:29: undefined: Importer
./run_imports.go:32: undefined: WriteImports
./run_imports.go:37: undefined: Importer
```

**Comment by zolotov** [comment]
And what if you run `go build -i run_imports`?

**Comment by jtseatwave** [comment]
same output

**Comment by zolotov** [comment]
Ok, thank you. And the last one: `go build -n`. Just want to know what exactly go build does if there are no any parameter.

**Comment by jtseatwave** [comment]
```
go build -n .

#
# [directory]/catalogue-import/couchbase
#

mkdir -p $WORK/[directory]/catalogue-import/couchbase/_obj/
[...]

PR #1414, opened 2015-03-05
Naive fix for #1413

ccustine:
I have been running this for a few days to get go application run configs to work.  I'm sure this could be done much more elegantly and the run configs seem ready for a refactoring but this works for me so I thought it might be useful.  I have some other comments for run configs that I will add directly to the thread on #1413.

### DIFF
--- a/src/com/goide/runconfig/GoRunConfigurationBase.java
+++ b/src/com/goide/runconfig/GoRunConfigurationBase.java
@@ -48,6 +48,7 @@ public abstract class GoRunConfigurationBase<RunningState extends GoRunningState
   private static final String PARAMETERS_NAME = "parameters";
   private static final String PASS_PARENT_ENV = "pass_parent_env";
 
+  @NotNull private String myFilePath = "";
   @NotNull private String myWorkingDirectory = "";
   @NotNull private String myParams = "";
   @NotNull private Map<String, String> myCustomEnvironment = ContainerUtil.newHashMap();
@@ -194,4 +195,14 @@ public abstract class GoRunConfigurationBase<RunningState extends GoRunningState
   public void setWorkingDirectory(@NotNull String workingDirectory) {
     myWorkingDirectory = workingDirectory;
   }
+
+  @NotNull
+  public String getFilePath() {
+    return myFilePath;
+  }
+
+  public void setFilePath(@NotNull String filePath) {
+    myFilePath = filePath;
+  }
+
 }
--- a/src/com/goide/runconfig/GoRunConfigurationWithMain.java
+++ b/src/com/goide/runconfig/GoRunConfigurationWithMain.java
@@ -38,11 +38,9 @@ import org.jetbrains.annotations.NotNull;
 public abstract class GoRunConfigurationWithMain<T extends GoRunningState> extends GoRunConfigurationBase<T> {
   private static final String FILE_PATH_NAME = "file_path";
   
-  @NotNull private String myFilePath = "";
-
   public GoRunConfigurationWithMain(String name, GoModuleBasedConfiguration configurationModule, ConfigurationFactory factory) {
     super(name, configurationModule, factory);
-    myFilePath = getWorkingDirectory();
+    setFilePath(getWorkingDirectory());
   }
 
   @Override
@@ -50,15 +48,15 @@ public abstract class GoRunConfigurationWithMain<T extends GoRunningState> exten
     super.readExternal(element);
     String filePathValue = JDOMExternalizerUtil.getFirstChildValueAttribute(element, FILE_PATH_NAME);
     if (filePathValue != null) {
-      myFilePath = filePathValue;
+      setFilePath(filePathValue);
     }
   }
 
   @Override
   public void writeExternal(Element element) throws WriteExternalException {
     super.writeExternal(element);
-    if (StringUtil.isNotEmpty(myFilePath)) {
-      JDOMExternalizerUtil.addElementWithValueAttribute(element, FILE_PATH_NAME, myFilePath);
+    if (StringUtil.isNotEmpty(getFilePath())) {
+      JDOMExternalizerUtil.addElementWithValueAttribute(element, FILE_PATH_NAME, getFilePath());
     }
   }
 
@@ -83,12 +81,4 @@ public abstract class GoRunConfigurationWithMain<T extends GoRunningState> exten
     }
   }
 
-  @NotNull
-  public String getFilePath() {
-    return myFilePath;
-  }
-
-  public void setFilePath(@NotNull String filePath) {
-    myFilePath = filePath;
-  }
 }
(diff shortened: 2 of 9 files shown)


src/com/goide/runconfig/GoRunConfigurationBase.java
```diff
@@ -194,4 +195,14 @@ public String getWorkingDirectory() {
   public void setWorkingDirectory(@NotNull String workingDirectory) {
     myWorkingDirectory = workingDirectory;
   }
+
+  @NotNull
+  public String getFilePath() {
```
zolotov: Why do you need it in base class? Why do you need duplicates of this methods in GoTestRunConfiguration?

src/com/goide/runconfig/GoRunConfigurationWithMain.java
```diff
@@ -38,11 +38,9 @@
 public abstract class GoRunConfigurationWithMain<T extends GoRunningState> extends GoRunConfigurationBase<T> {
   private static final String FILE_PATH_NAME = "file_path";
   
-  @NotNull private String myFilePath = "";
-
   public GoRunConfigurationWithMain(String name, GoModuleBasedConfiguration configurationModule, ConfigurationFactory factory) {
     super(name, configurationModule, factory);
-    myFilePath = getWorkingDirectory();
+    setFilePath(getWorkingDirectory());
```
zolotov: You have defined the field in base class, so store the field there and remove storing duplicates in concrete run configurations

src/com/goide/runconfig/application/GoApplicationRunConfiguration.java
```diff
@@ -50,4 +51,10 @@ protected ModuleBasedConfiguration createInstance() {
   protected GoApplicationRunningState newRunningState(@NotNull ExecutionEnvironment env, @NotNull Module module) {
     return new GoApplicationRunningState(env, module, this);
   }
+
+  @Override
```
zolotov: Redundant method

src/com/goide/runconfig/application/GoApplicationRunConfiguration.java
```diff
@@ -17,32 +17,33 @@
...
+import com.intellij.execution.configurations.RuntimeConfigurationException;
 import com.intellij.execution.runners.ExecutionEnvironment;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.options.SettingsEditor;
 import com.intellij.openapi.project.Project;
 import org.jetbrains.annotations.NotNull;
 
-public class GoApplicationConfiguration extends GoRunConfigurationWithMain<GoApplicationRunningState> {
-  public GoApplicationConfiguration(Project project, String name, @NotNull ConfigurationType configurationType) {
+public class GoApplicationRunConfiguration extends GoRunConfigurationBase<GoApplicationRunningState> {
```
zolotov: GoRunConfigurationWithMain now has the only implementation and it seems to be redundant.

src/com/goide/runconfig/GoRunConfigurationBase.java
ccustine: The change to move filePath up to the base is probably unnecessary now.  I had originally used this to point to the app root directory until I looked up the source root and eliminated the file path from the run app config entirely.  Also, I totally overlooked the TestRun configs so I will fix these tomorrow.

Like I said, totally naive approach and I have never done Idea plugin work before so there are probably much better ways to do this later.

src/com/goide/runconfig/application/GoApplicationRunningState.java
```diff
@@ -52,12 +54,13 @@ protected ProcessHandler startProcess() throws ExecutionException {
     }
     try {
       ProcessOutput processOutput = new ProcessOutput();
+      VirtualFile[] sourceRoots = ModuleRootManager.getInstance(myModule).getSourceRoots(false); //Eventually to support multiple module dependencies
```
zolotov: And after a lot of code this is actually "fix", right? Well, I can't accept it unfortunately. Despite of strange inheritance and duplicating code you've made, here are several problems:

1. Lost an ability to configure main file, so it's impossible to use run configuration in the projects that have several main files.
2. Lost an ability to configure working directory. User can configure it but you will replace it anyway.
3. Source roots. Well, it's just wrong :) Despite of the fact that sourceRoots could be empty (especially in Small IDEs) and you'll get IOOBE in this case. Source roots have no any semantic sense in Go support, first source root could point anywhere, any user can mark any directory or even project root as source root and it doesn't affect (and shouldn't) go-support part, especially it shouldn't affect unconfigurable run configurations.

src/com/goide/runconfig/application/GoApplicationRunConfiguration.java
ccustine: Good point, I will clean all of this up tomorrow.  Thanks for the feedback, and let me know if you have something better in the works. This was all just a quick hack on my part and I can see that there is probably a much more elegant way to do this eventually.

zolotov:
Thank you for help. I see that this is a naive fix, but current implementation is quite naive too and I don't see any reasons fix something in naive way and break other things that works in already implemented naive approach.
I like your point in #1413 about `something like the Java Application run config`, if you have particular thoughts about it (like how it should look like, what user should be able to configurate, how to run go build depends on user's configuration and so on), please share them with me.

src/com/goide/runconfig/application/GoApplicationRunConfiguration.java
zolotov: I don't have any in works yet, I do the plugin just a day in a week. But I'd definitely like to cooperate with you and think of something more elegant and covering as much as possible cases solution :)

ccustine:
@zolotov Thanks for the comments, and you are absolutely correct. I think the main thing is that to succesfully `go build -o <tempxecutable>` it seems that you have to run in top package directory with no other arguments (for instance `go build -o tmpexe ./...` and `go build -o tmpexe main.go` fail for different reasons) so `go build -o tmpexe` is the only way I have found to make that work.  My attempt to discover the root based on the selected module was just a hack as I try to learn the plugin sdk :-)

One thing that I have not experimented with is having multiple Go modules in a single project and seeing if the run configurations will work.  I assume this should be supported eventually so I will try to verify soon.

I'm really interested to get all of these scenarios working properly because I have projects with all 3 variations of single main() in file, pkg main and main() in application, and the sub directory approach to getting multiple executables.  I will give it some more thought as I learn this plugin sdk and see what I can come up with.

zolotov:
@ccustine thanks, looking forward to your comments. It's really great that you have experience with different workflows in go, because you know most of main contributors are not really go-guys so we have huge lack in understanding workflow questions.

About SDK, instead of source roots in this plugin we usually operate with following entities: 
– sdk ($GOROOT), that you can retrieve with `GoSdkService.getInstance().getSdkHomePath()`
– libraries ($GOPATH), that you can get their 'src' directories with `GoSdkUtil.getGoPathsSources`
– importpaths (GoFile.getImportPath() will return path to file relative to $GOPATH or $GOROOT) and package (GoFile.getPackage())

Also GoSdkUtil contains some handy methods, maybe you'll find them helpful.

ccustine:
I'm going to close this PR for now.  I am working on another approach which should be better and I will just open a new PR because it might be a few days.

zolotov:
> I am working on another approach which should be better and I will just open a new PR because it might be a few days.

Looking forward to it!

PR #1458, opened 2015-03-19
[WIP] Add default folding for constants (fixes #1393)

### LINKED ISSUE #1393: [1.maybe] Display const value rather than name
```go
package main

const demo = "Demo text here"
func main() {
	println(demo) // IDE should display Demo text here not demo by default
}
```

Question: should I use a FoldingBuilder like in the example from https://confluence.jetbrains.com/display/IntelliJIDEA/Folding+Builder to solve this?

dlsniper:
Hi,

Before merging, this needs some consideration to it.
It can be annoying to fold the const into its values by default when:
- the const value is in non-decimal representation
- cost value is like this: https://github.com/golang/go/blob/master/src/time/time.go#L80 or https://github.com/golang/go/blob/master/src/time/time.go#L449

As such, I'd like to play a bit more with it and do const folding to values only when the value is a string to begin with then expand if other use-cases are found.

Thus, the two questions I have:
- how do I get the const value?
- how wrong is the code I've done?

Thank you :)

### DIFF
--- a/src/com/goide/editor/GoFoldingBuilder.java
+++ b/src/com/goide/editor/GoFoldingBuilder.java
@@ -19,6 +19,7 @@ package com.goide.editor;
 import com.goide.GoParserDefinition;
 import com.goide.GoTypes;
 import com.goide.psi.*;
+import com.goide.psi.impl.GoReference;
 import com.intellij.codeInsight.folding.CodeFoldingSettings;
 import com.intellij.lang.ASTNode;
 import com.intellij.lang.folding.FoldingBuilderEx;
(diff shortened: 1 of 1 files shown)


ignatov:
In my opinion it's really not necessary.

dlsniper:
Hmm, then do you think it would be possible to change maybe the tooltip for consts (the one that appears when you ctrl + hover a const) to display it's value? sometimes I'd really like to know what's under the value without getting the value of it.
It's particularly useful if your consts are strings with placeholders and you use fmt.Sprintf for example.

ignatov:
Tooltips are much better. BTW, you can press Alt+Space on Mac or Ctrl+Shift+I on Windows/Linux and see an implementation.
https://www.dropbox.com/s/ny8b3pifrmqlnzn/Screenshot%202015-03-22%2015.48.28.png?dl=0

PR #1472, opened 2015-03-21
Parameters help does not work for method calls made on struct fields

horkhe:
Parameter help is not provided for calls that are made on struct fields, e.g. `a.b.f(1, "foo")`. This PR fixes this deficiency.


ignatov:
Thanks

PR #1478, opened 2015-03-24
Switch back to GoFoldingBuilder from master branch

xian:
It supports folding anonymous functions, and seems to generally work pretty well.


zolotov:
What exactly was introduced? Where is tests for new functionality? Have you run at least existing tests?

Issue #1512, opened 2015-04-01
GAE Windows installations should search for 'goapp.bat' instead of 'goapp.exe'


nisckis:
GAE Windows installations should search for `goapp.bat` instead of `goapp.exe`.

zolotov:
Fixed by @nisckis

PR #1513, opened 2015-04-01
GAE Windows installations should search for 'goapp.bat' instead of 'goapp.exe'

nisckis:
GAE Windows installations should search for `goapp.bat` instead of `goapp.exe`

Related Issue is #1512

### DIFF
--- a/src/com/goide/sdk/GoSdkService.java
+++ b/src/com/goide/sdk/GoSdkService.java
@@ -12,6 +12,7 @@ import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VfsUtilCore;
 import com.intellij.openapi.vfs.VirtualFile;
+import com.intellij.util.PathUtil;
 import org.jetbrains.annotations.Contract;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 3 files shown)


src/com/goide/sdk/GoSdkService.java
```diff
@@ -74,7 +75,7 @@ public static String getGoExecutablePath(@Nullable String sdkHomePath) {
         // gcloud and standalone installations
         return sdkHomePath.endsWith(GoConstants.GCLOUD_APP_ENGINE_DIRECTORY_PATH)
                ? FileUtil.join(StringUtil.trimEnd(sdkHomePath, GoConstants.GCLOUD_APP_ENGINE_DIRECTORY_PATH), "bin", executable)
-               : FileUtil.join(sdkHomePath, executable);
+               : FileUtil.join(sdkHomePath, SystemInfo.isWindows ? StringUtil.replace(executable, ".exe", ".bat") : executable);
```
zolotov: What about gcloud installation?

nisckis: I don't know, I'll install and test it.

zolotov: Ok, thanks.

nisckis: Sorry for the delay, gcloud windows installer took like forever, and `gcloud components update gae-go`wasn't fast either.

OK, this is amazing, GAE for Windows uses `goapp.bat`but gcloud for Windows uses `goapp.cmd`.

For gcloud it should be `? FileUtil.join(StringUtil.trimEnd(sdkHomePath, GoConstants.GCLOUD_APP_ENGINE_DIRECTORY_PATH), "bin", SystemInfo.isWindows ? StringUtil.replace(executable, ".exe", ".cmd") : executable)`, so the return statement would be
```java
        return sdkHomePath.endsWith(GoConstants.GCLOUD_APP_ENGINE_DIRECTORY_PATH)
               ? FileUtil.join(StringUtil.trimEnd(sdkHomePath, GoConstants.GCLOUD_APP_ENGINE_DIRECTORY_PATH), "bin", SystemInfo.isWindows ? StringUtil.replace(executable, ".exe", ".cmd") : executable)
               : FileUtil.join(sdkHomePath, SystemInfo.isWindows ? StringUtil.replace(executable, ".exe", ".bat") : executable);
```

Should I create a new pull request which includes gae and gcloud for Windows correction?

zolotov: > OK, this is amazing, GAE for Windows uses goapp.batbut gcloud for Windows uses goapp.cmd.

Haha :) That's really funny. 

> Should I create a new pull request which includes gae and gcloud for Windows correction?

Just push correction in your fork-master, the PR should be updated automatically (hope so).

And one note. There is a line 
```
String executable = GoEnvironmentUtil.getBinaryFileNameForPath(GoConstants.GAE_EXECUTABLE_NAME);
```

And it looks strange to append executable with "exe" (getBinaryNameForPath does it) and just after that replace "exe" with something else. Please inline executable value inside each FileUtil.join invocation. Maybe it makes sense to create util methods `getBinaryOrCmdFileName` and `getBinaryOrBatFileName`

zolotov: And feel free to add your name to contributors.md

nisckis:
It took me more time than I expected because I've found one gcloud sdk home directory selection that was not working as expected.

If you install gcloud under `C:\google\cloudSDK\` you will have three goapp files:
1. C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine\goroot\bin\goapp.exe
2. C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine\goapp.bat
3. C:\google\cloudSDK\google-cloud-sdk\bin\goapp.cmd

If we select `C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine\goroot` as SDK home it will search for `C:/google/cloudSDK/google-cloud-sdk\bin\goapp.cmd` and everything works as expected.

If we select `C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine` as SDK home, it will search for:
- `C:/google/cloudSDK/google-cloud-sdk/platform/google_appengine\bin\go.exe` which doesn't exist.
- `C:/google/cloudSDK/google-cloud-sdk\bin\goapp.cmd` which exists and everything works as expected.

If we select `C:\google\cloudSDK\google-cloud-sdk\platform\` as SDK home, it will search for `C:/google/cloudSDK/google-cloud-sdk/platform\bin\go.exe` twice, it will fail and we will receive the message `The selected directory is not a valid home for Go SDK` as expected.

If we select `C:\google\cloudSDK\google-cloud-sdk\` it will test if `C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine` ends with `/platform/google_appengine`, which doesn't match so instead of looking for `goapp.cmd` it will search for `goapp.bat`.
By the way, it would match because of `2`, but I preferred transforming the path to a system independent one so it will search the file "gcloud way".

Finally if we select `C:\google\cloudSDK\` as SDK home, it will search for `C:/google/cloudSDK\bin\go.exe` twice, it will fail and we will receive the message `The selected directory is not a valid home for Go SDK` as expected.

zolotov:
Thanks for investigation!

> If we select C:\google\cloudSDK\google-cloud-sdk\platform\google_appengine\goroot as SDK home it will search for C:/google/cloudSDK/google-cloud-sdk\bin\goapp.cmd and everything works as expected.

It's the only possible way, other directories will be adjusted to one of these paths or won't be accepted as sdk root.

PR #1524, opened 2015-04-02
Introduce different parameter list for go tool and the user app

### LINKED ISSUE #1517: Broken run configuration
Run/Debug configuration stopped working:
screenshot - 010415 - 20 10 12

Looks like command line arguments are all mixed up.

IDEA 14.1.1
Plugin 0.9.242

dlsniper:
There's one thing that this doesn't do properly right now, sent the ```Program arguments``` to the actual executable, I'm still working on that on, should be here soon.

### DIFF
--- a/src/com/goide/appengine/run/GoAppEngineRunningState.java
+++ b/src/com/goide/appengine/run/GoAppEngineRunningState.java
@@ -14,10 +14,11 @@ public class GoAppEngineRunningState extends GoRunningState<GoAppEngineRunConfig
                                  @NotNull GoAppEngineRunConfiguration configuration) {
     super(env, module, configuration);
   }
-  
+
   @Override
   protected GoExecutor patchExecutor(@NotNull GoExecutor executor) throws ExecutionException {
     executor.withParameters("serve");
+    executor.withParameterString(myConfiguration.getGoParams());
     String host = myConfiguration.getHost();
     String port = myConfiguration.getPort();
     if (StringUtil.isNotEmpty(host)) {
--- a/src/com/goide/runconfig/GoBuildingRunner.java
+++ b/src/com/goide/runconfig/GoBuildingRunner.java
@@ -60,11 +60,17 @@ public class GoBuildingRunner extends AsyncGenericProgramRunner {
   protected Promise<RunProfileStarter> prepare(@NotNull ExecutionEnvironment environment, @NotNull RunProfileState state)
     throws ExecutionException {
     final File tmpFile;
+    final String params;
     try {
-      RunnerAndConfigurationSettings settings = environment.getRunnerAndConfigurationSettings();
+      final RunnerAndConfigurationSettings settings = environment.getRunnerAndConfigurationSettings();
       tmpFile = FileUtil.createTempFile(settings != null ? settings.getName() : "application", "go", true);
       if (!tmpFile.setExecutable(true)) {
         throw new ExecutionException("Can't make temporary file executable (" + tmpFile.getAbsolutePath());
+      }
+      if (settings != null && settings.getConfiguration() instanceof GoRunConfigurationBase) {
+        params = ((GoRunConfigurationBase) settings.getConfiguration()).getGoParams();
+      } else {
+        params = "";
       }
     }
     catch (IOException e) {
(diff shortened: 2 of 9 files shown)


zolotov:
looks very complicated, I've made just a little mistake [here|https://github.com/go-lang-plugin-org/go-lang-idea-plugin/commit/72311e6ff842110fed086260e4b51c949d4d801f#diff-1c5c76ad6487f1717502730d3c5b93ddR48] and IMO it's not worth so big changes and introducing several types of executable's parameters.

zolotov:
Adding `build` parameters and `program` parameter in Go Application RC is make sense though.

dlsniper:
Yes, that's my reason. Please don't merge it just yet as there's one more change I need to do.

dlsniper:
Ready for review. I've introduced two different notions to ```GoExecutor```
- GoToolName -> sets the name of the go tool that should be executed. For example a ```go tool``` is ```test```, in which case the executor should be called with ```withGoToolName("test")```
- GoParameterList -> allows setting parameters to the go tool that's being executed

This way the plugin will be flexible in the future to execute things like gccgo builder with minimal changes. Also, it allows users to configure the build properties of the ```build``` and ```run``` tools for example.

zolotov:
I don't really like it. Separating parameters into three different groups (and the only difference is parameters order) for fixing my little mistake is overkill. 
Also GoExecutor is more about preparing environment from context and running anything, not about running go tools. So extra semantic about go tools looks redundant there.

> This way the plugin will be flexible in the future to execute things like gccgo builder with minimal changes. 

I believe it was flexible before as well, the change doesn't bring flexibility improvements.

> Also, it allows users to configure the build properties of the build and run tools for example.

It was allowed before.

And here is the most strange thing. The initial mistake was made because I wanted to make the single point of executor constructing for all running states and I forgot about parameters order. If each `running state` creates their own executor then there won't bugs (hence your changes is not needed): `states` will control parameters order by themselves. But your changes make states to control executor construction anyway (e.g. https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1524/files#diff-2ee2dfc8111803a6fbc76437f38f24eaR42). So one part of your changes make redundant the other part :)

But I really like the idea of separating parameters while building and running for Go Application run configuraiton. I think that it would be nice if you create a separate PR for this, and I'll try to think of neat fix for parameters order bug.

dlsniper:
I see. Should I revert this change then: https://github.com/dlsniper/google-go-lang-idea-plugin/commit/5ba217fac40be796f7dea2aa733673be53abbcb2 ? Then there should be two different parameters.

I've had to introduce the change because I was setting the GoParameter then the order was skewed so instead of running ```go build -h``` it was running ```go -h build```.

zolotov:
> I've had to introduce the change because I was setting the GoParameter then the order was skewed so instead of running go build -h it was running go -h build.

True, but it's bug of running state, not goExecutor's. I'll fix it.

> I see. Should I revert this change then: dlsniper@5ba217f ? Then there should be two different parameters.

https://github.com/dlsniper/google-go-lang-idea-plugin/commit/44f308af069eba9bc59d332675dff211066bba7d this is ok, except `go build` should use string from new field (like `Build parameters`) and `go run` should use common parameters string from RunConfigurationBase just because it's called `Program parameters`.

dlsniper:
I've removed the previous commit.

> except go build should use string from new field (like Build parameters) and go run should use common parameters string from RunConfigurationBase just because it's called Program parameters.

I'm sorry, this is a bit not clear. With this PR there will be two fields available for every command runner, the ```go tool arguments``` and ```Program arguments```. The ```go tool arguments``` should be used for ```go run / go build / go test``` while ```Program arguments``` is added to the program that ```go build``` generates or to ```go run``` after the file / package name. I've tested all the options aside from GDB one.

zolotov:
I'm not sure that 'run file' and 'test' configurations need customizable tool's parameters. Anyway, GoExecutor shouldn't know about this.

dlsniper:
Done. I've now understood what you mean by the fact that the go executor shouldn't know about go tool specific parameters. Hope this is better.

dlsniper:
bump :) (also, can you please check the other PRs? Thank you!)

dlsniper:
Friendly bump for review when you can. This would be useful as I currently want to run the tests with a build tag and I also have flags for the tests themselves. Thank you!

PR #1526, opened 2015-04-02
Introduce FAQ

dlsniper:
Hopefully this will be useful to some.


zolotov:
@dlsniper I believe you can push such things without PR, it's unlikely that I can advice anything useful about docs and faqs, furthermore you're doing it really good.

PR #1551, opened 2015-04-11
Allow build configurations with packages (fixes #1520)

### LINKED ISSUE #1520: Allow running projects composed of multiple files
for instance:
I can run in terminal like this : go run test0.go test1.go
How can I run this in Idea ?

dlsniper:
The only thing that remains to be done, aside from cleaning the interface a bit, is to show the "Package" option only when the dialog is shown for the "go build configuration".


dlsniper:
I think it fixes #1413 as well

dlsniper:
I've rebased and tested this, still works. Please let me know if it's good to merge. Thank you! :)

PR #1552, opened 2015-04-11
Don't fold empty types (fixes #1549)

### LINKED ISSUE #1549: Prevent folding for interface{}
Currently the interface ``` interface{} ``` folds to ``` interface{...} ``` which can be annoying to do as it's misleading.

dlsniper:
I'm not sure if the test quality is good enough but it seems it's passing and testing it in real world works as well.


ignatov:
@dlsniper Please see the corresponding cleanup commit.

dlsniper:
I understand. Thank you!

PR #1553, opened 2015-04-11
[WIP] Highlight field names accordingly (fixes #1534)

### LINKED ISSUE #1534: Coloring Issues
https://www.dropbox.com/s/d7jaqqxv7twfw6t/Screenshot%202015-04-06%2019.57.25.png?dl=0

In that code, in the struct literal, var1 should be highlighted just as it is in the println statement no? It is treated as simply an identifier in the colors and fonts list for go.

Also, why are packages treated as identifiers as well? Should they not have their own colors in the fonts and colors tab? I haven't done any real investigation but I would imagine there are perhaps more issues similar to this which should be resolved. But really, awesome plugin so far!

dlsniper:
Hi.

I know this is not the right solution as I'm changing the "gen" files directly but I'd like some input on how to do this better.
Ideally the ```isPublic()``` method should be present on all types (maybe it's stupid?) so that we can always know when something is public or not.

What do you think? How should I approach this?

### DIFF
--- a/src/com/goide/GoAnnotator.java
+++ b/src/com/goide/GoAnnotator.java
@@ -18,6 +18,7 @@ package com.goide;
 
 import com.goide.highlighting.GoSyntaxHighlightingColors;
 import com.goide.psi.*;
+import com.goide.psi.impl.GoFieldNameImpl;
 import com.goide.psi.impl.GoPsiImplUtil;
 import com.goide.psi.impl.GoReference;
 import com.intellij.lang.annotation.AnnotationHolder;
@@ -25,6 +26,7 @@ import com.intellij.lang.annotation.Annotator;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.editor.colors.TextAttributesKey;
 import com.intellij.openapi.editor.markup.TextAttributes;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
 import com.intellij.util.containers.ContainerUtil;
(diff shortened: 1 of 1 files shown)


ignatov:
Please don't touch the gen sources. Use Grammar-Kit instead of.

dlsniper:
So, should I ```isPublic()``` to ```GoReferenceExpressionBase``` then? Does it make sense? Sorry, that part of how the plugin works is really unclear for me.

ignatov:
Please rework your PR and regen the whole PSI from grammar.

dlsniper:
I've updated this but I'm pretty sure this is not what you meant when you said that I need to update the parser. I've tried to understand where I can add the ```isPublic``` method so that ```GoFieldName``` contains it and I couldn't figure it out. If this will be changed to properly have the parser changes I'll learn from it then. Thank you.

src/com/goide/GoAnnotator.java
```diff
@@ -133,6 +137,12 @@ private static TextAttributesKey getColor(GoFieldDefinition o) {
               : GoSyntaxHighlightingColors.STRUCT_LOCAL_MEMBER;
   }
 
+  private static TextAttributesKey getColor(GoFieldNameImpl o) {
```
ignatov: Why do you use a particular implementation instead of the interface?

src/com/goide/GoAnnotator.java
```diff
@@ -133,6 +137,12 @@ private static TextAttributesKey getColor(GoFieldDefinition o) {
...
+    return StringUtil.isCapitalized(o.getName())
```
ignatov: As far as I understand it's `com.intellij.psi.impl.PsiElementBase#getName`, so I believe that condition will return false.

src/com/goide/GoAnnotator.java
dlsniper: I need this particular one because I don't know how to add a ```isPublic()``` method to ```GoFieldName``` or one of the interfaces that is implemented by it so that it returns ```StringUtil.isCapitalized(o.getName())```.

dlsniper: Also, the implementation has access to the ```getName()``` method which the interface doesn't expose.

PR #1554, opened 2015-04-11
Run golint and go vet as external inspections (fixes #342)

### LINKED ISSUE #342: go lint support
It would be nice if the plugin would support ``` go lint ``` as an external tool for running for running against a package / module / file.

The only issue with it is that it's not always available as it needs to be installed by the user (maybe we should have a: 'Install go lint' option first').

go lint documentation can be found [here](https://github.com/golang/lint).

I've opened this to keep track of it.

Thanks.

dlsniper:


dlsniper:
I don't know how to make the description to properly appear in the settings panel (even if I've left the IDE create the html file).

Ideally, I'll add the go vet inspections as well after this gets merged but I don't really know where to start and do all the abstraction work so that only the settings file "*Inspection" class is created and then the "*ExternalAnnotator" just implements the abstraction and functions like getting the "ProcessOutput" as well as the other cosmetics functions.

I wonder if the abstraction for it could be extracted to the platform itself at some point?

What do you think? Does it make sense?

src/META-INF/plugin.xml
```diff
@@ -26,7 +26,7 @@
     <p>Alpha pre-release of the 1.0.0 version.</p>
     <p>Doesn't contain all the functionality of the 0.9.x branch but has a completely reworked internals.
       It's faster than 0.9.x, refactoring works to some degree and has native support for gopath packages.</p>
-    
+
```
ignatov: No need to reformat the whole file. As for me, it's better to separate formatting/adding a new functional (just for better git history).

dlsniper: Sorry, that's my default setting to remove all trailing spaces from all lines when saving files which seems it's not in this project settings. I'll see if I can change the commit somehow.
Maybe it's a good idea to create PR to cleanup the sources and remove unused imports as well?

dlsniper:
I've pushed go vet as well as it will make the review process simpler / faster. go vet is basically a copy paste of golint with small tweaks for the tool to execute and how to parse the line (plus where to get the line itself).

src/META-INF/plugin.xml
```diff
@@ -69,9 +69,12 @@
     <applicationService serviceImplementation="com.goide.project.GoApplicationLibrariesService"/>
     <projectService serviceImplementation="com.goide.project.GoProjectLibrariesService"/>
     <moduleService serviceImplementation="com.goide.project.GoModuleLibrariesService"/>
-    <projectConfigurable groupId="language" provider="com.goide.configuration.GoConfigurableProvider" dynamic="true" 
+    <projectConfigurable groupId="language" provider="com.goide.configuration.GoConfigurableProvider" dynamic="true"
                          bundle="do.not.touch.this.attribute" />
-    
+
+    <projectService serviceImplementation="com.goide.inspections.external.golint.GoLintSettings"/>
```
zolotov: where is that?

src/com/goide/util/GoExecutor.java
```diff
@@ -277,6 +288,11 @@ private void showOutput(@NotNull OSProcessHandler originalHandler, @NotNull Hist
     }
   }
 
+  public ProcessOutput getProcessOutput() throws ExecutionException {
```
zolotov: It's quite surprising when a getter starts the process.

dlsniper: Agreed. I've renamed it to ```runWithProcessOutput()```.

src/META-INF/plugin.xml
dlsniper: I've forgot to delete it from here.

dlsniper:
I've rebased and squashed this. Please let me know if there's anything else I should do to have this merged. Thank you!

dlsniper:
Ping for this one :-)

zolotov:
Cannot merge it. It just works badly. Mostly it's because it's not supposed by the platform that External Annotator will be invoked on file system. Hope it will be fixed in IDEA 15.

dlsniper:
I see. I was on the phone and Github doesn't display nicely in the web interface. Should I rebase it for merging? Or because of the platform issues it wouldn't get in anyway?

zolotov:
Yes, I meant platform issue as the reason of merging problems. It just doesn't work, highlighting is wrong most of the time. The only chance to merge it is to disable inspection in the fly mode, but it's not that users want either.

dlsniper:
I see, like I said in another ticket, I think it's better to support all the inspections golint has, even if we duplicate functionality as it means we can provide fixes for them for example.

I know I'll go a bit of topic but. 

What's the performance impact, from your experience with the other languages, when we'll have say 30-40 inspections (I think that's a good number considering all the possible inspections from golint and other tickets as well). Is there something that we could to for example run a single function checker that combines the job of 3-4 inspections so that we avoid walking the tree 3-4 times and we do it just once? Or that part is usually not a problem compared to the inspection weight themselves?

zolotov:
> I see, like I said in another ticket, I think it's better to support all the inspections golint has, even if we duplicate functionality as it means we can provide fixes for them for example.

Yes, it's the preferable way. Besides quickfixes we can allow user to disable/suppress some of them as well.

> What's the performance impact, from your experience with the other languages, when we'll have say 30-40 inspections (I think that's a good number considering all the possible inspections from golint and other tickets as well). 

It's ok, java has much more inspections, they are run concurrently and it shouldn't be a problem.

> Is there something that we could to for example run a single function checker that combines the job of 3-4 inspections so that we avoid walking the tree 3-4 times and we do it just once? 

Yes, there are couple approaches to do less work in the very similar inspections, but for now it's premature optimization. As for walking the tree, platform does it for us, psi tree is not traversed for each inspection separately.

PR #1555, opened 2015-04-12
Initial go get manager

dlsniper:
This allows people to use [go-search.org](http://go-search.org) to find packages they are looking for.

There are a few things that could definitely be improved:
- set focus on the input box when the window is displayed
- add a search field like IDEA itself has (the one with the x button on the right side)
- deduplicate pacakges (maybe this one should be in the go-search repository?)
- detect when a package is already installed and run go get -u instead of go get
- the UI?
- the whole code itself? :D

Here's a screenshot of it in action:
snapshot40

### DIFF
--- a/src/META-INF/plugin.xml
+++ b/src/META-INF/plugin.xml
@@ -26,7 +26,7 @@
     <p>Alpha pre-release of the 1.0.0 version.</p>
     <p>Doesn't contain all the functionality of the 0.9.x branch but has a completely reworked internals.
       It's faster than 0.9.x, refactoring works to some degree and has native support for gopath packages.</p>
-    
+
     <h3>Compatibility</h3>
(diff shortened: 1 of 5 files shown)


zolotov:
pushed in separate branch 'package-searcher'

zolotov:
To be honest it's the ugliest dialog on my laptop with thousand borders and blinking components :) Plus it definitely shouldn't be a window.

Despite of UI there are many problems with yellow code, NPEs, concurrency and memory usage. I don't even know where I should start the review. I'll try to clean it up in few days and then feel free to continue developing in branch. At the moment it's not good enough even for nightly builds.

screenshot 2015-04-24 04 30 14

dlsniper:
All I can say is: I'm sorry you have to go thru this. Thank you for your patience!

Issue #1023, opened 2014-11-19
v1.0.0: Redeclared method is not reported


horkhe:
```Go
type Foo int

func bar(a int) {}
func (foo *Foo) bar(a int) {} // Ok: a package can contain a function and a method with the same

func bar() // Ok: error reported: "bar redeclared in this block"
func (foo *Foo) bar() {} // Should be error: "(*Foo).bar redeclared in this block"
func (foo Foo) bar() {} // Should be error: "method redeclared: Foo.bar"
```

linniksa:
It can be closed?

dlsniper:
Yes, thank you!

PR #1578, opened 2015-04-26
Duplicate method inspection

linniksa:
Simple implementation
relates issue(s): #1023

### DIFF
--- a/src/com/goide/inspections/GoDuplicateMethodInspection.java
+++ b/src/com/goide/inspections/GoDuplicateMethodInspection.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Mihai Toader, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/com/goide/inspections/GoDuplicateMethodInspection.java
```diff
@@ -0,0 +1,72 @@
...
+import java.util.Collection;
+import java.util.List;
+
+
+public class GoDuplicateMethodInspection extends GoInspectionBase {
+  @Override
+  protected void checkFile(@NotNull GoFile file, @NotNull final ProblemsHolder problemsHolder) {
+    final MultiMap<String, GoMethodDeclaration> map = new MultiMap<String, GoMethodDeclaration>();
+    List<GoFile> files = GoPsiImplUtil.getAllPackageFiles(file);
+    for (GoFile goFile : files) {
```
zolotov: Please use GoMethodIndex instead of constructing your own map.

linniksa:
> Please use GoMethodIndex instead of constructing your own map.

done, also `GoDuplicateFunctionInspection` should be updated to use `GoFunctionIndex`?

zolotov:
Merged, thanks.

> done, also GoDuplicateFunctionInspection should be updated to use GoFunctionIndex?

Yes, it would be nice. Also I hope @ignatov will share his thought about global duplications inspection

ignatov:
Yep, we should unify these inspection in a single one.

linniksa:
@ignatov 
something like that: https://github.com/linniksa/go-lang-idea-plugin/commit/ba7be67945fcf2eab69b9599c8d04f5d87a5a552 ?

PR #1597, opened 2015-05-02
Process +build flags (fixes #1199)

### LINKED ISSUE #1199: [1.0] support +build flag
```go
package main

import "net"

var _ net.TCPListener

func main() {
}
```
Unresolved type 'TCPListener'

plugin: 1.0.0.alpha#11
os: windows
idea 14.0.3 (IC-139.1117)

linniksa:
fixes #1199

Also currently two files in same directory (gen1, get2 for example)

```go
// +build ignore

package main;

func test() {

}
```
produce test function is duplicates

### DIFF
--- a/src/com/goide/util/GoUtil.java
+++ b/src/com/goide/util/GoUtil.java
@@ -30,7 +30,6 @@ import com.intellij.openapi.ui.TextBrowseFolderListener;
 import com.intellij.openapi.ui.TextFieldWithBrowseButton;
 import com.intellij.openapi.util.Condition;
 import com.intellij.openapi.util.SystemInfo;
-import com.intellij.openapi.util.io.FileUtil;
 import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.openapi.vfs.VirtualFile;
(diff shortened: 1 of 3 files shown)


zolotov:
Looks nice, thanks. I think we should extract matching to separate class that will accept target os and build-constraints. So it will be easy to test isolated class and GoUtil won't be so big :)

zolotov:
Also please keep the logic with go.target.os property until there is no UI for that.

src/com/goide/util/GoUtil.java
```diff
@@ -49,38 +49,180 @@
...
+    // TODO support .c, .cpp and other
+    if (!(file instanceof GoFile)) {
+      return false;
+    }
+
+    // Look for +build comments to accept or reject the file.
+    // TODO support for custom build tags
+    String flags = ((GoFile)file).getBuildFlags();
+
+    if (null == flags || flags.isEmpty()) {
```
zolotov: Cannot be empty but anyway: StringUtil.isEmpty checks both conditions.

linniksa:
done

PR #1602, opened 2015-05-03
Better regex matcher for Go tip (fix #1279)

### LINKED ISSUE #1279: [1.0] Add support for Go tip SDK
Currently it's not possible to add the Go tip SDK (aka custom build from git repository, usually on HEAD). The SDK detection rules should be improved to support this case.

dlsniper:
This should prevent future developments of the go devel version changes to stop the plugin recognizing it.


zolotov:
What version string didn't work?

dlsniper:
@zolotov I've extracted the version detection from the reading of it and added a few tests from the go versions I have installed.

zolotov:
Cool! That's that I was going to do.

PR #1640, opened 2015-05-23
[WIP] Integrate delve debugger

dlsniper:
Delve debugger https://github.com/derekparker/delve integration


dlsniper:
I've opened this to make it easier to track progress / ask help when needed. As such, this is a very good moment to ask for some guidance.

We can interact with delve over http (not websockets, implementation changed from the original but it's not relevant right now).

One way to run delve is:
```bash
dlv -addr="127.0.0.1:6663" -log -headless run  ./demo.go
```

This will open an http server on port 6663 and will accept connections.

We could also offer the option to do a remote debugging session and just start sending requests to the said IP address.

However, before even getting to that point, I'd like to know how can I actually make the debugger work to the point where I could actually integrate the interfacing with delve.

One of the problems that I have, for example, is that I get a failed assertion if I call ```initBreakpoints()``` whereas the gdb code works with the same call, see https://github.com/go-lang-plugin-org/go-lang-idea-plugin/commit/7f00f486b3bf8379e9d04193295dd9b1ca61d55d#diff-2e8c834d3a06a8c3b1a8638af52e0137R82

Any pointers on how to move forward? Thank you!

dlsniper:
I'll rebase this but I'm kinda stuck on progress due to the above info. Should I ask on the developer forum on devnet?

dlsniper:
Some good news, I've managed to get past the IDE error after a bunch of mindless copy-paste from the GDB implementation. Now the fun part starts, implementing commands to be sent to Delve and process the output to it + make sense of the difference between what Delve offers vs what IDEA needs.

dlsniper:
Small update on this one. I'm at the stage where the debugger should start the binary but it's not doing so. Next step will be to replace the copied infrastructure from the GDB with one specific to delve. This means that instead of launching the debugging binary, the IDE should connect to it via json-rpc (the format that delve now supports) and, if all goes well, start interacting with it.

antonikonovalov:
Hi! When you planned merge?

ignatov:
This branch is out of date.

PR #1654, opened 2015-05-28
Add inspection for functions with the same name as builtin functions (fixes #1450)

### LINKED ISSUE #1450: Flag overriding builtin functions
Taking the code from #1449 ```new``` should be flagged with a weak warning about overriding the ```new``` function from the language.

```go
package main

import "fmt"

func new(v bool) { // flag this as overriding the new function from builtin language functions
    fmt.Printf("Got %v\n", v)
}

func main() {
    myVar := true // Unused variable 'myVar'
    new(myVar) // Unresolved type 'myVar'
}
```

dlsniper:
Lets see if I've understood anything from today :D 
I'm not really sure about the 
```java
    String functionName = psiElement.getText();
    for (GoFunctionDeclaration builtinFunctionDeclaration : builtin.getFunctions()) {
      if (builtinFunctionDeclaration.getIdentifier().getText().equals(functionName)) {
```
part but I don't see another way to get the identifier text out of it.

Also, I'm sorry for the commit noise related to the white space at the end of the lines.

One last thing that's not clear to me is why all the tests that break now break because of the missing description which wasn't there in the first place. On master everything works as expected, on this branch once I have errors/warnings from the inspections, tests break.

Thank you!

### DIFF
--- a/src/com/goide/inspections/GoRenameQuickFix.java
+++ b/src/com/goide/inspections/GoRenameQuickFix.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Mihai Toader, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.GoFunctionDeclaration;
+import com.goide.psi.GoVarDefinition;
+import com.intellij.codeInspection.LocalQuickFixOnPsiElement;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.refactoring.rename.RenameProcessor;
+import org.jetbrains.annotations.NotNull;
+
+public class GoRenameQuickFix extends LocalQuickFixOnPsiElement {
+  private final String myNewName;
+  private final PsiElement myElement;
+
+  protected GoRenameQuickFix(@NotNull PsiElement element, String newName) {
+    super(element);
+    myNewName = newName;
+    myElement = element;
+  }
+
+
+  @Override
+  public void invoke(@NotNull Project project, @NotNull PsiFile file, @NotNull PsiElement startElement, @NotNull PsiElement endElement) {
+    if (!startElement.isWritable()) {
+      return;
+    }
+
+    new RenameProcessor(project, startElement, myNewName, false, true).run();
+  }
+
+  @NotNull
+  public String getFamilyName() {
+    return "Go";
+  }
+
+  @NotNull
+  @Override
+  public String getText() {
+    if (myElement instanceof GoFunctionDeclaration) {
+      return "Rename function to " + myNewName;
+    }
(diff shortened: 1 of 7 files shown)


dlsniper:
Another thing, is it possible to return a rename refactoring as a quick fix from the inspection? As that would be a good fix for this particular problem.

zolotov:
You can make a wrapper, e.g. https://github.com/JetBrains/intellij-community/blob/master/python/src/com/jetbrains/python/inspections/quickfix/RenameParameterQuickFix.java

dlsniper:
I've added it but I'm not happy with the refactoring as it highlights the whole function instead of the identifier only. I'll improve it when I'll have some battery on my laptop. Thanks.

dlsniper:
Todo: add the same for variable name and built-in types, see https://groups.google.com/forum/m/#!topic/golang-nuts/J_xAwZ2ckyY

src/com/goide/inspections/GoDefinedAsBuiltinFunction.java
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  protected void check(@NotNull GoFunctionDeclaration goFunction, @NotNull ProblemsHolder holder) {
+    GoFile builtin = GoSdkUtil.findBuiltinFile(goFunction);
+    if (builtin == null) {
+      return;
+    }
+    String functionName = goFunction.getIdentifier().getText();
+    for (GoFunctionDeclaration builtinFunctionDeclaration : builtin.getFunctions()) {
+      if (builtinFunctionDeclaration.getIdentifier().getText().equals(functionName)) {
```
zolotov: getName instead of getIdentifier().getText()

src/com/goide/inspections/GoDefinedAsBuiltinFunction.java
```diff
@@ -0,0 +1,55 @@
...
+        holder.registerProblem(goFunction, errorText(functionName), ProblemHighlightType.GENERIC_ERROR_OR_WARNING, new GoRenameFunctionQuickFix("my" + functionName));
```
zolotov: register problem on goFunction.getIdentifier

src/com/goide/inspections/GoRenameFunctionQuickFix.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.goide.psi.GoFunctionDeclaration;
+import com.intellij.codeInspection.LocalQuickFix;
+import com.intellij.codeInspection.ProblemDescriptor;
+import com.intellij.openapi.project.Project;
+import com.intellij.psi.PsiElement;
+import com.intellij.refactoring.rename.RenameProcessor;
+import org.jetbrains.annotations.NotNull;
+
+public class GoRenameFunctionQuickFix implements LocalQuickFix {
```
zolotov: extend LocalQuickFixOnPsiElement, then you'll be able to pass real GoFunction there despite of registering problem on its identifier

zolotov:
As for variables, I suggest to make all of these in single inspection. E.g. JS plugin has similar inspection ReservedWordUsedAsName that checks all entities

dlsniper:
I've implemented all the feedback, thank you. I've also squashed all the commits into a single one. Let me know if there's anything else needed.

src/com/goide/inspections/GoRenameQuickFix.java
```diff
@@ -0,0 +1,66 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class GoRenameQuickFix extends LocalQuickFixOnPsiElement {
+  private final String myNewName;
+  private final PsiElement myElement;
+
+  protected GoRenameQuickFix(@NotNull PsiElement element, String newName) {
+    super(element);
+    myNewName = newName;
+    myElement = element;
```
zolotov: Do not store psi element in quick fix. It's memory leak. Compute text during quickfix constructing or in inspection.

zolotov: I'll fix it while merging.

dlsniper: I can still do the change if you want. And thanks for the heads-up about the memory leak, very good to know :)

PR #1668, opened 2015-06-01
Assignment count mismatch inspection (fixes #1219 #1842 #1906)

### LINKED ISSUE #1219: Assignment count mismatch should be reported
```go
package main
var c, d = func() { }
```

dlsniper:


dlsniper:
This doesn't work for cases when there function calls are made on types that should first be resolved (for example a custom ```type f func() (int, string)```

dlsniper:
Also it will conflict with the previous inspection commit so I'd like that one to be merged first then I can fix this one.

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -97,43 +98,43 @@ protected boolean isWriteActionRequired() {
   public void testForRange()    { doTest(); }
   public void testMismatch()    { doTest(); }
   public void testStubParams()  { doTest(); }
-  
+
```
ignatov: Please push your formatting changes as a separate commit. Thanks.

ignatov:
Please rebase it.

dlsniper:
I've done the rebasing. Any hints on how can I make it resolve custom types like this: ```type f func() (int, string)```? Or should I add a test that fails with the example to make it more clear?

ignatov:
Please add a test. And rebase it again.

dlsniper:
Hi, here's a test that fails: https://travis-ci.org/go-lang-plugin-org/go-lang-idea-plugin/builds/66123029#L261 and there are additional tests in ``` GoHighlightingTest ``` which fail due to the way I'm handling resolving for right side of equality here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1668/files#diff-ea03e59b3460dbbf6ce0de675989071aR67 I guess I should use something that the plugin has to resolve directly to the final type and then just use that information but I don't know what to use. Please let me know if you need further details. Thank you.

dlsniper:
Ping for this one. The tests that fail are in the ```GoHighlightingTest``` already. Should I add separate ones?

ignatov:
Please add some. And rebase the PR.

dlsniper:
Rebased and added separate test.

dlsniper:
Hi, I've rebased this one to the latest master. I'm still not sure how to fix this so any pointers are greatly appreciated.

dlsniper:
@ignatov there are a couple of cases where this fails, any help / pointers are greatly appreciated. Thank you.

dlsniper:
I've updated the PR to the best of my knowledge for now and only a handful tests fail now but I don't really know how to fix them.

Also, it's intriguing that `com.goide.inspections.GoHighlightingTest.testCgotest` test fails as the `unsafe` package is in the SDK and when running in standalone mode everything else works ok.

Guidance is very much appreciated. Thank you.

dlsniper:
@ignatov should I try to rebase this or should I close this someone else can work on it. Thank you.

ignatov:
Something wrong with tests on CI, please have a look

dlsniper:
@ignatov that's intended. 
The tests from inspection should fail because I don't know how to correctly resolve everything, that's why I was asking for help (and I still need some help there).

As for the sdk tests failing, I don't know why that happens, I'll try to look into it but the PR should touch that section at all.

Thank you.

zolotov:
> As for the sdk tests failing, I don't know why that happens, I'll try to look into it but the PR should touch that section at all.

Rebase

dlsniper:
Done. I guess the TeamCity integration is disabled for now?

dlsniper:
I only have to handle a few more cases here but I think I'll be very close to do a type inference system with this PR (which might be an indicator of doing something wrong or might be a good thing).

dlsniper:
@ignatov this is ready for review.

dlsniper:
Based on the feedback from @sjamesr I've refactored the implementation a bit and moved the functions to `GoPsiImplUtil` so that they can be reused in other places if needed.

dlsniper:
Removed dead-code and reverted changes to unrelated files.

dlsniper:
I've started a project-wide inspection for Docker and I've found a few places where this doesn't work as expected. I'll update the issue as soon as I fix the false-positives.

dlsniper:
I'm currently inspecting my whole GOPATH, which is about 20k different files (and about 5.7 million lines of Go) to see if I've missed any other checks. So far I've found at least a dozen more things since I've wrote earlier but progress is easy to make right now. Hopefully I'll be able to fix everything that I come across in the next few hours and make this properly available for review.

I've also noticed that TeamCity sometimes fails with getting the Gradle part setup, don't kno where I should report it (or if it's known).

dlsniper:
Currently the tests fail but I'm not sure why this happens as I can't replicate it on a standalone version. I'll look more into this in the comming days.

dlsniper:
Judging from what I have in my GOPATH, there are still issues with the current code, aside from those in the tests already, where this is not working as expected.

@ignatov in my quest to solve this I think I've gotten pretty close to a type inference solution. Could you please have a look on the code and see if it would make sense to change it from `getArity` to `getRealType` (for lack of a better word) and then port the `getArity` functionality on it?

I think I'm close to the limits of what I can do with various ifs in "random" places and I'd rather not butcher this if it's possible to make it more useful.

Thank you.

ignatov:
> @ignatov in my quest to solve this I think I've gotten pretty close to a type inference solution.
I'm not sure.

dlsniper:
I'll need to rework this PR a bit so that instead of having the arity
returned we get the types for all the entries of a List<GoExpression>.
After we have the types, it will be easier to call the arity function on
each of the type and have it compute the result.
If we have a function call inside the assignment, then we should only show
the error about multiple-value call in a single-value context error (or no
value for function call).
Hopefully it should make the code easier to read as well.

dlsniper:
@ignatov if you ignore the previous comment, this PR is ready for review. I've inspected as much as I could of the Kubernetes project, ` go get -v -u -t k8s.io/kubernetes/... ` (maybe do it 1-2 times), until eventually the IDE ran out of memory (maybe I'm leaking some memory somewhere?)
Other than that, maybe there's some need for caching so that we don't always resolve the whole tree for the same call (but might be a premature optimization).

ignatov:
I'm a bit busy right now, will review later, OK?

dlsniper:
Yup, sure, nothing to rush. I've left the comment so that you know the state for it.

dlsniper:
I've found a few additional cases I'm not handling so I'll fix the filing test soon.

dlsniper:
The latest master changes related to functions broken this, I'll fix it tomorrow.

dlsniper:
Done. rebased and all works. I've processed my whole GOPATH (I have docker, kubernetes and a few other large code bases) and nothing came up. I've also processed Go 1.6 Beta 1 and again nothing. There might be some other cases which I don't process right now but lets see what the users will find.

There's also a method in `GoPsiImplUtil` called `isFuncInFuncSignature` which I'm not very proud of but it's the best I can do right now to check if the signature we are resolving is in fact the signature of the function or of a function which returns a function.

dlsniper:
@ignatov i've tried to remove unrelated changes to the best of my abilities. There are still some diffs because of whitespace being trimmed in files I've changed (such as GoPsiImplUtil). If I should change those as well, please let me know. I'm not sure why my IDE settings are different in this regard as I'm using the settings that came with the code.

ignatov:
Obsole?

dlsniper:
> Obsole?

No, at least the inspection is not in the plugin yet. Should I update it? It's a big effort to review it, I know, but I don't see how else to get this done. Thank you.

ignatov:
OK, please update

PR #1689, opened 2015-06-21
Show parameters info in closure calls (fixes #1687)

### LINKED ISSUE #1687: Parameter info not show for closure function calls
In a function call like this
```go
go func(param1, param2 string) (return1, return2 string) {
    return param1, param2
}("1", "2"<cursor>)
```

the parameter information is not show when invoked at ```<cursor>```

dlsniper:


ignatov:
Merged manually.

PR #1710, opened 2015-06-28
Disable debugging for now

dlsniper:
I'd like to disable the debug options for now as we don't really support them (also I want to disable GDB until I have some time to look at it and improve it, apparently it should support apps compiled with gccgo??)

### DIFF
--- a/src/com/goide/runconfig/GoBuildingRunner.java
+++ b/src/com/goide/runconfig/GoBuildingRunner.java
@@ -26,6 +26,7 @@ import com.intellij.execution.RunProfileStarter;
 import com.intellij.execution.RunnerAndConfigurationSettings;
 import com.intellij.execution.configurations.RunProfile;
 import com.intellij.execution.configurations.RunProfileState;
+import com.intellij.execution.executors.DefaultDebugExecutor;
 import com.intellij.execution.executors.DefaultRunExecutor;
 import com.intellij.execution.process.ProcessAdapter;
 import com.intellij.execution.process.ProcessEvent;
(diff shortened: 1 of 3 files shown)


dlsniper:
I've also noticed that even if I mark an executor as being not supported, it will still be shown enabled in the menu, I've opened up this one: https://youtrack.jetbrains.com/issue/IDEA-142018

src/com/goide/runconfig/GoRunner.java
```diff
@@ -32,6 +33,9 @@ public String getRunnerId() {
 
   @Override
   public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
+    if(DefaultDebugExecutor.EXECUTOR_ID.equals(executorId)) {
```
zolotov: what's the point? The method is always false for DefaultDebugExecutor.EXECUTOR_ID without your changes.

src/com/goide/runconfig/GoBuildingRunner.java
```diff
@@ -55,6 +56,9 @@ public String getRunnerId() {
 
   @Override
   public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
+    if(DefaultDebugExecutor.EXECUTOR_ID.equals(executorId)) {
```
zolotov: what's the point? The method is always false for DefaultDebugExecutor.EXECUTOR_ID without your changes.

dlsniper: I should learn boolean conditions I guess :D I didn't realized that's working because the menu option for debugging is always enabled and I what was in my mind was to always disable it forcefully.

src/com/goide/runconfig/GoRunner.java
```diff
@@ -17,6 +17,7 @@
 package com.goide.runconfig;
 
 import com.intellij.execution.configurations.RunProfile;
+import com.intellij.execution.executors.DefaultDebugExecutor;
```
zolotov: unused import

Issue #1560, opened 2015-04-18
Bug: private funcs and struct fields accessible in package_test test files


SchumacherFM:
![screen shot 2015-04-16 at 8 07 41 pm](https://cloud.githubusercontent.com/assets/372027/7214836/428fba2e-e600-11e4-956c-42a917e16b8a.png)

![screen shot 2015-04-18 at 7 20 56 pm](https://cloud.githubusercontent.com/assets/372027/7214842/8eff9dde-e600-11e4-8f92-1ef71423a195.png)

Reproduceable:

file1:

```go
package intlbug

import "strings"

type MyPrivates struct {
	value1 int
	value2 string
}

func (p MyPrivates) AFunc() string {
	return strings.Repeat(p.value2, 2)
}
```

file2 is in the same package as file1:

```go
package intlbug_test

import (
	"testing"

	"intlbug"
)

func TestAFunc(t *testing.T) {
	mp := intlbug.MyPrivates{}
	mp.....
}
```

Expected result: I cannot access the fields of my privates :laughing:

Idea 14.1.1, go-lang-idea-plugin: latest commit from branch master. Runs on OSX latest.

Thanks for having a look into this!

ignatov:
Thanks for reporting.

Sergey Ignatov

ignatov:
Sorry, can't reproduce. Please attach a layout (output from `tree` util) or just a small project for reproducing.

SchumacherFM:
Here is the repo: https://github.com/SchumacherFM/intellijGoPluginBug `go get github.com/SchumacherFM/intellijGoPluginBug`

It is still reproduceable.

ignatov:
Thanks, http://linux.die.net/man/1/tree

SchumacherFM:
Cool! Thanks a lot!

PR #1715, opened 2015-07-02
Get file package names in terms of Go

eugene-auduchinok:
I've got `testPackageWithTestPrefix()` failing. How should I update the test?

### DIFF
--- a/src/com/goide/inspections/GoMultiplePackagesInspection.java
+++ b/src/com/goide/inspections/GoMultiplePackagesInspection.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Mihai Toader, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.GoConstants;
+import com.goide.psi.GoFile;
+import com.goide.psi.GoPackageClause;
+import com.goide.util.GoUtil;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.psi.PsiDirectory;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+
+public class GoMultiplePackagesInspection extends GoInspectionBase {
+
+  @Override
+  protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsHolder) {
+    if (!GoUtil.allowed(file)) {
+      return;
+    }
+    GoPackageClause packageClause = file.getPackage();
+    if (packageClause != null) {
+      String packageName = file.getPackageName();
+      if (packageName == null || packageName.equals(GoConstants.DOCUMENTATION)){
+        return;
+      }
+      PsiDirectory dir = file.getContainingDirectory();
+      Collection<String> packages = GoUtil.getAllPackagesInDirectory(dir);
+      packages.remove(GoConstants.DOCUMENTATION);
+      if (packages.size() > 1) {
+          problemsHolder.registerProblem(packageClause, "Multiple packages in directory", new GoMultiplePackagesQuickFix(packageClause, packages, problemsHolder.isOnTheFly()));
+      }
+    }
+  }
+}
--- a/src/com/goide/inspections/GoMultiplePackagesInspection.java
+++ b/src/com/goide/inspections/GoMultiplePackagesInspection.java
@@ -20,10 +20,12 @@ import com.goide.GoConstants;
 import com.goide.psi.GoFile;
 import com.goide.psi.GoPackageClause;
 import com.goide.util.GoUtil;
+import com.intellij.codeInspection.LocalQuickFix;
 import com.intellij.codeInspection.ProblemsHolder;
 import com.intellij.psi.PsiDirectory;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.ArrayList;
 import java.util.Collection;
 
 public class GoMultiplePackagesInspection extends GoInspectionBase {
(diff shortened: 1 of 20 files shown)


src/com/goide/psi/GoFile.java
```diff
@@ -414,7 +415,11 @@ public String getPackageName() {
     if (packageClause != null) {
       PsiElement packageIdentifier = packageClause.getIdentifier();
       if (packageIdentifier != null) {
-        return packageIdentifier.getText().trim();
+        String name = packageIdentifier.getText().trim();
+        if (GoTestFinder.isTestFile(this)) {
```
zolotov: 1. could be rewritten with ternary operator (not required)
2. the case when isTestFile(file)=false and package name ends with _test is not covered with test
3. the case when isTestFile(file)=true and package name doesn't end with _test is not covered with test

zolotov:
Add tests and rebase with master

zolotov:
> I've got testPackageWithTestPrefix() failing. How should I update the test?

Disable it (rename to _testPackageWithTestPrefix), I'll fix it later

tests/com/goide/completion/GoCompletionTest.java
```diff
@@ -437,7 +437,7 @@ public void testPackageNames() {
     myFixture.configureByText("test_file.go", "package fromFile");
     myFixture.configureByText("test.go", "package <caret>");
     myFixture.completeBasic();
-    assertSameElements(myFixture.getLookupElementStrings(), "fromTest", "fromTest_test", "fromFile", "main");
+    assertSameElements(myFixture.getLookupElementStrings(), "fromTest", "fromFile", "main");
```
zolotov: 1. completion still should suggest *_test packages in test files. Add test for that.
2. does following code in GoCompletionContributor is still needed?
```
if (packageName.endsWith(GoConstants.TEST_SUFFIX)) {
          result.addElement(LookupElementBuilder.create(StringUtil.trimEnd(packageName, GoConstants.TEST_SUFFIX)));
        }
```

eugene-auduchinok:
Need to add tests for inspection and quick fix.

src/com/goide/inspections/GoMultiplePackagesQuickFix.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  private static void renamePackagesInDirectory(@NotNull final Project project,
+                                                @NotNull final PsiDirectory dir,
+                                                @NotNull final String newName) {
+    WriteCommandAction.runWriteCommandAction(project, new Runnable() {
+      @Override
+      public void run() {
+        for (PsiFile file : dir.getFiles()) {
+          if (GoUtil.allowed(file)) {
```
zolotov: Is this really needed?

src/com/goide/inspections/GoMultiplePackagesInspection.java
```diff
@@ -0,0 +1,50 @@
...
+import com.goide.psi.GoFile;
+import com.goide.psi.GoPackageClause;
+import com.goide.util.GoUtil;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.psi.PsiDirectory;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.Collection;
+
+public class GoMultiplePackagesInspection extends GoInspectionBase {
```
zolotov: Should be tested

src/com/goide/inspections/GoMultiplePackagesQuickFix.java
```diff
@@ -0,0 +1,107 @@
...
+
+public class GoMultiplePackagesQuickFix extends LocalQuickFixAndIntentionActionOnPsiElement {
+  private Collection<String> myPackages;
+  private String myPackageName;
+  private boolean myIsOneTheFly;
+
+  protected GoMultiplePackagesQuickFix(@NotNull GoPackageClause packageClause, Collection<String> packages, boolean isOnTheFly) {
+    super(packageClause);
+    myPackages = packages;
+    myPackageName = ((GoFile)packageClause.getContainingFile()).getPackageName();
```
zolotov: Since packageClause is passed only for retrieving packageName it's better to pass packageName

src/com/goide/inspections/GoMultiplePackagesInspection.java
```diff
@@ -0,0 +1,50 @@
...
+    if (packageClause != null) {
+      String packageName = file.getPackageName();
+      if (packageName == null || packageName.equals(GoConstants.DOCUMENTATION)){
+        return;
+      }
+      PsiDirectory dir = file.getContainingDirectory();
+      Collection<String> packages = GoUtil.getAllPackagesInDirectory(dir);
+      packages.remove(GoConstants.DOCUMENTATION);
+      if (packages.size() > 1) {
+          problemsHolder.registerProblem(packageClause, "Multiple packages in directory", new GoMultiplePackagesQuickFix(packageClause, packages, problemsHolder.isOnTheFly()));
```
zolotov: For !onTheFly mode you should add quick fix for variant of package renaming

src/com/goide/util/GoUtil.java
```diff
@@ -195,9 +195,9 @@ public static boolean couldBeReferenceTo(@NotNull PsiElement definition, @NotNul
       public Result<Collection<String>> compute() {
         Collection<String> set = ContainerUtil.newLinkedHashSet();
         for (PsiFile file : dir.getFiles()) {
-          if (file instanceof GoFile) {
+          if (allowed(file)) {
```
zolotov: Is this really needed? If it so than `dir` is wrong dependency for cache. Result of `allowed` invocation might change on changing IDE settings

src/com/goide/inspections/GoMultiplePackagesQuickFix.java
eugene-auduchinok: It's passed into the base class constructor.

src/com/goide/util/GoUtil.java
eugene-auduchinok: Do we need to account packages in files that are not allowed for build?
https://github.com/golang/go/blob/master/src/go/build/build.go#L447
I'm not sure if I get it right.

tests/com/goide/completion/GoCompletionTest.java
eugene-auduchinok: Fixed.

ignatov:
Sorry, I've commented another PR. BTW, will be cool to fix this: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1731#issuecomment-119647823

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -131,7 +132,7 @@ public void testInnerTypesFromOtherPackage() {
 
   public void testDuplicatesNoLocalResolveForTest() {
     myFixture.configureByText("a.go", "package i; type P struct { v1 int }");
-    myFixture.configureByText("b.go", "package i_test; import ( \".\" ); func <warning>f</warning>() { print(i.P{}.<error>v1</error>) }");
+    myFixture.configureByText("b.go", "<error>package i_test</error>; import ( \".\" ); func <warning>f</warning>() { print(i.P{}.<error>v1</error>) }");
```
zolotov: Changing getPackageName() logic broke the initial issue (#1560). With this change you've fixed the test, not the bug. 
I'll fix it, but be careful in the future.

src/com/goide/inspections/GoMultiplePackagesQuickFix.java
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  private static void renamePackagesInDirectory(@NotNull final Project project,
+                                                @NotNull final PsiDirectory dir,
+                                                @NotNull final String newName) {
+    WriteCommandAction.runWriteCommandAction(project, new Runnable() {
+      @Override
+      public void run() {
+        for (PsiFile file : dir.getFiles()) {
+          if (file instanceof GoFile && !GoUtil.libraryDirectoryToIgnore(file.getName())) {
```
zolotov: Does it look like library directory?

eugene-auduchinok: Nope, sorry. Will extract it's logic to a different method.

zolotov: Don't. PR is quite big already and it's not convenient to review it, so I'll cleanup and merge it tonight.
After merging please see cleanup commits. Also it would be nice if you add extra test about non highlighting not-allowed files with the inspection.

PR #1731, opened 2015-07-08
Use proper package name when create file

eugene-auduchinok:


src/com/goide/actions/GoCreateFileAction.java
```diff
@@ -41,7 +46,12 @@
   protected PsiFile createFile(String name, @NotNull String templateName, @NotNull PsiDirectory dir) {
     FileTemplate template = FileTemplateManager.getInstance(dir.getProject()).getInternalTemplate(templateName);
     Properties properties = new Properties();
-    properties.setProperty(PACKAGE, ContainerUtil.getLastItem(StringUtil.split(dir.getName(), "-")));
+    Collection<String> packages = GoUtil.getAllPackagesInDirectory(dir);
+    String packageName = packages.isEmpty() ? GoPsiImplUtil.makePackageNameValid(ContainerUtil.getLastItem(StringUtil.split(dir.getName(), "-"))) : ArrayUtil.toStringArray(packages)[0];
```
zolotov: Why do you need to split directory name? Doesn't makePackageNameValid do you for you?

src/com/goide/actions/GoCreateFileAction.java
```diff
@@ -41,7 +46,12 @@
...
+    if (name.endsWith(GoConstants.TEST_SUFFIX_WITH_EXTENSION) || name.endsWith(GoConstants.TEST_SUFFIX)) {
```
zolotov: > ArrayUtil.toStringArray(packages)[0]

ContainerUtil.getFirstItem()

zolotov: Why do you need to check `name.endsWith(GoConstants.TEST_SUFFIX)`?

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -816,8 +816,24 @@ public static GoImportSpec addImportSpec(@NotNull GoImportDeclaration declaratio
...
+      if (Character.isWhitespace(ch) ||  illegalChars.indexOf(ch) != -1) {
+        name.setCharAt(i, '_');
+      }
+    }
+    return name.toString();
+  }
+
   public static String getLocalPackageName(@NotNull String importPath) {
-    return PathUtil.getFileName(importPath);
+    return makePackageNameValid(PathUtil.getFileName(importPath));
```
zolotov: inline method

src/com/goide/actions/GoCreateFileAction.java
eugene-auduchinok: Because it checks user's input that may have no extension.

src/com/goide/actions/GoCreateFileAction.java
```diff
@@ -41,7 +44,12 @@
   protected PsiFile createFile(String name, @NotNull String templateName, @NotNull PsiDirectory dir) {
     FileTemplate template = FileTemplateManager.getInstance(dir.getProject()).getInternalTemplate(templateName);
     Properties properties = new Properties();
-    properties.setProperty(PACKAGE, ContainerUtil.getLastItem(StringUtil.split(dir.getName(), "-")));
+    Collection<String> packages = GoUtil.getAllPackagesInDirectory(dir);
```
zolotov: It makes sense just to getFirstItem and check for null instead of checking isEmpty and then hope that nothing will be changed in future and getFirstItem will return notnull value.

ignatov:
Please avoid an empty string from the package name completion: 


screenshot 2015-07-08 10 14 57

PR #1738, opened 2015-07-10
Add import cgo in test inspection

eugene-auduchinok:


src/com/goide/inspections/GoCGOInTestInspection.java
```diff
@@ -0,0 +1,42 @@
...
+import com.goide.psi.GoFile;
+import com.goide.psi.GoImportSpec;
+import com.goide.runconfig.testing.GoTestFinder;
+import com.intellij.codeInspection.ProblemsHolder;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public class GoCGOInTestInspection extends GoInspectionBase {
+  private static final String C_PATH = "C";
```
zolotov: Extract to GoConstants, there is one more usage in com.goide.psi.GoVisitor#visitImportString

src/com/goide/inspections/GoCGOInTestInspection.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.goide.psi.GoFile;
+import com.goide.psi.GoImportSpec;
+import com.goide.runconfig.testing.GoTestFinder;
+import com.intellij.codeInspection.ProblemsHolder;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public class GoCGOInTestInspection extends GoInspectionBase {
```
zolotov: Cgo is not an abbreviation, please rename to GoCgoInTestInspection, make sure that you rename a file (it's kind of tricky on case-insensitive file systems)

PR #1749, opened 2015-07-14
Add GoTestingDeclarationsInspection

eugene-auduchinok:
All testing functions (Test/Benchmark/Example) in test files should have appropriate signature (http://golang.org/pkg/testing/) and the first symbol after testing prefix must not be a lowercase letter.

I've created anonymous inner classes for quick fixes but I'm not sure if it had been better to create separated named classes.
Would it be better if I combine quick fixes into a single class? Or would it be better if I split inspection otherwise?

I haven't created tests yet as I'm not sure if I've done all these things right.

### DIFF
--- a/src/com/goide/inspections/GoTestSignaturesInspection.java
+++ b/src/com/goide/inspections/GoTestSignaturesInspection.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Mihai Toader, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.GoConstants;
+import com.goide.psi.*;
+import com.goide.psi.impl.GoElementFactory;
+import com.goide.runconfig.testing.GoTestFinder;
+import com.goide.runconfig.testing.GoTestFunctionType;
+import com.intellij.codeInspection.LocalQuickFixBase;
+import com.intellij.codeInspection.ProblemDescriptor;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.openapi.project.Project;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+import java.util.Locale;
+
+public class GoTestSignaturesInspection extends GoInspectionBase {
+  private static String getTestingAlias(GoImportSpec testingImportSpec) {
+    return testingImportSpec == null ? GoConstants.TESTING_PATH :
+           !testingImportSpec.isDot() ? StringUtil.notNullize(testingImportSpec.getAlias(), GoConstants.TESTING_PATH) : "";
+  }
+
+  private static String getProperParamType(@NotNull String testingAlias, GoTestFunctionType type) {
+    return "*" + (testingAlias.isEmpty() ? "" : testingAlias + ".") + type.getParamType();
+  }
+
+  @Override
+  protected void checkFile(@NotNull final GoFile file, @NotNull ProblemsHolder problemsHolder) {
+    if (!GoTestFinder.isTestFile(file)) return;
+    GoImportSpec testingImportSpec = file.getImportedPackagesMap().get(GoConstants.TESTING_PATH);
+    String testingAlias = getTestingAlias(testingImportSpec);
+    for (GoFunctionDeclaration function : file.getFunctions()) {
+      GoTestFunctionType type = GoTestFunctionType.fromName(function.getName());
+      if (type == null) continue;
+      GoSignature signature = function.getSignature();
+      if (signature == null) continue;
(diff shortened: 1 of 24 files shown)


src/com/goide/inspections/GoTestingDeclarationsInspection.java
```diff
@@ -0,0 +1,132 @@
...
+import org.jetbrains.annotations.NotNull;
+
+import java.util.List;
+
+public class GoTestingDeclarationsInspection extends GoInspectionBase {
+  private enum TestingType {TEST, TEST_MAIN, BENCHMARK, EXAMPLE}
+
+  private static TestingType getTestingType(@NotNull String name) {
+    if (name.startsWith(GoConstants.TEST_PREFIX)) {
+      return name.toLowerCase().equals("testmain") ? TestingType.TEST_MAIN : TestingType.TEST;
```
zolotov: equalsIgnoreCase

src/com/goide/inspections/GoTestingDeclarationsInspection.java
```diff
@@ -0,0 +1,132 @@
...
+    }
+    if (name.startsWith(GoConstants.BENCHMARK_PREFIX)) return TestingType.BENCHMARK;
+    if (name.startsWith(GoConstants.EXAMPLE_PREFIX)) return TestingType.EXAMPLE;
+    return null;
+  }
+
+  private static String getPrefixFromType(@NotNull TestingType type) {
+    switch (type) {
```
zolotov: I believe TestingType can tell about its prefix without switch/case

src/com/goide/inspections/GoTestingDeclarationsInspection.java
```diff
@@ -0,0 +1,132 @@
...
+  }
+
+  @Override
+  protected void checkFile(@NotNull final GoFile file, @NotNull ProblemsHolder problemsHolder) {
+    if (!GoTestFinder.isTestFile(file)) return;
+    GoImportSpec testingImportSpec = file.getImportedPackagesMap().get(GoConstants.TESTING_PATH);
+    String testingAlias = null;
+    if (testingImportSpec != null) {
+      testingAlias = testingImportSpec.getAlias();
+      testingAlias = testingAlias == null ? testingImportSpec.getPath() : testingAlias;
```
zolotov: getLocalPackageName is more appropriate than getPath.
+ should be tested with dot-imports line `import . "testing"`

src/com/goide/inspections/GoTestingDeclarationsInspection.java
```diff
@@ -0,0 +1,132 @@
...
+
+      final String prefix = getPrefixFromType(type);
+      final String testName = name.substring(prefix.length());
+      if (!StringUtil.isCapitalized(testName)) {
+        final PsiElement identifier = function.getIdentifier();
+        problemsHolder.registerProblem(identifier, "Test name must start with capital letter",
+                                       new LocalQuickFixBase("Capitalize " + testName) {
+                                         @Override
+                                         public void applyFix(@NotNull Project project, @NotNull ProblemDescriptor descriptor) {
+                                           identifier.replace(
```
zolotov: memory leak

dlsniper:
I like the idea about checking for the capital letter.

However all the other parts of the signature are hard to catch because one could use

- gocheck
```
func (ds *DarkSide) TestSomething_Something(c *C) {}
```

- goconvey (maintains the standard test API)
```
func TestBowlingGameScoring(t *testing.T) {}
```

I'm not sure what other testing frameworks are doing but, aside from BDD ones, which all seem to have their own thing way different than the standard API.

zolotov:
@dlsniper checking for capital letter is wrong approach. The right one implemented in GoTestFinder#checkPrefix

zolotov:
> However all the other parts of the signature are hard to catch because one could use

This inspection works for gotest only

eugene-auduchinok:
Oops. I didn't notice failing yesterday. Will fix it asap.

zolotov:
Anyway looks good, tests are still needed though

src/com/goide/inspections/GoTestSignaturesInspection.java
```diff
@@ -0,0 +1,106 @@
...
+    GoImportSpec testingImportSpec = file.getImportedPackagesMap().get(GoConstants.TESTING_PATH);
+    String testingAlias = getTestingAlias(testingImportSpec);
+    for (GoFunctionDeclaration function : file.getFunctions()) {
+      GoTestFunctionType type = GoTestFunctionType.fromName(function.getName());
+      if (type == null) continue;
+      GoSignature signature = function.getSignature();
+      if (signature == null) continue;
+      List<GoParameterDeclaration> params = signature.getParameters().getParameterDeclarationList();
+      if (type == GoTestFunctionType.EXAMPLE) {
+        if (params.size() != 0) {
```
zolotov: !isEmpty

src/com/goide/inspections/GoTestSignaturesInspection.java
```diff
@@ -0,0 +1,106 @@
...
+        signature.replace(GoElementFactory.createFunctionSignatureFromText(project, ""));
+        return;
+      }
+      GoImportSpec testingImportSpec = (((GoFile)file).getImportedPackagesMap().get(GoConstants.TESTING_PATH));
+      if (testingImportSpec == null) {
+        ((GoFile)file).addImport(GoConstants.TESTING_PATH, null);
+      }
+      String paramType = getProperParamType(getTestingAlias(testingImportSpec), myType);
+      signature.replace(GoElementFactory.createFunctionSignatureFromText(project, myType.getParamType().toLowerCase() + " " + paramType));
+    }
```
zolotov: add locale to toLowerCase invokation

src/com/goide/inspections/GoTestSignaturesInspection.java
```diff
@@ -0,0 +1,106 @@
...
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+public class GoTestSignaturesInspection extends GoInspectionBase {
+  private static String getTestingAlias(GoImportSpec testingImportSpec) {
+    String alias = GoConstants.TESTING_PATH;
+    if (testingImportSpec != null) {
```
zolotov: alias variable is redundant

PR #1773, opened 2015-07-26
The "New > Go File" menu item doesn't display where it should on PyCharm.

loic:
screen shot 2015-07-26 at 17 25 56

This PR is a shot in the dark since I don't know how to test it out, but hopefully it points someone in the right direction.

### DIFF
--- a/src/META-INF/plugin.xml
+++ b/src/META-INF/plugin.xml
@@ -240,7 +240,7 @@
   <actions>
     <action id="Go.NewGoFile" class="com.goide.actions.GoCreateFileAction"
             text="Go File" description="Create new Go file">
-      <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFile"/>
+      <add-to-group group-id="NewGroup" anchor="before" relative-to-action="NewFromTemplate"/>
     </action>
 
     <action id="GoShowTypeInternalAction"


zolotov:
> menu item doesn't display where it should on PyCharm.

Where is it should be displayed? Please attach screenshot before and after changes. Thank you.

loic:
I'd say correct location is in the second block along with all the other language template files such as "HTML File" or "JavaScript File".

I found the `NewFromTemplate` directive in the source code of other IDEA plugins, so I suppose it does the trick, but I haven't actually tried it (so no after screenshot) because I don't know how to build this plugin from source. If you have build instructions somewhere I can give it a try.

dlsniper:
@loic you can check out this document: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/master/CONTRIBUTING.md#getting-started-with-the-plugin-development if you need any help, drop me/us a line on Gitter

zolotov:
> I found the NewFromTemplate directive in the source code of other IDEA plugins, so I suppose it does the trick, but I haven't actually tried it (so no after screenshot) because I don't know how to build this plugin from source. If you have build instructions somewhere I can give it a try.

Yes, you suggested absolutely correct solution. But I think that `create Go file` action is displayed in the right place :) Actually I think that it makes sense to lift up `create Python file` action in PyCharm instead, but it's beyond my competence.

loic:
As someone who exclusively uses "New File" and "New Directory" instead of language templates, I find it a bit odd that it's no longer the first choice, especially in a polyglot IDE.

zolotov:
> As someone who exclusively uses "New File" and "New Directory" instead of language templates, I find it a bit odd that it's no longer the first choice, especially in a polyglot IDE.

There are two subgroups in NewGroup: NewGroup1 and NewFromTemplate. NewFromTemplate contains all templates from settings and it can contain plenty variants. NewGroup1 contains the most important actions like 'create module', `create java class` or `create groovy class` in IntelliJ IDEA, `create ruby file` in RubyMine, `create php class` in PhpStorm and so on. So it's ok to lift up *important* languages (PyCharm is an exception, have no idea why). 
Since user installs golang plugin explicitly, I can suggest that creating go files is quite important for him. 
But I admit that `create go file` actions might be placed after `create file/directory` actions, in this case you can write something like `<add-to-group group-id="NewGroup" anchor="after" relative-to-action="NewDir"/>`

Issue #1779, opened 2015-07-31
very odd problem with colouring


nhooyr:
<img width="1440" alt="screen shot 2015-07-31 at 6 32 05 am" src="https://cloud.githubusercontent.com/assets/10180857/9005654/ed2dc27a-374d-11e5-833d-f93cef19d6b2.png">

not sure what the problem is but as you can see on line 141 and more the methods of the variable C are not highlighted. Code base is here on github. file tunnel.go and method client. 

https://github.com/aubble/goTunneLS/blob/master/tunnel.go

dlsniper:
Can you please share your color scheme?

aubble:
absolutely :P https://www.dropbox.com/s/ehb4zzrobjktkup/aubble.jar?dl=0
made it myself :D what should I name it?

aubble:
Oh noticed the same problem on line 134 with ln.Accept()?

PR #1782, opened 2015-07-31
Improve highlighter (fixes #1778 #1779)

### LINKED ISSUE #1778: colors & fonts settings clash
the package struct/interface colors clash with type specification and the plugin ends up using the type specification in the code (the package struct/interface colors are never used because type specification colors are used). also in the highlight sample privateField/func uses identifiers colors which isn't true. It would also be nice if the highlighting sample used all the colors, in every combination, if need be I can create this sample. This is all on the latest version nightly.

dlsniper:


dlsniper:
It's a work in progress as I've opened this because I don't have a way to test this locally with tests and I hope Travis will catch any issues.

dlsniper:
@ignatov what do you think, should I merge the Package, Interface and Struct methods into a single type so that they are easier to process? They would be only `Exported function` and `Private function`. If so, I can finish this one quicker (and the benefit will be that the users have less things to take care of).

dlsniper:
@ignatov this can be merged as it solves some highlighting issues and also simplifies the highlight for methods and functions (see my previous comment). Thank you.

ignatov:
Please cover the brand new functionality with tests, see GoHighlightingAnnotatorTest.

dlsniper:
I'm sorry, I don't understand how to do that. I've added a file like this:
```
package main

func <info descr="null"><info descr="signature_owner">Demo</info></info>() {
}
```

And the test works, but there's no reference to how the function name itself is going to be highlighted.
The changes I've done only reflect in the colors configurable by the end-user. I'd be happy to add the test if I understand how it should be done. Thank you for your help.

zolotov:
> The changes I've done only reflect in the colors configurable by the end-user. 

Doesn't look like that. It seems that you've fixed some annotating logic.

```
if (o instanceof PsiWhiteSpace) {
     return;
}
```

```
PsiElement identifier = ((GoReceiver)o).getIdentifier();
+      if (identifier != null) {
+        setHighlighting(o, holder, METHOD_RECEIVER, "receiver");
+      }
```

```
else if (resolve instanceof GoFunctionOrMethodDeclaration || resolve instanceof GoMethodSpec) {
       setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");		       setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");
     }		     }
```

dlsniper:
I meant in the second commit. Still, I would like to add a test for it but I don't understand how to do it for for the changeset to reflect that new state. I'll try to find some examples in other plugins / IntelliJ Community sources as I don't understand `GoHighlightingAnnotatorTest`.

zolotov:
Do not test the second commit.

> I don't understand how to do it for for the changeset to reflect that new state. I'll try to find some examples in other plugins / IntelliJ Community sources as I don't understand GoHighlightingAnnotatorTest.

Just tell me what you fixed with the following changes and I'll write tests for you:
```
if (o instanceof PsiWhiteSpace) {
     return;
}
```

```
PsiElement identifier = ((GoReceiver)o).getIdentifier();
+      if (identifier != null) {
+        setHighlighting(o, holder, METHOD_RECEIVER, "receiver");
+      }
else if (resolve instanceof GoFunctionOrMethodDeclaration || resolve instanceof GoMethodSpec) {
       setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");               setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");
     }   
```

dlsniper:
Sorry for delay in responding:

- I've removed this (it was a pointless thing I was trying to achieve, please ignore that)
```
if (o instanceof PsiWhiteSpace) {
     return;
}
```

- this was wrong
```
+      PsiElement identifier = ((GoReceiver)o).getIdentifier();
+      if (identifier != null) {
+        setHighlighting(o, holder, METHOD_RECEIVER, "receiver");
+      }
```
and should have been (now fixed)

```
+      PsiElement identifier = ((GoReceiver)o).getIdentifier();
+      if (identifier != null) {
+        setHighlighting(identifier, holder, METHOD_RECEIVER, "receiver");
+      }
```
It changes the highlighting from this:
snapshot95
to
snapshot96

- this unifies the highlighting for methods on structs and for functions
```
else if (resolve instanceof GoFunctionOrMethodDeclaration || resolve instanceof GoMethodSpec) {
       setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");               setHighlighting(o.getIdentifier(), holder, getColor((GoNamedSignatureOwner)resolve), "func");
     }  
```

Please let me know if I can do anything else to help out.

zolotov:
And what't the examples of fixing #1778 and #1779?

dlsniper:
For #1778 I've removed any difference between functions, methods on structs and methods on interfaces as we currently don't have a way to determine if a struct implements and interface or not and to get a list of all interfaces implemented by a struct. So rather than confuse people with too many options that won't work accordingly, we should display only what we can and revisit this when we have a solution for the above mentioned problem.

For #1779 I' using
```go
package main

type (
	de interface {
		Demo() error
	}

	dem struct{}
)

func (a dem) Demo() error {
	return error("demo")
}

func main() {
	b := dem{}
	b.Demo().Error()
}

```

dlsniper:
@zolotov thank you for the tests, I know how how to write those myself ^^

PR #1783, opened 2015-07-31
Allow tests on TravisCI to run on the faster infrastructure

dlsniper:


dlsniper:
Seems Travis doesn't agree with this.. I'll follow up with them.

zolotov:
We won't use travis in the near future so it's not necessary.

PR #1785, opened 2015-08-03
Fix nil assignment inspection for short vars, constants declarations 

eugene-auduchinok:
There was an inspection for nil assignments to variables but it hadn't work for short assignments or constants, so I've fixed it.

### DIFF
--- a/src/com/goide/inspections/GoAssignmentNilWithoutExplicitType.java
+++ b/src/com/goide/inspections/GoAssignmentNilWithoutExplicitType.java
@@ -18,9 +18,11 @@ package com.goide.inspections;
 
 import com.goide.GoConstants;
 import com.goide.psi.*;
+import com.goide.psi.impl.GoPsiImplUtil;
 import com.goide.psi.impl.GoReferenceExpressionImpl;
 import com.intellij.codeInspection.LocalInspectionToolSession;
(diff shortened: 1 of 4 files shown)


src/com/goide/inspections/GoAssignmentNilWithoutExplicitType.java
```diff
@@ -32,12 +35,38 @@ protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNul
...
+
+      private void checkConst(@NotNull GoConstSpec spec) {
+        if (spec.getType() != null) return;
+        checkExpressions(spec.getExpressionList());
+      }
+
+      private void checkExpressions(@NotNull List<GoExpression> expressions) {
+        for (GoExpression expr : expressions) {
+          if (expr instanceof GoReferenceExpressionImpl) {
+            // todo check if there is 'nil' var/const
```
zolotov: What's this todo about?

eugene-auduchinok: It's a weird thing. If you've declared a variable with name `nil`, say `var nil = 0` then these kinds of assignments are correct.
The following code successfully compiles.
```go
package nil
var nil = 123
var x = nil
```

PR #1792, opened 2015-08-07
Auto import needed packages

eugene-auduchinok:
It can be disabled from Settings | Editor | General | Auto import, however it needs Java auto import enabled.


eugene-auduchinok:
Needs adding tests, though.

src/com/goide/codeInsight/imports/GoCodeInsightSettings.java
```diff
@@ -0,0 +1,44 @@
...
+import com.intellij.openapi.components.*;
+import com.intellij.util.xmlb.XmlSerializerUtil;
+import org.jetbrains.annotations.Nullable;
+
+@State(
+  name = "GoCodeInsightSettings",
+  storages = @Storage(file = StoragePathMacros.APP_CONFIG + "/editor.codeinsight.xml")
+)
+public class GoCodeInsightSettings implements PersistentStateComponent<GoCodeInsightSettings> {
+  public boolean ADD_UNAMBIGUOUS_IMPORTS_ON_THE_FLY = true;
```
zolotov: incapsulate

src/com/goide/psi/impl/imports/GoReferenceImporter.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    final PsiReference reference = file.findReferenceAt(offset);
+    if (reference instanceof GoReference) {
+      GoImportPackageQuickFix fix = new GoImportPackageQuickFix(reference);
+      if (fix.isAvailable(file.getProject(), editor, file)) {
+        WriteCommandAction.runWriteCommandAction(file.getProject(), new Runnable() {
+          @Override
+          public void run() {
+            new GoImportPackageQuickFix(reference).applyFix();
```
zolotov: why do you need second quickFix?
are you sure that fix is still applicable when write actions is started?

src/com/goide/psi/impl/imports/GoReferenceImporter.java
```diff
@@ -0,0 +1,87 @@
...
+    int endOffset = document.getLineEndOffset(lineNumber);
+
+    List<PsiElement> elements = CollectHighlightsUtil.getElementsInRange(file, startOffset, endOffset);
+    for (PsiElement element : elements) {
+      if (element instanceof GoElement) {
+        for (PsiReference reference : element.getReferences()) {
+          if (reference instanceof GoReference) {
+            GoImportPackageQuickFix fix = new GoImportPackageQuickFix(reference);
+            if (fix.isAvailable(file.getProject(), editor, file)) {
+              new GoImportPackageQuickFix(reference).applyFix();
```
zolotov: Are you sure that you don't need write lock here?

src/com/goide/codeInsight/imports/GoAutoImportOptions.java
```diff
@@ -0,0 +1,72 @@
...
+
+  @Nullable
+  @Override
+  public JComponent createComponent() {
+    return myPanel;
+  }
+
+  @Override
+  public boolean isModified() {
+    GoCodeInsightSettings settings = GoCodeInsightSettings.getInstance();
```
zolotov: I believe you can use service auto-injection. Just add GoCodeInsightSettings as a constructor parameter

dlsniper:
@zolotov  / @eugene-auduchinok as a quick question, will this work for the small IDEs / Android Studio as well?

zolotov:
Didn't checked, but there is no any IDEA specific things so it should

dlsniper:
ok, I'll try to investigate later. Thanks

PR #1799, opened 2015-08-14
Fix no new variable inspection for shadowing in 'for clause' case

### LINKED ISSUE #1752: No new variables inspection should account for variable shadowing rules
The code below can run but the go plugin says it contains an error:

```go
package main

import (
	"fmt"
	"strconv"
)

func main() {
	ShouldAllowParameterShadowing("param")
}

func ShouldAllowParameterShadowing(i string) {
	fmt.Println(i)

	for i := 0; i < 3; i++ { // Inspection error "No new variables on left side of :="
		fmt.Println(strconv.Itoa(i))
	}
}
```

But it will run correctly and output:
```
$ go run main.go
param
0
1
2
```

eugene-auduchinok:
Fixes #1752

### DIFF
--- a/src/com/goide/inspections/GoNoNewVariablesInspection.java
+++ b/src/com/goide/inspections/GoNoNewVariablesInspection.java
@@ -16,6 +16,7 @@
 
 package com.goide.inspections;
 
+import com.goide.psi.GoForClause;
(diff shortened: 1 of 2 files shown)


zolotov:
Reverted, iIt wasn't a proper fix

PR #1803, opened 2015-08-16
Better range parsing

zolotov:
Needed for keyword completion contributor (range keyword completion). The problem is that for some reason parsing result are different for single-test running and for `all tests` running, please investigate.


zolotov:
Any updates?

PR #1807, opened 2015-08-17
Optimize imports on the fly

eugene-auduchinok:


src/com/goide/codeInsight/imports/GoCodeInsightSettings.java
```diff
@@ -26,6 +26,7 @@
 )
 public class GoCodeInsightSettings implements PersistentStateComponent<GoCodeInsightSettings> {
   private boolean myShowImportPopup = true;
+  private boolean myOptimizeImportsOnTheFly = false;
```
zolotov: enable by default

src/com/goide/codeInsight/imports/GoOptimizeImportsPass.java
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  @Override
+  protected void collectInformationWithProgress(@NotNull ProgressIndicator progress) {
+    progress.checkCanceled();
+  }
+
+  @Override
+  protected void applyInformationWithProgress() {
+    final Runnable runnable = new GoImportOptimizer().processFile(myFile);
```
zolotov: So are you going to calculate all imports to remove right in EDT? Don't you think it's not really efficient?

ignatov:
Please update copyrights before merge.

nevernet:
what's status now of this PR?

zolotov:
@nevernet it's open and doesn't work good

PR #1810, opened 2015-08-18
ForClause var shadowing resolve fix

### LINKED ISSUE #1752: No new variables inspection should account for variable shadowing rules

eugene-auduchinok:
A fix for wrong variable resolve in for clause. I hope it didn't break anything not covered with tests.
Fixes #1752.

### DIFF
--- a/testData/highlighting/shortVars.go
+++ b/testData/highlighting/shortVars.go
@@ -66,3 +66,17 @@ func Test2() (err error) {
     
     return err
 }
+
+func _(i int) {
+    for i := 1; i < 3; i++ {
+        _ = i
+    }
+
+    if i := 1; i > 0 {
+        _ = i
+    }
+
+    switch i := 1; {
+        case true: _ = i
+    }
+}
--- a/src/com/goide/psi/impl/GoPsiImplUtil.java
+++ b/src/com/goide/psi/impl/GoPsiImplUtil.java
@@ -148,7 +148,7 @@ public class GoPsiImplUtil {
   @Nullable
   public static PsiReference getReference(@NotNull GoVarDefinition o) {
     GoShortVarDeclaration shortDeclaration = PsiTreeUtil.getParentOfType(o, GoShortVarDeclaration.class);
-    boolean createRef = PsiTreeUtil.getParentOfType(shortDeclaration, GoBlock.class, GoIfStatement.class, GoSwitchStatement.class, GoSelectStatement.class) instanceof GoBlock;
+    boolean createRef = PsiTreeUtil.getParentOfType(shortDeclaration, GoBlock.class, GoForStatement.class, GoIfStatement.class, GoSwitchStatement.class, GoSelectStatement.class) instanceof GoBlock;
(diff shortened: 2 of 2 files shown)


eugene-auduchinok:
Squashed in a single commit.

testData/highlighting/shortVars.go
```diff
@@ -65,4 +65,10 @@ func Test2() (err error) {
     }
     
     return err
-}
\ No newline at end of file
+}
+
+func <warning>shadowing</warning>(i int) {
```
ignatov: It's better to use _ as a name for such functions.

src/com/goide/psi/impl/GoVarProcessor.java
```diff
@@ -35,6 +36,7 @@ public GoVarProcessor(@NotNull PsiElement origin, boolean completion) {
   public GoVarProcessor(@NotNull PsiElement requestedName, @NotNull PsiElement origin, boolean completion, boolean delegate) {
     super(requestedName, origin, completion);
     myImShortVarDeclaration = PsiTreeUtil.getParentOfType(origin, GoShortVarDeclaration.class) != null && !delegate;
+    myImInForClause = PsiTreeUtil.getParentOfType(origin, GoForClause.class) != null;
```
ignatov: What's about if/switch/... statements?

testData/highlighting/shortVars.go
eugene-auduchinok: Fixed and amended to the last commit.

src/com/goide/psi/impl/GoVarProcessor.java
eugene-auduchinok: I've checked this and it works without these checks.

ignatov: Please add a test for if statement as well.

ignatov: And for other cases too.

ignatov:
Looks good. So, all those changes should be squashed into one single commit.

src/com/goide/psi/impl/GoVarProcessor.java
```diff
@@ -25,9 +25,9 @@
 
 public class GoVarProcessor extends GoScopeProcessorBase {
   private final boolean myImShortVarDeclaration;
-  private final PsiElement myParentGuard; 
+  private final PsiElement myParentGuard;
```
ignatov: Please revert this file. It doesn't have any changes.

eugene-auduchinok:
Squashed.

PR #1813, opened 2015-08-19
Rename import alias

eugene-auduchinok:


eugene-auduchinok:
Please don't review yet.

ignatov:
Please squash all your commit into one.

eugene-auduchinok:
I've done some resolve fixes and squashed everything to a single commit.

src/com/goide/refactor/GoRenameImportSpecProcessor.java
```diff
@@ -0,0 +1,42 @@
...
+
+import com.goide.psi.GoImportSpec;
+import com.goide.psi.impl.GoPsiImplUtil;
+import com.intellij.openapi.editor.Editor;
+import com.intellij.psi.PsiElement;
+import com.intellij.refactoring.rename.RenamePsiElementProcessor;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class GoRenameImportSpecProcessor extends RenamePsiElementProcessor {
```
zolotov: Why're you need this?

eugene-auduchinok: To handle a dot import, couldn't find other way to prevent it from starting renaming and getting exception in MemberInplaceRenamer.
A possible solution is to resolve a dot symbol *itself* to the import path directory but I don't like the idea and seems like it will lead to another overriding named element logic for import specs.

eugene-auduchinok:
Added dot renaming case test.

src/com/goide/refactor/GoRenameImportSpecProcessor.java
zolotov: > A possible solution is to resolve a dot symbol itself to the import path directory but I don't like the idea and seems like it will lead to another overriding named element logic for import specs.

Yep, it sounds not good. But renaming directory staying on dot is quite strange too. Maybe just restrict renaming of dot? Is this true that if this method returns `null` then rename refactoring is considered as impossible?

eugene-auduchinok: `substituteElementToRename ` doc says:
> @return the element to rename, or null if the rename refactoring should be canceled.

It may show a popup when user tries to rename a dot and return `null` if you agree with such logic.

zolotov: I thinks it's good

eugene-auduchinok:
Added popup for dot import case, squashed into single commit.

PR #1819, opened 2015-08-20
Fix auto import formatting

### LINKED ISSUE #1816: Autoimport doesn't prepend a tab character in the import block
Hi,

I've noticed that in the version b1fe90b7d4474ee9552fddc6045153720303e275 when using the autoimport/import suggestion feature, the resulting import statement will not be prepended with a tab character.

For example:
```go
import (
	"fmt"
)

func demo() {
	io.Copy // when completing this one the issue manifests
}
```

becomes:
```go
import (
	"fmt"
"io"
)

func demo() {
	io.Copy // when completing this one the issue manifests
}
```

eugene-auduchinok:
Fixes #1816.

### DIFF
--- a/src/com/goide/codeInsight/imports/GoImportPackageQuickFix.java
+++ b/src/com/goide/codeInsight/imports/GoImportPackageQuickFix.java
@@ -33,6 +33,7 @@ import com.intellij.codeInspection.LocalQuickFixAndIntentionActionOnPsiElement;
 import com.intellij.openapi.actionSystem.ActionManager;
 import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.command.CommandProcessor;
+import com.intellij.openapi.command.WriteCommandAction;
 import com.intellij.openapi.editor.Editor;
 import com.intellij.openapi.keymap.KeymapUtil;
 import com.intellij.openapi.project.Project;
@@ -220,7 +221,7 @@ public class GoImportPackageQuickFix extends LocalQuickFixAndIntentionActionOnPs
       CommandProcessor.getInstance().runUndoTransparentAction(new Runnable() {
         @Override
         public void run() {
-          ApplicationManager.getApplication().runWriteAction(new Runnable() {
+          WriteCommandAction.runWriteCommandAction(file.getProject(), new Runnable() {
             @Override
             public void run() {
               if (!isAvailable()) return;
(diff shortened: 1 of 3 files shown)


zolotov:
Nested commands looks stupid. Any thought about the reasons of current behaviour?

zolotov:
Any news on this?

eugene-auduchinok:
I'm going to investigate it more deeply tonight.

eugene-auduchinok:
I found that in Java they use similar solution (at least for case when there's an only possible class to import): wrap the actual change into a command and then wrap the command into `UndoTransparentAction`.

The fix action is created in `doFix` in `ImportClassFixBase`.
Inside `addImport` called from action's `execute` they create a command for making changes to documents and post-command stuff, and `execute` call itself is wrapped into a `UndoTransparentAction` when there's the only possible import.
https://github.com/JetBrains/intellij-community/blob/b73f7f9c7a77f57a5ba47c597799f538e1db3de4/java/java-impl/src/com/intellij/codeInsight/daemon/impl/quickfix/ImportClassFixBase.java#L314-L314
https://github.com/JetBrains/intellij-community/blob/c1edd1602746520c63386270a8735f7705fed964/java/java-impl/src/com/intellij/codeInsight/daemon/impl/actions/AddImportAction.java#L201
https://github.com/JetBrains/intellij-community/blob/b73f7f9c7a77f57a5ba47c597799f538e1db3de4/java/java-impl/src/com/intellij/codeInsight/daemon/impl/quickfix/ImportClassFixBase.java#L327-L327

I've tried splitting implementations for cases when a command is needed (on the fly import) and when it's not, but such nesting is still needed for the first case to make it undo transparent and to format the imported code properly.

zolotov:
Sorry, I'm too lazy to explain this once again, beside we've debugged this together and discussed it already. So in short:

> Inside addImport called from action's execute they create a command for making changes to documents and post-command stuff, and execute call itself is wrapped into a UndoTransparentAction when there's the only possible import.

This scenarios covers only single case, there are 4 or 5 approaches to initiate import adding. The funny thing that the second approach is just in the same file: https://github.com/JetBrains/intellij-community/blob/b73f7f9c7a77f57a5ba47c597799f538e1db3de4/java/java-impl/src/com/intellij/codeInsight/daemon/impl/quickfix/ImportClassFixBase.java#L430

> I've tried splitting implementations for cases when a command is needed (on the fly import) and when it's not, but such nesting is still needed for the first case to make it undo transparent and to format the imported code properly.

I see, I'll do it myself then.

PR #1822, opened 2015-08-21
Use correct delve command to debug existing binary

dlsniper:
The delve version that is shipped with the plugin supports the `exec` command which fixes the following bug: currently the plugin uses the `debug` which builds the sources from the current working directory. However the plugin builds the sources correctly by itself in form of a binary file that should be ran by delve.


ignatov:
Thanks.

PR #1826, opened 2015-08-24
Allow to import own path from test file

### LINKED ISSUE #1751: Package name not suggested in auto-completion / import if package name is different than the folder inside test file
I have the following package:

```go
// $GOPATH/src/github.com/dlsniper/u/demo/fuzz/fuzz.go
package fuzy

func DemoFunc(s string) string {
	return s+"1"
}
```

And the test for it
```go
// $GOPATH/src/github.com/dlsniper/u/demo/fuzz/fuzz_test.go
package fuzy_test

import "github.com/dlsniper/u/demo/<caret>" // here there's no package called fuzy

func Fuzz(s string) int {
	fuzy.DemoFunc(s) // this is not flagged as auto-importable
	return 0
}

```

The issue is that in the _test file, I can't get autocomplete/autoimport to work but if I move in a different package and start typing ```fuzy.De``` I get the expected result.

Plugin commit: e15c1a0fe30c3f80d6811aa0215c02400eba9d9d

eugene-auduchinok:
Fixes #1751.

### DIFF
--- a/src/com/goide/codeInsight/imports/GoImportPackageQuickFix.java
+++ b/src/com/goide/codeInsight/imports/GoImportPackageQuickFix.java
@@ -22,6 +22,7 @@ import com.goide.project.GoExcludedPathsSettings;
 import com.goide.psi.GoFile;
 import com.goide.psi.GoReferenceExpression;
 import com.goide.psi.GoTypeReferenceExpression;
+import com.goide.runconfig.testing.GoTestFinder;
 import com.goide.stubs.index.GoPackagesIndex;
 import com.goide.util.GoUtil;
 import com.intellij.codeInsight.hint.HintManager;
@@ -38,6 +39,7 @@ import com.intellij.openapi.project.Project;
 import com.intellij.openapi.ui.popup.JBPopupFactory;
 import com.intellij.openapi.util.Comparing;
 import com.intellij.openapi.util.TextRange;
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiDirectory;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiFile;
@@ -160,8 +162,11 @@ public class GoImportPackageQuickFix extends LocalQuickFixAndIntentionActionOnPs
     if (myPackagesToImport == null) {
       final GlobalSearchScope scope = GoUtil.moduleScope(element);
       PsiFile file = element.getContainingFile();
-      final PsiDirectory parentDirectory = file != null ? file.getParent() : null;
+      boolean isTestFile = GoTestFinder.isTestFile(file) && GoTestFinder.isTestPackageName(((GoFile)file).getPackageName());
+      final String packageName = isTestFile ? ((GoFile)file).getPackageNameWithoutTestSuffix() : null;
+      final boolean allowSamePath = !StringUtil.isEmpty(packageName);
       Project project = element.getProject();
+      final PsiDirectory parentDirectory = file != null ? file.getParent() : null;
       final GoExcludedPathsSettings excludedSettings = GoExcludedPathsSettings.getInstance(project);
       Collection<GoFile> es = StubIndex.getElements(GoPackagesIndex.KEY, myPackageName, project, scope, GoFile.class);
       myPackagesToImport = sorted(skipNulls(map2Set(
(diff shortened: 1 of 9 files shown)


src/com/goide/codeInsight/imports/GoImportPackageQuickFix.java
```diff
@@ -170,7 +172,7 @@ private static boolean notQualified(@Nullable PsiElement startElement) {
           @Nullable
           @Override
           public String fun(@NotNull GoFile file) {
-            String importPath = parentDirectory == null || !parentDirectory.isEquivalentTo(file.getParent()) ? file.getImportPath() : null;
+            String importPath = parentDirectory == null || isFromTestFile || !parentDirectory.isEquivalentTo(file.getParent()) ? file.getImportPath() : null;
```
zolotov: Test?

src/com/goide/completion/GoAutoImportCompletionContributor.java
```diff
@@ -86,7 +87,7 @@ protected void addCompletions(@NotNull CompletionParameters parameters,
         Project project = position.getProject();
         GlobalSearchScope scope = GoUtil.moduleScopeWithoutTests(file);
         VirtualFile containingDirectory = file.getVirtualFile().getParent();
-        if (containingDirectory != null) {
+        if (!GoTestFinder.isTestFile(file) && containingDirectory != null) {
```
zolotov: Test?

eugene-auduchinok:
Added tests, squashed into a single commit.

zolotov:
Two cases has special check `importPath.equals(contextImportPath)`, you've replaced it with `isTest || importPath.equals(contextImportPath)`, so the question is what's special with tests that allows you to do it?

As for GoAutoImportCompletionContributor, why did you decide that if we're invoking completion in test file then we should complete symbols from whole module including current package?

eugene-auduchinok:
According to golang specs: "Test files that declare a package with the suffix "_test" will be compiled as a separate package, and then linked and run with the main test binary."
And if we're allowed to import the package without a "_test" suffix in the same path, why won't we complete it?

eugene-auduchinok:
Amended update with package checking and more tests.

src/com/goide/completion/GoAutoImportCompletionContributor.java
```diff
@@ -306,18 +325,19 @@ public boolean process(GoNamedElement element) {
       return true;
     }
 
+    private boolean samePackage(@NotNull GoFile file, @Nullable String importPath) {
+      //noinspection ConstantConditions
```
zolotov: you rely on some complicated knowledges that are unreachable to IDE and not easy for understanding for developer. I bet it will burst one day.

src/com/goide/completion/GoAutoImportCompletionContributor.java
```diff
@@ -291,11 +307,14 @@ public void setName(@NotNull String name) {
     @Override
     public boolean process(GoNamedElement element) {
       ProgressManager.checkCanceled();
+      GoFile declarationFile = element.getContainingFile();
+      String importPath = declarationFile.getImportPath();
       Boolean allowed = null;
       ExistingImportData importData = null;
+      if (myAllowSamePath && !samePackage(declarationFile, importPath)) return true;
```
zolotov: do not ignore symbols from other packages

src/com/goide/runconfig/testing/GoTestFinder.java
```diff
@@ -39,13 +39,17 @@
...
-    return file != null && file instanceof GoFile && file.getName().endsWith(GoConstants.TEST_SUFFIX_WITH_EXTENSION);
+    return file instanceof GoFile && file.getName().endsWith(GoConstants.TEST_SUFFIX_WITH_EXTENSION);
   }
 
   public static boolean isTestFile(@Nullable VirtualFile file) {
     return file != null && file.getFileType() == GoFileType.INSTANCE && file.getNameWithoutExtension().endsWith(GoConstants.TEST_SUFFIX);
   }
 
+  public static boolean isTestPackageName(@Nullable String packageName) {
+    return packageName != null && packageName.length() > GoConstants.TEST_SUFFIX.length() && packageName.endsWith(GoConstants.TEST_SUFFIX);
```
zolotov: length checking looks redundant

src/com/goide/inspections/GoMultiplePackagesInspection.java
```diff
@@ -32,7 +32,7 @@
 public class GoMultiplePackagesInspection extends GoInspectionBase {
   @Override
   protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsHolder) {
-    if (((ScratchFileType)ScratchFileType.INSTANCE).isMyFileType(file.getVirtualFile())) return;
+    if (file.isScratchFile()) return;
```
zolotov: It's bad idea to include changes that aren't related to review

eugene-auduchinok: Sorry, I had the same check in `GoAutoImportCompletionContributor` until realised that check is not needed.
I will move this change to another pull request.

zolotov: ok

Issue #1461, opened 2015-03-19
Redeclared package import inspection


dlsniper:
the following should produce an error:

```go
import (
	"crypto/rand"
	"math/rand"
)
```

```
./file.go:Line: rand redeclared as imported package name
        previous declaration at ./file.go:Line
```

sjamesr:
I'm having a go at this, it's my first attempt at a change to the go plugin

dlsniper:
That's great, if there's something that we can help out, let us know. And please feel free to open a PR as soon as you have something or you get stuck somewhere ;)

sjamesr:
Oh, this case is apparently already handled by GoUnusedImportDeclaration, I'm just going to add it to GoHighlightingTest unit tests

sjamesr:
The patch is done, before I create a pull request, I have to get it OK'd by Google's open source legal folks.

I should already have a CLA on file with Jetbrains.

dlsniper:
There's no need for the CLA for JetBrains, the plugin is under Apache 2 licence :)

PR #1848, opened 2015-09-08
consider aliases and path names when looking for dupe imports

sjamesr:
Addresses https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1461

### DIFF
--- a/src/com/goide/inspections/GoUnusedImportDeclaration.java
+++ b/src/com/goide/inspections/GoUnusedImportDeclaration.java
@@ -31,6 +31,10 @@ import com.intellij.util.containers.MultiMap;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.Map;
+
 public class GoUnusedImportDeclaration extends GoInspectionBase {
   @Nullable private final static LocalQuickFix OPTIMIZE_QUICK_FIX = new LocalQuickFixBase("Optimize imports") {
     @Override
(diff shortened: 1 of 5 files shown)


ignatov:
Thanks

ignatov:
As for contribution agreement, please download the [PDF](http://www.jetbrains.org/download/attachments/983565/JB_ContributorLicenseAgreement.pdf?version=3&modificationDate=1416937277000) and then print it out, sign (mention you github account name there), scan and email to contribute@jetbrains.com

sjamesr:
Hey Sergey,

Jetbrains should already have a CLA on file for me (sent September 11,
2014). Please let me know if you need another.

Regards,
James

src/com/goide/inspections/GoUnusedImportDeclaration.java
```diff
@@ -85,4 +105,25 @@ public void visitElement(@NotNull PsiElement o) {
       }
     });
   }
+
+  /**
+   * Returns a map of import 'name' to import spec, where name is either the import's alias or the last component of the import's path. The
+   * returned map orders imports the same way as they occur in the file.
+   */
+  private static MultiMap<String, GoImportSpec> getImportNameMap(@NotNull GoFile file) {
```
zolotov: What's the difference with GoFile#getImportMap method?

src/com/goide/inspections/GoUnusedImportDeclaration.java
```diff
@@ -62,6 +69,19 @@ protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsH
     for (GoImportSpec duplicatedImportSpec : GoImportOptimizer.findDuplicatedEntries(importMap)) {
       problemsHolder.registerProblem(duplicatedImportSpec, "Redeclared import", ProblemHighlightType.GENERIC_ERROR, OPTIMIZE_QUICK_FIX);
     }
+
+    // Find imports that are referred to by the same alias or last pathname component.
+    MultiMap<String, GoImportSpec> nameMap = getImportNameMap(file);
```
zolotov: This logic partly duplicates the loop above that is using GoImportOptimizer#findDuplicatedEntries. This leads to duplicated error message
screenshot 2015-09-09 21 44 35

src/com/goide/inspections/GoUnusedImportDeclaration.java
```diff
@@ -62,6 +69,19 @@ protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsH
...
+    for (Map.Entry<String, Collection<GoImportSpec>> specs : nameMap.entrySet()) {
+      // These iterators are guaranteed to be non-empty.
+      Iterator<GoImportSpec> imports = specs.getValue().iterator();
+      // Add warnings to the second and subsequent imports.
+      imports.next();
+      while (imports.hasNext()) {
+        problemsHolder.registerProblem(imports.next(), "Redeclared import", ProblemHighlightType.GENERIC_ERROR, OPTIMIZE_QUICK_FIX);
```
zolotov: OptimizeImports quick fix does nothing with partially matched imports so this problem shouldn't declare it and possible fix.

src/com/goide/inspections/GoUnusedImportDeclaration.java
sjamesr: Should be fixed, now I only suggest a fix if the import wasn't already flagged in the above loop.

src/com/goide/inspections/GoUnusedImportDeclaration.java
sjamesr: Removed, it caused a redeclared import to be detected in another test, can you make sure the test case is still correct?

src/com/goide/inspections/GoUnusedImportDeclaration.java
sjamesr: Done, thanks

sjamesr:
Thank you for the review, the commit should be in better shape now

src/com/goide/inspections/GoUnusedImportDeclaration.java
zolotov: Sure, I'll check

src/com/goide/inspections/GoUnusedImportDeclaration.java
zolotov: Looks like it most consistent solution

Issue #1853, opened 2015-09-08
Unit test icons do not reflect the pass/fail status of the tests in runner


jen20:
Version: 0.9.488

When running unit tests, the icons for the tests do not reflect the pass/fail status of each test as expected.

### Steps to reproduce

1. Add a `.go` file with unit tests
1. Run the tests (via the context menu in this case)
1. Wait for the tests to pass or fail (as reported by the console output and the status bar)

**Expected result:** the icons show green/red for pass/fail as when running JUnit tests in IntelliJ
**Observed result:** the icons remain in the "indeterminate" state as per this screenshot:

<img width="1122" alt="Test Window" src="https://cloud.githubusercontent.com/assets/548524/9741669/a68ff478-5629-11e5-8b21-7d1270642b3c.png">

sjamesr:
I think I have a fix, just waiting for Google legal team to approve the (tiny) patch

PR #1854, opened 2015-09-08
fix leading spaces causing failure to match test name

sjamesr:
Addresses https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1853


zolotov:
Thanks! How about adding test for this case in order to avoid regressions in future? See GotestEventsConverterTest

sjamesr:
Done, please take a look.

PR #1860, opened 2015-09-12
fix bug in redeclared import detector

sjamesr:
The previous version used GoFile.getImportMap(), which returns entries
for every package in a given package source directory, even those with
build tags such as "build +ignore".

This change introduces getImportNameMap, which does no package
resolution and simply looks at names introduced into the current source
file.

### DIFF
--- a/src/com/goide/inspections/GoUnusedImportDeclaration.java
+++ b/src/com/goide/inspections/GoUnusedImportDeclaration.java
@@ -24,17 +24,16 @@ import com.intellij.codeInspection.*;
 import com.intellij.lang.ImportOptimizer;
 import com.intellij.openapi.command.WriteCommandAction;
 import com.intellij.openapi.project.Project;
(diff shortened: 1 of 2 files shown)


zolotov:
> does no package resolution and simply looks at names introduced into the current source
file.

You can't say what names was introduced without resolution, import-path and packages are completely different things, you can't do any decisions about duplicated symbols in scope based on just import path.

zolotov:
Won't merge, sorry, it's just incorrect logic.

sjamesr:
No worries, I didn't realize that import "foo" could cause something other
than "foo" to be brought into the namespace :)

On Sat, Sep 12, 2015 at 9:31 AM, Alexander Zolotov <notifications@github.com

zolotov:
I made quick example, maybe you'll be interested
screen shot 2015-09-12 at 19 40 40

Issue #1858, opened 2015-09-12
Unexpected "Redeclared import error"


vburenin:
Import appears as redeclared.
strconv imported once and used in two places.
Bug didn't show up in 491 build.

![image](https://cloud.githubusercontent.com/assets/4350891/9830194/9eaaaea8-58d9-11e5-9e8d-6bdce78d75d3.png)
![image](https://cloud.githubusercontent.com/assets/4350891/9830240/c69dd62c-58db-11e5-84b3-084731bbc9bd.png)

![image](https://cloud.githubusercontent.com/assets/4350891/9830186/57521bcc-58d9-11e5-97f0-5d610b5435f2.png)
![image](https://cloud.githubusercontent.com/assets/4350891/9830191/6cc10536-58d9-11e5-9f9a-0c9b6f504588.png)

dlsniper:
@vburenin do you have a code sample that can reproduce the bug? Thank you.

sjamesr:
I can reproduce this problem at HEAD, I think the problem is that:

GoFile.getImportMap() will return an import map like

"runtime" -> "import runtime"
"main" -> "import runtime", "import strconv"

even though no explicit "import main" ever occurred. This is because GoFile
calls GoUtil.getAllPackagesInDirectory(dir, true) to resolve all packages
referred to by an import.

The redeclared input check I added in
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/commit/e220fc0ecc79e3a7af836c0909db3b6b0f5ea067
will add a redeclared import warning if the package name has more than one
import spec in the import map.

I'm not sure what the correct solution is. :)

sjamesr:
I think https://github.com/sjamesr/go-lang-idea-plugin/commit/1c92482379861453456fc542ce657c559fe5715e should fix it, I can't think of any other way that doesn't involve treating packages like "main" specially or understanding build tags.

dlsniper:
I'm wondering if we should split this inspection into two different ones, regardless of duplicating the code. The reason is that it will make things easier to disable / enable. What do you guys think?

zolotov:
@sjamesr Why do you think that `main` package should be ignored? What's special with it? Why for other cases `main` package shouldn't be ignored? What if I import some package that has files with `main` package only? Why this package is considered as unimportable? Should we handle this situation too?

In short, I definitely don't want to duplicate code and make any hasty solution just to fix some particular bug.

[...]

PR #1861, opened 2015-09-13
fix spurious "redeclared import" warnings

sjamesr:
This patch modifies GoUtil.getAllPackagesInDirectory to test for build tags
when evaluating whether a file should be scanned for packages.

This also modifies GoFile.hasCPathImport to simply look for 'import "C"'
rather than getImportMap. This avoids an infinite recursion
(getImportMap depends on getAllPackagesInDirectory).

This fixes the bug identified in
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1858

Read on for a brief description of the bug:

In the Go source distribution, you have situations like

src/runtime/mkduff.go -> package main, // build +ignore
src/runtime/trace.go -> package runtime, no build tag

src/strconv/makeisprint.go -> package main, // build +ignore
src/strconv/itoa.go -> package strconv, no build tag

In your program, you do

import "runtime"
import "strconv" <--- redeclared import error

This is because GoUtil.getImportMap returns

main -> [runtime, strconv] <--- two imports with the same name,
                                spurious "redeclared import"
runtime -> [runtime]
strconv -> [strconv]

### DIFF
--- a/src/com/goide/psi/GoFile.java
+++ b/src/com/goide/psi/GoFile.java
@@ -432,7 +432,12 @@ public class GoFile extends PsiFileBase {
   }
 
   public boolean hasCPathImport() {
-    return getImportMap().containsKey(GoConstants.C_PATH);
+    for (GoImportSpec importSpec : getImports()) {
+      if (GoConstants.C_PATH.equals(importSpec.getPath())) {
+        return true;
+      }
+    }
+    return false;
   }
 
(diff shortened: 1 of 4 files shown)


zolotov:
Looks good to me.
There is one more inspection that depends on `getAllPackagesInDirectory` method, it's `GoMultiplePackagesInspection`. Please write the test for this inspection with ignored packages.

sjamesr:
Done, please take a look!

PR #1864, opened 2015-09-15
Change libraries to Go Packages. Fix some typos

dlsniper:
In the interest of not confusing people, I think we should call these `Go packages` rather that `libraries` as the term is better suited for the end-user.

### DIFF
--- a/jps-plugin/src/com/goide/jps/model/JpsGoApplicationLibrariesSerializer.java
+++ b/jps-plugin/src/com/goide/jps/model/JpsGoApplicationLibrariesSerializer.java
@@ -17,7 +17,7 @@
 package com.goide.jps.model;
 
 import com.goide.GoConstants;
-import com.goide.GoLibrariesState;
+import com.goide.GoPathState;
 import com.intellij.util.xmlb.XmlSerializer;
 import org.jdom.Element;
 import org.jetbrains.annotations.NotNull;
@@ -31,7 +31,7 @@ public class JpsGoApplicationLibrariesSerializer extends JpsGlobalExtensionSeria
 
   @Override
   public void loadExtension(@NotNull JpsGlobal global, @NotNull Element componentTag) {
-    GoLibrariesState librariesState = XmlSerializer.deserialize(componentTag, GoLibrariesState.class);
+    GoPathState librariesState = XmlSerializer.deserialize(componentTag, GoPathState.class);
     JpsGoLibrariesExtensionService.getInstance().setApplicationLibrariesState(global, librariesState);
   }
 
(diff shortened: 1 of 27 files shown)


src/com/goide/project/GoLibrariesService.java
```diff
@@ -38,7 +38,7 @@
 import java.util.Set;
 
 public abstract class GoLibrariesService<T extends GoLibrariesState> extends SimpleModificationTracker implements PersistentStateComponent<T> {
-  public static final Topic<LibrariesListener> LIBRARIES_TOPIC = new Topic<LibrariesListener>("libraries changes", LibrariesListener.class);
+  public static final Topic<LibrariesListener> LIBRARIES_TOPIC = new Topic<LibrariesListener>("Go packages changes", LibrariesListener.class);
```
zolotov: LIBRARIES_TOPIC with "go packages changes" is strange

src/com/goide/configuration/GoLibrariesConfigurableProvider.java
```diff
@@ -116,10 +116,10 @@ public String fun(VirtualFile file) {
               return file.getUrl();
             }
           });
-        result.add(new GoLibrariesConfigurable("Global libraries", GoApplicationLibrariesService.getInstance(), urlsFromEnv));
+        result.add(new GoLibrariesConfigurable("Global Go Packages", GoApplicationLibrariesService.getInstance(), urlsFromEnv));
```
zolotov: LibrariesConfigurable and LibrariesService with Go Package name are strange

zolotov: Moreover GOPATH is better name for that

src/com/goide/project/GoLibrariesService.java
dlsniper: I've changed this as the argument name is `displayName` and I thought this might be actually used to display something to the end user. If it's not, I'll revert it.

zolotov: I don't mind of changing its display name, I'm against inconsistent naming

dlsniper: Did you meant the name of the constant itself, from `LIBRARIES_TOPIC` to `GOPATH_TOPIC` (I think it's better suited that `Go Packages` in this context).

zolotov: yes

dlsniper:
Hi, sorry for delay, I've finally managed to get the time to fix everything. Things should be consistently (re)named now both for the end-user and for the internal things as well. I haven't touched the storage naming so that nothing gets mixed up for the users (hopefully). Please let me know if it's ok (or I've went too far). Thank you.

Issue #1866, opened 2015-09-17
Redeclared import path: underscore


SchumacherFM:
On commit d7e7b53f0cc3fc059717fd3d13ebc8fff3485f25

```go
package main

import (
	"image"
	_ "image/gif"
	_ "image/jpeg"
	_ "image/png" // flagged as redeclared import
	_ "image/gif" // flagged as redeclared import
	"image/color"
)

func main() {
	_ = image.YCbCr{}
	_ = color.Alpha{}
}
```
Original report:
I'm using: Latest commit from master, Intelij IDEA 14.1.4, OSX 10.10.5
Please see print screen: 
![screen shot 2015-09-17 at 8 30 16 pm](https://cloud.githubusercontent.com/assets/372027/9942466/6fa19652-5d7b-11e5-9133-653e44c72e2a.png)

dlsniper:
@SchumacherFM do you think you could provide a sample code? I'll try to do something as well meanwhile.

SchumacherFM:
Here you go: https://github.com/SchumacherFM/mediamock/blob/master/csv.go

dlsniper:
Thanks, I've managed to isolate it to something smaller, I've added it to the ticket body.

sjamesr:
My bad, the attached pull request (https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1868) should resolve it.

dlsniper:
@sjamesr awesome, thank you!

zolotov:
@sjamesr btw, if commit message will contain `Fixes issue_id` string then issue will be closed automatically after merge.

sjamesr:
Ahh, good to know, thanks!

On Thu, Sep 17, 2015 at 2:22 PM, Alexander Zolotov <notifications@github.com

SchumacherFM:
Thanks guys! :-)

PR #1868, opened 2015-09-17
Don't look at '_' or '.' imports while looking for redeclarations

sjamesr:
Addresses https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1866.


src/com/goide/inspections/GoUnusedImportDeclaration.java
```diff
@@ -70,6 +70,9 @@ protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsH
     }
 
     for (Map.Entry<String, Collection<GoImportSpec>> specs : importMap.entrySet()) {
+      if (".".equals(specs.getKey()) || "_".equals(specs.getKey())) {
```
zolotov: Please reimplement with invoking isDot and isForSideEffects on imports.next() object

sjamesr: Done, please take a look!

zolotov:
Looks good, thanks

PR #1884, opened 2015-09-20
Improve representation of variables rendered by the debugger

dlsniper:
This attempts to make variables to be displayed nicer when debugging is being used.
It also adds some clarity for numeric types as right now it's hard to distinguish between int and uint64 for example.

Before:
image

After:
image

### DIFF
--- a/src/com/goide/dlv/DlvStackFrame.java
+++ b/src/com/goide/dlv/DlvStackFrame.java
@@ -19,6 +19,7 @@ package com.goide.dlv;
 import com.goide.GoIcons;
 import com.goide.dlv.protocol.DlvApi;
 import com.goide.dlv.protocol.DlvRequest;
+import com.goide.highlighting.GoSyntaxHighlightingColors;
 import com.goide.psi.*;
 import com.intellij.debugger.impl.DebuggerUtilsEx;
 import com.intellij.icons.AllIcons;
@@ -40,8 +41,6 @@ import com.intellij.xdebugger.XDebuggerUtil;
 import com.intellij.xdebugger.XSourcePosition;
 import com.intellij.xdebugger.evaluation.XDebuggerEvaluator;
 import com.intellij.xdebugger.frame.*;
-import com.intellij.xdebugger.frame.presentation.XNumericValuePresentation;
-import com.intellij.xdebugger.frame.presentation.XStringValuePresentation;
 import com.intellij.xdebugger.frame.presentation.XValuePresentation;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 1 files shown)


dlsniper:
At the moment complex types such as channels, maps, slices and structs are not supported yet by delve but once they are, I'll work on those as well.

dlsniper:
@ignatov what do you think, could this be merged as well? Thank you.

ignatov:
What about pointer?
screenshot 2015-09-20 23 17 22

src/com/goide/dlv/DlvStackFrame.java
```diff
@@ -148,20 +147,65 @@ public void computePresentation(@NotNull XValueNode node, @NotNull XValuePlace p
...
       }
 
       @Nullable
       private XValuePresentation getPresentation() {
-        String type = variable.type;
+        final String type = variable.type;
         final String value = variable.value;
-        if (NUMBERS.contains(type)) return new XNumericValuePresentation(value);
-        if ("struct string".equals(type)) return new XStringValuePresentation(value);
+        if (NUMBERS.contains(type)) return new XValuePresentation() {
```
ignatov: Why do you expand this XNumericValuePresentation(value)?

dlsniper: For the type of the value. Right now one can't distinguish between int, uint, int64 and so on, this should help out, see the screenshot.

src/com/goide/dlv/DlvStackFrame.java
```diff
@@ -148,20 +147,65 @@ public void computePresentation(@NotNull XValueNode node, @NotNull XValuePlace p
...
+            @NotNull
+            @Override
+            public String getType() {
+              return type;
+            }
+
             @Override
             public void renderValue(@NotNull XValueTextRenderer renderer) {
-              renderer.renderValue(value);
+              renderer.renderValue(value, GoSyntaxHighlightingColors.BUILTIN_TYPE_REFERENCE);
```
dlsniper: @ignatov I'm not sure about this change, I can revert it if needed.

ignatov:
This PR provides a lot of unused information about types for primitive values. Even in Java we don't show such information yet.
BTW, I'm looking forward to https://groups.google.com/forum/#!topic/delve-dev/Z0lDomkS9LI.

src/com/goide/dlv/DlvStackFrame.java
ignatov: Such type information adds lots of visual noise. Even in Java we don't have such visible information and we don't have any claim about it for many years.
screenshot 2015-09-22 11 17 53

PR #1910, opened 2015-10-08
When running build by TeamCity don't fail build if tests fails

VladRassokhin:
Since failed tests would be reported as problems, there would not be 'exit code 1' problem, and build artifacts (plugin.zip) would be build and published as artifact.
So it would be possible to download plugin even if some tests failed.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,9 @@ test {
     exceptionFormat = 'full'
   }
 }
+if (System.getenv("TEAMCITY_VERSION") != null) {
+    test.ignoreFailures = true
+}
 
 task preparePerformanceTestData << {
   downloadAndUnzip('https://storage.googleapis.com/golang/go1.4.2.src.tar.gz', 'go', 'go')


zolotov:
What? Why? I want to fail build if there are failed tests. And even if I didn't I'd fix this in teamcity build configuration, not in gradle.build with some magic constants.

VladRassokhin:
Main idea was to produce plugin artifact even if some tests failed. If you don't need that, ok.
Hm.. Now I see your point. This could be achieved with `--continue` gradle argument, completely forgot about that.

zolotov:
> Main idea was to produce plugin artifact even if some tests failed. If you don't need that, ok.

Indeed, it's a good point, but I still think that it should be configured at teamcity.

> This could be achieved with --continue gradle argument, completely forgot about that.

Will investigate. Thanks!

zolotov:
Hm, actually I've fixed this by replacing `clean test buildPlugin` with `clean buildPlugin test` :)

Issue #1784, opened 2015-08-02
Output is wrapped in console


zolotov:
Windows

![image](https://cloud.githubusercontent.com/assets/140920/9026148/e37fe4e8-392c-11e5-8291-4ba91535fca6.png)

chiknrice:
fixed in my local, can i pickup this one?

dlsniper:
@chiknrice sure, please send a PR with this.

zolotov:
https://youtrack.jetbrains.com/issue/IDEA-117552

ignatov:
An easy workaround is to set the Terminal shell path to 
`cmd.exe /K "mode con:cols=500 lines=9999&cmd.exe"`.

drathier:
@ignatov I cannot even get that workaround to work on windows 10 :(

markgoodhead:
I also tried that fix and it fixed the issue I also had in https://youtrack.jetbrains.com/issue/IDEA-117552 where the text would be invisible, or rather it made the point at which the text goes invisible much further out than before but that's better than nothing! 

However it didn't change the wrapping behaviour - I still get an output like the behaviour shown in the screenshot at the top of this post. I'm using IDEA 15.0.2 and Plugin 0.10.749 on Windows 7.

zolotov:
@markgoodhead As you can see Go plugin has nothing to do with that, please follow the patform issue at youtrack

markgoodhead:
@zolotov Sorry I'm a little confused - these are two separate issues right? The Invisible terminal text is tracked on the youtrack page and the Wrapping behaviour is tracked here? I got confused with the link to it - I thought it was implied that the issues were related in some way. I was trying to refer to the wrapping behaviour and not the invisible text behaviour.

zolotov:
@markgoodhead both issues are in platform. This issue (#1784) exists just in order to avoid issue duplicates and to notify users when it will fixed in some IDEA release.

dlsniper:
(sorry for noise) I've created the label `depends on IntelliJ platform` to better communicate issues such as this to users.

markgoodhead:
@zolotov Oh OK I understand now, thanks!

zolotov:
Pty is disabled on windows so the issue should be fixed for Go. If you're still have this issue please try to invoke Ctrl+Shift+Alt+/, then select Registry, find `run.processes.with.pty` property there and tell me what's the value of this property

Issue #1507, opened 2015-03-31
Investigate modifying stderr to not be just one color


JoeBruner:
Presently, when running tests, any message that is written to stderr is output in ANSI red.  This prevents libraries for logging such as http://github.com/Sirupsen/logrus from being able to display color coded messages at the prompt.  Writing to stderr from the console with logrus will color as expected.  Writing to stderr is a usefull feature as that's how the built in go `log` library works and by using `os.Stderr` the library is able to be dropped in easily and not break compatibility.

PR #1927, opened 2015-10-18
Fix for #1784

chiknrice:


zolotov:
It breaks #1507

chiknrice:
can you let me know how you found it breaks #1507 as All Tests task run fine - what test could I run to catch these kind of issues in the future, thanks!

zolotov:
I know that because pty processes were introduced initially in order to support coloring sequences in console.

Unfortunately it isn't covered with a test. Test requires running external process, real Go distribution and UI testing, and all these complexity doesn't correspond to potential impact of test.

zolotov:
Anyway, thanks for discovering the connection between pty and #1784 bug, I didn't know that this is the reason.

PR #1938, opened 2015-10-22
Fix IOBE (fixes #1937)

### LINKED ISSUE #1937: Tracking for auto-report 973780
I was using the plugin compiled from master and launched from IDEA into another IDEA instance using the `Go` launch configuration.
```
git rev-parse HEAD
573c609b91f624ef38e86b64aa26bbe8d1eed477
```

dlsniper:
This fixes https://ea.jetbrains.com/browser/ea_reports/973780

### DIFF
--- a/src/com/goide/dlv/breakpoint/DlvBreakpointType.java
+++ b/src/com/goide/dlv/breakpoint/DlvBreakpointType.java
@@ -59,7 +59,7 @@ public class DlvBreakpointType extends XLineBreakpointType<DlvBreakpointProperti
 
   private static boolean isLineBreakpointAvailable(@NotNull VirtualFile file, int line, @NotNull Project project) {
     Document document = FileDocumentManager.getInstance().getDocument(file);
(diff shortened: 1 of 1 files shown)


zolotov:
It's more interesting how did you get line=-1. XSourcePosition#getLine should return -1

dlsniper:
This happened in the following scenario:
- one breakpoint was set in the file with IDEA 14.1.5 and the nightly for that version 
- I've compiled the latest master and run it from IDEA using the `Go` Run Configuration
- opened the same project again in IDEA 15
- I've used Find actions and searched for `Line numb` when this exception happened.

So far this seems to have fixed the issue.

zolotov:
> So far this seems to have fixed the issue.

It doesn't mean that it's proper fix. If it's reproducible there are a lot of other plugins with the same problem, local !=-1 fixing is a wrong way for fixing such issues.

dlsniper:
I'll investigate more tonight, thanks for the explanation. I'll try and get
a consistent way of reproducing this.

ignatov:
Merged manually, thanks

PR #1950, opened 2015-11-01
Used as value inspection (fixes #1887)

### LINKED ISSUE #1887: a = b used as value
As an extension of #1577.

```go
package main

func main() {
	a := 1
	b := 2
	if a = b {  // should produce an error: a = b used as value
		println("a = b")
	} else {
		println("a != b")
	}
}
```

dlsniper:


dlsniper:
I've finished with this PR but for some reason the test fails and when running inside the IDE using the same file, the inspection(s) work correctly.

dlsniper:
I've updated this to also contain a test for the quick fix and everything works fine now from the tests. Please let me know if it's missing anything before it can be merged. Thank you.

dlsniper:
Rebased as well (all should be green now)

testData/highlighting/equalinif.go
```diff
@@ -0,0 +1,29 @@
+package equalinif
+
+func <warning descr="Unused function 'eqlfunc'">eqlfunc</warning>() {
```
zolotov: use _ name in order to avoid redundant inspection reports

src/com/goide/inspections/GoUsedAsValueInCondition.java
```diff
@@ -0,0 +1,122 @@
...
+          String left = StringUtil.join(o.getVarDefinitionList(), GET_TEXT_FUNCTION, ", ");
+          String right = StringUtil.join(o.getExpressionList(), GET_TEXT_FUNCTION, ", ");
+
+          holder.registerProblem(o, left + " := " + right + " used as value", GENERIC_ERROR_OR_WARNING, new GoAssignmentToComparisonQuickFix(o));
+        }
+      }
+    };
+  }
+
+  public static class GoAssignmentToComparisonQuickFix extends LocalQuickFixOnPsiElement {
```
zolotov: You don't need LocalQuickFixOnPsiElement because this fix is used in inspection only and it the same element in inspection report and in quick with. Please reimplement with LocalQuickFix.

Also I think it shouldn't be public

testData/highlighting/equalinif.go
dlsniper: Can you please give me an example? I'm not sure I understand what do you mean. Thank you.

src/com/goide/inspections/GoUsedAsValueInCondition.java
dlsniper: Done.

testData/highlighting/equalinif.go
ignatov: Just replace `<warning descr="Unused function 'eqlfunc'">eqlfunc</warning>` in your test data with `_`. There're lots of examples around, please have a look.

dlsniper: Ah, I thought I need to do something like `<warning descr="_">eqlfunc</warning>`. Done this as well. Thank you.

PR #1954, opened 2015-11-02
change project setup instructions to keep .idea config.

patflynn:
Update the project setup instructions, so that contributors know to replace the .idea configuration that IDEA Gradle import blows away.


zolotov:
thanks

PR #1960, opened 2015-11-04
Display function signature on ctrl+hover (fixes #1713)

### LINKED ISSUE #1713: Modify the CTRL+hover output to show function signature
This has bugged me for quite some time.  Here is an example of holding CTRL while hovering over a method/function in different languages:

Java:
image

Python:
image

Go:
image

As you can see with Java and Python, doing the CTRL hoverover will show what the parameters are to the method and what it might return.  With the Go plugin, all it shows is what file the function came from, which isn't nearly as useful.  This is is especially annoying in languages like go where you can do

    x := foo(y, z)

because I have no idea of telling what x is (is it an int? string?  interface{}?).

dlsniper:


ignatov:
Hmm, will be cool to a have a test for this.

src/com/goide/GoDocumentationProvider.java
```diff
@@ -149,11 +152,23 @@ else if (type != null) {
...
   private static String getParametersAsString(@NotNull GoParameters parameters) {
     return StringUtil.join(GoParameterInfoHandler.getParameterPresentations(parameters), ", ");
   }
+
+  @Nullable
+  @Override
+  public String getQuickNavigateInfo(PsiElement element, PsiElement originalElement) {
+    if (element instanceof GoNamedElement) {
+      String result = getSignature(element);
+      if (result != "") {
```
zolotov: StringUtil#nullize

src/com/goide/GoDocumentationProvider.java
```diff
@@ -46,6 +46,9 @@ public String generateDoc(PsiElement element, PsiElement originalElement) {
       Collection<PsiElement> children = PsiTreeUtil.findChildrenOfType(topLevel, element.getClass());
       boolean alone = children.size() == 1 && children.iterator().next().equals(element);
       String result = getSignature(element);
+      if (result != "") {
+        result = "<p><b>"+result+"</b></p>\n";
```
zolotov: code style

src/com/goide/GoDocumentationProvider.java
```diff
@@ -46,6 +46,9 @@ public String generateDoc(PsiElement element, PsiElement originalElement) {
       Collection<PsiElement> children = PsiTreeUtil.findChildrenOfType(topLevel, element.getClass());
       boolean alone = children.size() == 1 && children.iterator().next().equals(element);
       String result = getSignature(element);
+      if (result != "") {
```
zolotov: do not use != for comparing objects

dlsniper: Force of habit from comparing strings in Go (and lack of practice with Java)

src/com/goide/GoDocumentationProvider.java
dlsniper: Not sure how to use StringUtil#nullize in this context

dlsniper:
@ignatov I'd be happy to add those. Any example for this that I could have a look at? Also, is it too complex to make the parameter types link to their Go types like in the other languages?

ignatov:
Merged manually

PR #1991, opened 2015-11-11
Run go install before gotype (fixes #1990)

### LINKED ISSUE #1990: Gotype should  check an entire package instead of one file.
Right now, if your project has more than one file and plugin prduces errors. That's because currently gotype is only checking selected file without getting types from others, so it doesn't know anything about them. Solution is quite simple - just check the folder and not the file.

In case of `/src/github.com/username/goproject/main.go` should become `/src/github.com/username/goproject`

dlsniper:
In order for `gotype` to work, `go install` has to be run at least once on the package. As such, this PR makes it possible to run it once and then runs `gotype`.

### DIFF
--- a/src/com/goide/actions/tool/GoTypeFileAction.java
+++ b/src/com/goide/actions/tool/GoTypeFileAction.java
@@ -23,6 +23,8 @@ import com.intellij.openapi.vfs.VirtualFile;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.io.File;
+
 public class GoTypeFileAction extends GoDownloadableFileAction {
   public GoTypeFileAction() {
     super("gotype", "golang.org/x/tools/cmd/gotype");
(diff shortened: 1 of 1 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -133,7 +133,6 @@
           <option name="ALIGN_MULTILINE_THROWS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_EXTENDS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_PARENTHESIZED_EXPRESSION" value="true" />
-          <option name="ALIGN_SUBSEQUENT_SIMPLE_METHODS" value="true" />
```
ignatov: I don't think that this file should be committed.

dlsniper: Yes, I've noticed that as well, redoing right now :)

dlsniper: Done :)

zolotov:
1. goinstall may fails on particular package, you can't run it on just `some` directory
2. goinstall might not finished before you run gotype

dlsniper:
> 1. goinstall may fail on particular package, you can't run it on just some directory

Sure, but that's ok, actually. The idea is that if `go install` fails, so will `gotype` which means we can just run `gotype` on the directory / package and have it report the issues.

> 2. goinstall might not finish before you run gotype

That I actually don't know how to fix, please help.

zolotov:
> That I actually don't know how to fix, please help.

Me nether :) E.g. you can place gotype-executor inside consumer of goinstall executor

dlsniper:
I'll close this for now as it doesn't seem we'll have a solution soon.

Issue #1983, opened 2015-11-10
Add inspection to validate structure tag format.


stanislav-bios-baranov:

zolotov:
@stanislav-bios-baranov could you please attach a code sample for struct tags that should be highlighted by the new inspection?

dlsniper:
@zolotov I've updated the issue to include some examples of broken tags, hope it helps :)

stanislav-bios-baranov:
@dlsniper Excellent example, thank you. I have nothing to add.

sjamesr:
Just FYI, I'm trying to implement this one now.

zolotov:
@sjamesr good to know, thanks

sjamesr:
Incidentally, do we have any code in the plugin that implements [strconv.Unquote](https://golang.org/pkg/strconv/#Unquote)?

zolotov:
@sjamesr there is a platform method StringUtil#unquoteString(String, char). Also you can extract one from GoPsiImplUtil#getPath(com.goide.psi.GoImportString), just keep in mind that GoPsiImplUtil#getPath works also for non-completed quoted strings (without start or end quote).

PR #1996, opened 2015-11-14
add an inspection for StructTag

sjamesr:
This is an incomplete implementation of
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1983. It will
raise warnings if it detects a tag value that does not conform to the Go
StructTag convention described by https://golang.org/pkg/reflect/#StructTag.

Currently, the inspector does not validate that the tag value strings are valid
and is therefore overly permissive.

Its implementation is essentially ported from validateStructTag in
https://github.com/golang/tools/blob/master/cmd/vet/structtag.go.

### DIFF
--- a/src/com/goide/inspections/GoStructTagInspection.java
+++ b/src/com/goide/inspections/GoStructTagInspection.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.GoFieldDeclaration;
+import com.goide.psi.GoStructType;
+import com.goide.psi.GoTag;
+import com.goide.psi.GoVisitor;
+import com.goide.psi.impl.GoPsiImplUtil;
+import com.goide.util.GoStringLiteralEscaper;
+import com.intellij.codeInspection.LocalInspectionToolSession;
+import com.intellij.codeInspection.ProblemsHolder;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Implements <a href="https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1983"/>, an
+ * inspector that warns if a go StructTag is not well-formed according to Go language conventions.
+ */
+public class GoStructTagInspection extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder,
+                                     @SuppressWarnings({"UnusedParameters", "For future"}) @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitStructType(@NotNull GoStructType o) {
+        for (GoFieldDeclaration field : o.getFieldDeclarationList()) {
+          if (field.getTag() == null) {
+            continue;
+          }
+          if (!isValidTag(field.getTag())) {
+            holder.registerProblem(field.getTag(), "Bad syntax for struct tag value");
+          }
+        }
+      }
+    };
+  }
+
+  // Implementation based on validateStructTag from the go vet tool:
+  // https://github.com/golang/tools/blob/master/cmd/vet/structtag.go.
+  private boolean isValidTag(GoTag tag) {
+    StringBuilder tagText = new StringBuilder(GoPsiImplUtil.unquote(tag.getText()));
+    if (tagText.length() != tag.getText().length() - 2) {
+      // In this case, the tag was not quoted and therefore is invalid.
+      return false;
+    }
+    while (tagText.length() > 0) {
+      int i;
+      i = 0;
+
+      // Skip leading spaces.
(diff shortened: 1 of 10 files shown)


sjamesr:
Here's what it looks like:

screen shot 2015-11-14 at 10 57 48 am

src/com/goide/inspections/GoStructTagInspection.java
```diff
@@ -0,0 +1,107 @@
...
+          i++;
+        }
+        i++;
+      }
+
+      if (i >= tagText.length()) {
+        return false;
+      }
+
+      // TODO(sjamesr): check strconv.Unquote(quotedValue)
```
ignatov: Please use GoPsiImplUtil#unquote

src/com/goide/inspections/GoStructTagInspection.java
```diff
@@ -0,0 +1,107 @@
...
+                                     @SuppressWarnings({"UnusedParameters", "For future"}) @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitStructType(@NotNull GoStructType o) {
+        for (GoFieldDeclaration field : o.getFieldDeclarationList()) {
+          if (field.getTag() == null) {
+            continue;
+          }
+          if (!isValidTag(field.getTag())) {
+            holder.registerProblem(field.getTag(), "Bad syntax for struct tag value");
```
ignatov: Will be cool to provide a helpful information in addition to "Bad syntax for struct tag value".

ignatov:
Looks good to me, thanks a lot

src/com/goide/inspections/GoStructTagInspection.java
sjamesr: As I understand it, that method isn't equivalent to strconv.Unquote because it won't resolve escape sequences within the string. Also strconv.Unquote will return errors if it finds mismatched quotes, bad escape sequences, etc.

What should I do? Right now I'm tempted just to do:
    quotedValue.size() >= 2 && quotedValue.charAt(0) == quotedValue.charAt(quotedValue.size() - 1) && isQuote(quotedValue.charAt(0))

This will give us a pretty good start.

Is `GoStringLiteralEscaper#decode` appropriate? We could decode what's inside the quotes using that method, if it returns true, assume the quoted value is valid.

src/com/goide/inspections/GoStructTagInspection.java
sjamesr: I added a longer description, when the user clicks the "more" button in the tooltip, they'll get a link to the struct tag conventions as well as the description of the conventions.

Did you have anything else in mind?

src/com/goide/inspections/GoStructTagInspection.java
sjamesr: Actually, that check isn't necessary, we already know it's a "-quoted string

src/com/goide/inspections/GoStructTagInspection.java
ignatov: You're right, I missed the second paragraph in GoStructTag.html

src/com/goide/inspections/GoStructTagInspection.java
ignatov: OK, you can create a separate method for your needs, but it's useful to keep them close to each other – for the future refactoring.

sjamesr: Do you think it's good to submit as-is with the TODO in place? I think in practice the current check will do quite a good job.

ignatov: Yes, we can merge it with a todo.

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -373,6 +374,21 @@ public VirtualFile compute() throws Throwable {
...
+                                      + "    WrongTag2           string <warning>`json:\"demo`</warning>\n"
+                                      + "    WrongTag3           string <warning>`json:demo`</warning>\n"
+                                      + "    WrongTag4           string <warning>` json:demo`</warning>\n"
+                                      + "    WrongTag5           string <warning>\"json:demo\"</warning>\n"
+                                      + "    WrongTag6           string <warning>\"json:'demo'\"</warning>\n"
+                                      + "    WrongTag7           string <warning>`json:\"demo\"; bson:\"demo\"`</warning>\n"
+                                      + "}");
+    myFixture.checkHighlighting();
+  }
+
```
ignatov: Please move this code snippet to the separate file in testData/highlighting folder.

ignatov: Please add an unclosed tag: ```WrongTag8 string `json:```

src/com/goide/inspections/GoStructTagInspection.java
```diff
@@ -0,0 +1,107 @@
...
+          }
+        }
+      }
+    };
+  }
+
+  // Implementation based on validateStructTag from the go vet tool:
+  // https://github.com/golang/tools/blob/master/cmd/vet/structtag.go.
+  private boolean isValidTag(GoTag tag) {
+    StringBuilder tagText = new StringBuilder(tag.getText().substring(1, tag.getTextLength() - 1));
```
zolotov: I insist on using GoPsiImplUtil#unquote. Remember that IDE has to handle incomplete go-code, this `substring` is unsafe and can produce IOOBE e.g. on this code:

```go
package demo;

type demoBrokenTags struct {
	WrongTag9           string `
```

tests/com/goide/inspections/GoHighlightingTest.java
sjamesr: Done and done!

src/com/goide/inspections/GoStructTagInspection.java
sjamesr: Ahh, I didn't remember that we could be running on invalid Go code, my bad! Should be fixed now, please take another look.

src/com/goide/inspections/GoStructTagInspection.java
sjamesr: I probably added it after you looked at it! :)

Issue #2002, opened 2015-11-16
Inspection for "continue is not in a loop"


dlsniper:
```go
func demo(a bool) {
	if a {
		continue // should be an error: continue is not in a loop
	} 
}
```

The quick fix should offer to convert this to a return statement.

sjamesr:
Should we offer the quick fix only if the continue inside a function returning void?

dlsniper:
For now I think it would be good enough to let the user add the missing return parameters as a return statement without all the params will be flagged as error anyway.

PR #2005, opened 2015-11-16
Add 'replace with return' quick fix

sjamesr:
This change adds a 'replace with return' quick fix and addresses
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2002.

Not sure if this is a worthwhile quickfix (not hard to just type 'return')...


zolotov:
Test please. Use `GoRenameVarQuickFixTest` as an example.

sjamesr:
Done, please take another look.

Complements on the test framework, it's really nice!

zolotov:
Thanks

PR #2020, opened 2015-11-21
Update docker to the latest version

dlsniper:
As Docker 1.5 is deprecated and soon docker hub will stop working with 1.5 version, I think this version should be updated. Please see the following link for details: https://blog.docker.com/2015/10/docker-hub-deprecation-1-5/
Hope it helps.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,8 +98,8 @@ test {
 }
 
 task preparePerformanceTestData << {
-  downloadAndUnzip('https://storage.googleapis.com/golang/go1.4.2.src.tar.gz', 'go', 'go')
-  downloadAndUnzip('https://github.com/docker/docker/archive/v1.5.0.tar.gz', 'docker-1.5.0', 'docker')
+  downloadAndUnzip('https://storage.googleapis.com/golang/go1.5.1.src.tar.gz', 'go', 'go')
+  downloadAndUnzip('https://github.com/docker/docker/archive/v1.9.1.tar.gz', 'docker-1.9.1', 'docker')
 }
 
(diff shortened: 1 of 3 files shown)


zolotov:
Broken tests: https://teamcity.jetbrains.com/viewLog.html?buildId=632459&tab=buildResultsDiv&buildTypeId=IntellijIdeaPlugins_Go_PerformanceTests

dlsniper:
Ah, I think I should have read the sources better, Go and Docker are used only for the source code not for the actual functionality.
Let me know if I should add the changes to fix this or not in this context.
And as a final question, shouldn't kubernetes be added as part of the tests as the kubernetes source tree is about 2x larger in number of files and 3x larger in terms of lines of code?

zolotov:
> Let me know if I should add the changes to fix this or not in this context.

Updating docker makes sense probably. Not for performance testings but for inspections. Anyway if you want to change the testdata you should check all mismatched inspection reports: maybe by updating docker you revealed some bug. 

> And as a final question, shouldn't Kubernetes be added as part of the tests as the kubernetes source tree is about 2x larger in number of files and 3x larger in terms of lines of code?

I think we're fine with docker sources so far.

dlsniper:
I've updated the files with the tests. I've sampled a few results for correctness and it looked ok. Unfortunately from what I can see the TeamCity integration doesn't work if I to a squash and a force push into the branch (or I don't know where to look for it at least).

dlsniper:
Disregard the TeamCity part, it seems it takes a bit to queue it. Inow see a build in pending state.

zolotov:
So have you figured out why there is a unused import?

dlsniper:
Yes. The unused import actually is a redundant import declaration: ` import math "math"`. The description of the error was properly done.

PR #1603, opened 2015-05-03
Duplicate function and method inspections combined into one class

linniksa:

PR #2021, opened 2015-11-21
Duplicate function and method inspections combined into one class

dlsniper:
Continuing on the work from @linniksa on https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1603


dlsniper:
Done, tests should be updated now and all should work.

src/com/goide/inspections/GoDuplicateFunctionOrMethodInspection.java
```diff
@@ -33,19 +36,56 @@
...
     Project project = file.getProject();
     String packageName = file.getPackageName();
     GlobalSearchScope scope = GoPsiImplUtil.packageScope(file);
 
+    for (final GoFunctionDeclaration func : file.getFunctions()) {
+      if (func.isBlank()) continue;
+
+      final String funcName = func.getName();
+      if (funcName == null) continue;
+      if (INIT.equals(funcName) & zeroArity(func)) continue;
```
zolotov: & => &&

src/com/goide/inspections/GoDuplicateFunctionOrMethodInspection.java
```diff
@@ -33,19 +36,56 @@
 
 import java.util.Collection;
 
-public class GoDuplicateMethodInspection extends GoInspectionBase {
+import static com.goide.GoConstants.INIT;
+import static com.goide.GoConstants.MAIN;
+
+public class GoDuplicateFunctionOrMethodInspection extends GoInspectionBase {
   @Override
   protected void checkFile(@NotNull GoFile file, @NotNull ProblemsHolder problemsHolder) {
```
zolotov: Shouldn't be check file, build a visitor

dlsniper:
Done. I've made a separate commit so that we still keep the original contribution.

PR #2023, opened 2015-11-21
Use symlinked path for breakpoints (fix #1880)

### LINKED ISSUE #1880: Debugging in symlinked project paths doesn't work
Debugging doesn't work in folders with symbolic links. Step-by-step guide:
1) Create some folder:
```
mkdir ~/SomeFolder
```
2) Create symbolic link to this folder:
```
ln -s ~/SomeFolder ~/LinkToSomeFolder
```
3) Create simple project in ~/LinkToSomeFolder
4) Try to debug this project
5) Make sure that application doesn't stop on breakpoints

---
Original report:

When i run debug configuration for application, placed in the path with cyrillic characters, application doesn't stop on breakpoints. Same configuration works well in latin characters named path.

dlsniper:
Go won't store the target path of the symlink, so for example if the file is in `/home/florin/golang/src/github.com/dlsniper/u/demo/dbg/demo1/demo1.go` and the project is under `/home/florin/demo1` which points to `/home/florin/golang/src/github.com/dlsniper/u/demo/dbg/demo1` then when building `/home/florin/demo1` then the Go compiler will store the `/home/florin/demo1/demo1.go` not `/home/florin/golang/src/github.com/dlsniper/u/demo/dbg/demo1/demo1.go`. I'm sorry I can't explain it better, but this PR fixes the issue described in #1880 (I've tested locally for it). Thank you.

### DIFF
--- a/src/com/goide/dlv/DlvDebugProcess.java
+++ b/src/com/goide/dlv/DlvDebugProcess.java
@@ -264,7 +264,7 @@ public final class DlvDebugProcess extends DebugProcessImpl<RemoteVmConnection> 
       if (breakpointPosition == null) return;
       VirtualFile file = breakpointPosition.getFile();
       int line = breakpointPosition.getLine();
-      send(new DlvRequest.CreateBreakpoint(file.getCanonicalPath(), line + 1))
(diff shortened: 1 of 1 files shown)


ignatov:
Thanks, reproduced.

PR #2024, opened 2015-11-21
Add inspection for using defer in a for loop (fixes #1776)

### LINKED ISSUE #1776: Add inspection for doing a defer in an infinite loop
There should be an inspection for the following case:

```go
for {
    src, err := os.Open(srcName)
    defer src.Close() // this should be marked with a warning
}
```

```go
for {
    src, err := os.Open(srcName)
    if err != nil {
       break
    }
    defer src.Close() // this should be marked with a weak warning because there's a break in the for loop
}
```

The problem:
- os.Open requires the resource to be closed via ```Close()``` method
- if the defer statement is placed in an infinite for loop ```for { ... } ``` then the deferred statement will never be called thus resulting in resources which will be leaking / can't be properly closed / gc-ed

To be more clear, the defer inside the loop is the problem, the object might not have a ```Close()``` method, it might be something totally different.

dlsniper:
This adds the inspection that checks if a defer is called inside a for loop (which could potentially cause a resource leak).

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -226,6 +226,9 @@
     <localInspection language="go" displayName="Used as value in condition"
                      groupName="Control flow issues" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoUsedAsValueInCondition"/>
+    <localInspection language="go" displayName="Defer in loop"
+                     groupName="Control flow issues" enabledByDefault="true" level="WEAK WARNING"
+                     implementationClass="com.goide.inspections.GoDeferInLoop"/>
     <!-- /control flow issues -->
 
     <!-- general -->
--- a/src/com/goide/inspections/GoDeferInLoop.java
+++ b/src/com/goide/inspections/GoDeferInLoop.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.*;
+import com.intellij.codeInspection.LocalInspectionToolSession;
+import com.intellij.codeInspection.ProblemHighlightType;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
+import org.jetbrains.annotations.NotNull;
+
+public class GoDeferInLoop extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitDeferStatement(@NotNull GoDeferStatement o) {
+        PsiElement parent = PsiTreeUtil
+          .getParentOfType(o, GoForStatement.class, GoFunctionLit.class);
+
+        if (parent instanceof GoForStatement) {
+          holder.registerProblem(o, "Possible resource leak, \"defer\" is called in a for loop.", ProblemHighlightType.WEAK_WARNING);
+        }
+      }
+    };
+  }
+}
--- a/tests/com/goide/inspections/GoHighlightingTest.java
+++ b/tests/com/goide/inspections/GoHighlightingTest.java
@@ -57,7 +57,8 @@ public class GoHighlightingTest extends GoCodeInsightFixtureTestCase {
       GoRedeclareImportAsFunctionInspection.class,
       GoStructTagInspection.class,
       GoContinueNotInLoopInspection.class,
-      GoUsedAsValueInCondition.class
+      GoUsedAsValueInCondition.class,
+      GoDeferInLoop.class
     );
   }
 
@@ -286,6 +287,10 @@ public class GoHighlightingTest extends GoCodeInsightFixtureTestCase {
   public void testCGOImportInTestFile() {
     myFixture.configureByText("a_test.go", "package a; import<error>\"C\"</error>; import<error>\"C\"</error>;");
     myFixture.checkHighlighting();
+  }
+
+  public void testDeferInLoop() {
+    myFixture.testHighlighting(true, false, true, getTestName(true) + ".go");
   }
 
   public void testCGOImportInNonTestFile() {
--- a/resources/inspectionDescriptions/GoDeferInLoop.html
+++ b/resources/inspectionDescriptions/GoDeferInLoop.html
@@ -1,0 +1,21 @@
+<!--
+  ~ Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+  ~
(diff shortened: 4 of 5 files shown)


dlsniper:
I've flagged the inspection as sending back a `WEAK_WARNING` but it seems that the tests are failing if I do that. If I set the inspection to return the issue as a `GENERIC_ERROR_OR_WARNING` then the tests pass.

The reason behind the weak warning is that there might be cases where calling defer inside of a loop make sense.

I've also tried to abort the checking of the inspection earlier (exit if the checker encounters a method / function declaration) but I think there are some other cases which I didn't catched so I've decided to add the `GoFile` as a catch all for the rest of exit cases. If this proves to be too slow, it could be further optimized.

resources/META-INF/plugin.xml
```diff
@@ -226,6 +226,9 @@
     <localInspection language="go" displayName="Used as value in condition"
                      groupName="Control flow issues" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoUsedAsValueInCondition"/>
+    <localInspection language="go" displayName="Defer in loop"
+                     groupName="Control flow issues" enabledByDefault="true" level="WARNING"
```
zolotov: WEAK_WARNING

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -118,6 +119,7 @@ protected boolean isWriteActionRequired() {
   public void testStructTags()                { doTest(); }
   public void testContinue()                  { doTest(); }
   public void testEqualinif()                 { doTest(); }
+  public void testDeferInLoop()               { doTest(); }
```
zolotov: doTest() doesn't check weak_warnings

src/com/goide/inspections/GoDeferInLoop.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class GoDeferInLoop extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitDeferStatement(@NotNull GoDeferStatement o) {
+        PsiElement parent = o.getParent();
+        while (!((parent instanceof GoFunctionDeclaration) ||
```
zolotov: PsiTreeUtil.getParentOfType(o, GoForStatement.class)

resources/META-INF/plugin.xml
dlsniper: I've already tried that but I get the following error:
```
[  27275]  ERROR - ij.codeInspection.InspectionEP - Can't find highlight display level: WEAK_WARNING 
java.lang.Throwable
	at com.intellij.openapi.diagnostic.Logger.error(Logger.java:115)
	at com.intellij.codeInspection.InspectionEP.getDefaultLevel(InspectionEP.java:129)
	at com.intellij.codeInspection.ex.InspectionToolWrapper.getDefaultLevel(InspectionToolWrapper.java:155)
	at com.intellij.codeInspection.ex.InspectionProfileImpl.addTool(InspectionProfileImpl.java:607)
	at com.intellij.codeInspection.ex.InspectionProfileImpl.initialize(InspectionProfileImpl.java:553)
	at com.intellij.codeInspection.ex.InspectionProfileImpl.initInspectionTools(InspectionProfileImpl.java:535)
	at com.intellij.codeInspection.ex.InspectionProfileImpl.getAllEnabledInspectionTools(InspectionProfileImpl.java:494)
	at com.intellij.codeInspection.ex.InspectionProfileWrapper.init(InspectionProfileWrapper.java:87)
	at com.intellij.profile.codeInspection.InspectionProjectProfileManagerImpl.initProfileWrapper(InspectionProjectProfileManagerImpl.java:165)
	at com.intellij.profile.codeInspection.InspectionProjectProfileManagerImpl$1$1.run(InspectionProjectProfileManagerImpl.java:129)
	at com.intellij.openapi.application.impl.ApplicationImpl$8.run(ApplicationImpl.java:365)
	at java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:511)
	at java.util.concurrent.FutureTask.run(FutureTask.java:266)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
	at org.jetbrains.ide.PooledThreadExecutor$1$1.run(PooledThreadExecutor.java:55)
[  27276]  ERROR - ij.codeInspection.InspectionEP - IntelliJ IDEA 15.0  Build #IC-143.381.42 
[  27276]  ERROR - ij.codeInspection.InspectionEP - JDK: 1.8.0_66 
[  27276]  ERROR - ij.codeInspection.InspectionEP - VM: Java HotSpot(TM) 64-Bit Server VM 
[  27276]  ERROR - ij.codeInspection.InspectionEP - Vendor: Oracle Corporation 
[  27276]  ERROR - ij.codeInspection.InspectionEP - OS: Linux 
[  27276]  ERROR - ij.codeInspection.InspectionEP - Last Action:  
```

zolotov: You're right. Should be `WEAK WARNING`

src/com/goide/inspections/GoDeferInLoop.java
dlsniper: If I would use that one I need to understand how to make it check if the defer clause is contained by a closure for example. I've updated the code to and the test to reflect that case.

tests/com/goide/inspections/GoHighlightingTest.java
dlsniper: Updated the test to run in a different way. Thank you for pointing out the info about `doTest`

src/com/goide/inspections/GoDeferInLoop.java
zolotov: PsiTreeUtil.getParentOfType(o, GoForStatement.class, GoFunctionLit.class) instanceof GoForStatement

resources/META-INF/plugin.xml
dlsniper: Done, thanks

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -288,6 +289,11 @@ public void testCGOImportInTestFile() {
     myFixture.checkHighlighting();
   }
 
+  public void testDeferInFor() {
+    myFixture.configureByFile("deferInLoop.go");
```
zolotov: myFixture.testHighlighting(true, false, true, getTestName(true) + ".go");

src/com/goide/inspections/GoDeferInLoop.java
dlsniper: Done, thank you for helping me with this one.

src/com/goide/inspections/GoDeferInLoop.java
```diff
@@ -0,0 +1,44 @@
...
+public class GoDeferInLoop extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitDeferStatement(@NotNull GoDeferStatement o) {
+        PsiElement parent = PsiTreeUtil
+          .getParentOfType(o, GoForStatement.class,
+                           GoFunctionLit.class, GoFunctionDeclaration.class, GoMethodDeclaration.class, GoFile.class);
```
zolotov: Why do you need (GoFunctionDeclaration.class, GoMethodDeclaration.class, GoFile.class)? Is this some optimization? If so then I think it isn't needed.

dlsniper:
Done, thank you for your help!

src/com/goide/inspections/GoDeferInLoop.java
dlsniper: I see, I've read the source for `getParentOfType(element, ...classes)`  and adjusted the code accordingly.

zolotov:
Manually merged. Please see cleanup commit, there is an important fix with highlighting type

PR #2025, opened 2015-11-22
fix 'continue not in a loop' inspector

sjamesr:
Previously the inspector would look for the presence of a GoForStatement
parent of a continue statement. This did not take into account the
presence of 'go' or 'defer' calls, e.g.

```go
for {
    defer func() {
        continue // no error was raised, should be an error
    }()
}
```

The discussion at
https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2024
reminded me that the continue inspection suffers from this problem.


ignatov:
Thanks

PR #2027, opened 2015-11-22
Highlight exit points of a function when cursor is on `func`

ilinum:
If cursor is on func keyword highlight all the exit points of this function (return and panic statements)

I don't know if this feature is really needed, it is up to you whether to accept this PR.
However, such highlighting is supported in Scala plugin for IDEA, and I find it to be really useful.

func-cursor-screenshot

### DIFF
--- a/src/com/goide/highlighting/GoHighlightExitPointsHandlerFactory.java
+++ b/src/com/goide/highlighting/GoHighlightExitPointsHandlerFactory.java
@@ -99,9 +99,7 @@ public class GoHighlightExitPointsHandlerFactory extends HighlightUsagesHandlerF
         LeafPsiElement leaf = (LeafPsiElement)element;
         return leaf.getElementType() == GoTypes.RETURN || leaf.getElementType() == GoTypes.FUNC || isPanicCall(leaf);
       }
-      else {
-        return false;
-      }
+      return false;
     }
 
     private static boolean isPanicCall(@NotNull PsiElement e) {
--- a/src/com/goide/highlighting/GoHighlightExitPointsHandlerFactory.java
+++ b/src/com/goide/highlighting/GoHighlightExitPointsHandlerFactory.java
@@ -63,6 +63,9 @@ public class GoHighlightExitPointsHandlerFactory extends HighlightUsagesHandlerF
 
     @Override
     public void computeUsages(List<PsiElement> targets) {
+      if (myTarget instanceof LeafPsiElement && ((LeafPsiElement)myTarget).getElementType() == GoTypes.FUNC) {
+        addOccurrence(myTarget);
+      }
       new GoRecursiveVisitor() {
         @Override
         public void visitFunctionLit(@NotNull GoFunctionLit literal) {
@@ -85,10 +88,20 @@ public class GoHighlightExitPointsHandlerFactory extends HighlightUsagesHandlerF
     public static MyHandler createForElement(@NotNull Editor editor, PsiFile file, PsiElement element) {
       GoTypeOwner function = PsiTreeUtil.getParentOfType(element, GoFunctionLit.class, GoFunctionOrMethodDeclaration.class);
       if (function == null) return null;
-      if (element instanceof LeafPsiElement && ((LeafPsiElement)element).getElementType() == GoTypes.RETURN || isPanicCall(element)) {
+      if (shouldCreateMyHandler(element)) {
         return new MyHandler(editor, file, element, function);
       }
       return null;
+    }
+
+    private static boolean shouldCreateMyHandler(PsiElement element) {
+      if (element instanceof LeafPsiElement) {
+        LeafPsiElement leaf = (LeafPsiElement)element;
+        return leaf.getElementType() == GoTypes.RETURN || leaf.getElementType() == GoTypes.FUNC || isPanicCall(leaf);
+      }
+      else {
+        return false;
+      }
     }
 
     private static boolean isPanicCall(@NotNull PsiElement e) {
--- a/tests/com/goide/editor/GoExitPointsHighlightingTest.java
+++ b/tests/com/goide/editor/GoExitPointsHighlightingTest.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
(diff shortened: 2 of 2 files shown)


dlsniper:
Hi, thank you for the contribution, functionality wise it looks good. Does it work for methods (`func (a demo) Demo() {} `) as well?

ilinum:
Yes, it does

demo

ilinum:
However, here are two cases I just realized don't work so well. I'm not sure they're so bad though. Maybe this is a feature :)

not-so-good
not-so-good2

dlsniper:
Awesome. Thanks. I guess @ignatov / @zolotov can give you a proper code review. I've tried to search for tests of this functionality on github but I don't seem to be able to find any.

ilinum:
Cool!

I don't think there are any tests
I know that there is none for the same (or similar) feature in [Scala plugin](https://github.com/JetBrains/intellij-scala/commit/57841e8bebce13a5c2e4792474f86c1e9d3b6652)

src/com/goide/highlighting/GoHighlightExitPointsHandlerFactory.java
```diff
@@ -85,12 +88,22 @@ public void visitCallExpr(@NotNull GoCallExpr o) {
...
       }
       return null;
     }
 
+    private static boolean shouldCreateMyHandler(PsiElement element) {
+      if (element instanceof LeafPsiElement) {
+        LeafPsiElement leaf = (LeafPsiElement)element;
+        return leaf.getElementType() == GoTypes.RETURN || leaf.getElementType() == GoTypes.FUNC || isPanicCall(leaf);
+      }
+      else {
```
ignatov: A redundant else statement

ilinum: Do you mean that this method can be simplified to one big return statement?

I think it is more readable to have it as an if-else satatement

zolotov: No, just else statement is redundant, not whole if

ilinum: Oh, I see 
I'll fix it later today

zolotov:
That's how to test:

```java
HighlightUsagesHandlerBase<PsiElement> handler = HighlightUsagesHandler.createCustomHandler(myEditor, myFile);
assertNotNull(handler);
List<PsiElement> targets = handler.getTargets();
assertEquals(1, targets.size());
assertEquals("func", targets.get(0).getText());
handler.computeUsages(targets);
List<TextRange> readUsages = handler.getReadUsages();
// assert readUsages
```

ilinum:
So I removed the redundant `else` statement

I tried to write tests for it but it wasn't simple. The problem is that before creating MyHandler in MyHandler#createForElement, you go up the tree, which causes trouble in tests, because all of the element's parents are of type GeneratedParserUtilBase$DummyBlock.

So if you really feel that this issue should be tested, I'll get back to it in the next couple of days. But, in my opinion, this is not a very complex or big feature. Besides, judging from other languages, not a lot of changes will be made to it, so regression tests do not make a lot of sense. But, of course, you never know :)
I also understand that this logic is bad and you should never questioning writing tests :)

zolotov:
@ilinum I didn't understand the complexity of testing this. Could you publish the example of test you've tried to make?

In ideal world, since you have an access to intellij repository, the proper solution is to make this tests easy to write and patch all existing tests right in platform.

ilinum:
```
private void doTest(@NotNull String text, String... usages) {
  myFixture.configureByText("foo.go", text);
  HighlightUsagesHandlerBase<PsiElement> handler = HighlightUsagesHandler.createCustomHandler(myFixture.getEditor(), myFixture.getFile());
  assertNotNull(handler);
  List<PsiElement> targets = handler.getTargets();
  assertEquals(1, targets.size());

  handler.computeUsages(targets);
  List<TextRange> readUsages = handler.getReadUsages();
  assertEquals(usages.length, readUsages.size());

  List<String> textUsages = new ArrayList<String>();
  for (TextRange usage : readUsages) {
    String usageText = myFixture.getFile().getText().substring(usage.getStartOffset(), usage.getEndOffset());
    textUsages.add(usageText);
  }
  assertSameElements(Arrays.asList(usages), textUsages);
}
```

Here handler is `null` and the test fails. This is how the handler is created:
```
@Nullable
public static MyHandler createForElement(@NotNull Editor editor, PsiFile file, PsiElement element) {
  GoTypeOwner function = PsiTreeUtil.getParentOfType(element, GoFunctionLit.class, GoFunctionOrMethodDeclaration.class);
  if (function == null) return null;
  if (shouldCreateMyHandler(element)) {
    return new MyHandler(editor, file, element, function);
  }
  return null;
}
```

`getParentOfType()` returns null because all of element's parents are of type `GeneratedParserUtilBase$DummyBlock`.

I didn't investigate this any further, maybe I did something dumb and this is the reason why it doesn't work

zolotov:
@ilinum hm, works fine for me.

```java
public void testReturn() {
    myFixture.configureByText("foo.go", "package main; f<caret>unc name() string { return \"hello\"; }");
    @SuppressWarnings("unchecked")
    HighlightUsagesHandlerBase<PsiElement> handler = HighlightUsagesHandler.createCustomHandler(myFixture.getEditor(), myFixture.getFile());
    assertNotNull(handler);
    List<PsiElement> targets = handler.getTargets();
    assertEquals(1, targets.size());

    handler.computeUsages(targets);
    List<TextRange> readUsages = handler.getReadUsages();
    List<String> textUsages = new ArrayList<String>();
    for (TextRange usage : readUsages) {
      String usageText = myFixture.getFile().getText().substring(usage.getStartOffset(), usage.getEndOffset());
      textUsages.add(usageText);
    }
    System.out.println(textUsages);
  }
```

Output:
```
[func, return "hello"]
```

ilinum:
Okay, I was dumb, I forgot to include the package statement for the go snippet :)

ilinum:
Just pushed the tests

zolotov:
Thanks

zolotov:
@ilinum in case you want to make highlighting exit point even better: #2033

ilinum:
@zolotov thanks for the link!

PR #2032, opened 2015-11-22
Introduce inspection for placeholders (fixes #1392)

### LINKED ISSUE #1392: Provide support for Printf placeholders
Support for Printf placeholders should be added so that we can count how many placeholders are vs how many parameters are.

dlsniper:
This introduces the inspection to check the placeholders numbers vs number of arguments for functions like Printf or Sprintf

### DIFF
--- a/src/com/goide/inspections/GoPlaceholderCountInspection.java
+++ b/src/com/goide/inspections/GoPlaceholderCountInspection.java
@@ -1,0 +1,193 @@
+/*
+ * Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.*;
+import com.intellij.codeInspection.LocalInspectionToolSession;
+import com.intellij.codeInspection.ProblemHighlightType;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.openapi.util.Pair;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiReference;
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GoPlaceholderCountInspection extends GoInspectionBase {
+  private static final Pattern PLACEHOLDER_PATTERN =
+    Pattern.compile("(%#?((\\[\\d+\\]|\\*?)?\\.?)*(\\d*\\.\\d*)*\\s?(v|T|t|b|c|d|o|q|x|X|U|b|e|E|f|F|g|G|s|q|x|X|p))");
+
+  private static final Pattern INDEXED_PLACEHOLDER_PATTERN = Pattern.compile("(?:\\[(\\d+)\\])");
+
+  // This holds the name of the known formatting functions and position of the string to be formatted
+  private static final Map<String, Integer> FORMATTING_FUNCTIONS = ContainerUtil.newHashMap(
+    Pair.pair("printf", 0),
+    Pair.pair("sprintf", 0),
+    Pair.pair("errorf", 0),
+    Pair.pair("fprintf", 1),
+    Pair.pair("fscanf", 1),
+    Pair.pair("scanf", 0),
+    Pair.pair("sscanf", 1),
+    Pair.pair("fatalf", 0),
+    Pair.pair("panicf", 0));
+
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitCallExpr(@NotNull GoCallExpr o) {
+        PsiReference psiReference = o.getExpression().getReference();
+
+        if (psiReference == null) return;
+
+        PsiElement resolved = psiReference.resolve();
+        if (!(resolved instanceof GoFunctionOrMethodDeclaration)) return;
+        if (!isInCorrectPackage((GoFunctionOrMethodDeclaration)resolved)) return;
+
+        String name = getName((GoFunctionOrMethodDeclaration)resolved);
+        Integer placeholderPosition = FORMATTING_FUNCTIONS.get(name);
+        List<GoExpression> arguments = o.getArgumentList().getExpressionList();
+        if (arguments.size() <= placeholderPosition) return;
+
+        GoExpression placeholder = arguments.get(placeholderPosition);
+        if (!isStringPlaceholder(placeholder)) {
+          String message = "Value used for formatting text does not appear to be a string";
+          holder.registerProblem(placeholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
+          return;
+        }
+
+        int parameterCount = arguments.size() - placeholderPosition - 1;
+        int placeholdersCount = getPlaceholders(placeholder);
+        if (placeholdersCount == -1 || placeholdersCount == parameterCount) return;
+
+        String message = String.format("Got %d placeholder(s) for %d arguments(s)", placeholdersCount, parameterCount);
+        holder.registerProblem(placeholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
+      }
+    };
+  }
+
+  private static boolean isInCorrectPackage(@NotNull GoFunctionOrMethodDeclaration element) {
+    GoFile containingFile = element.getContainingFile();
+
+    return FORMATTING_FUNCTIONS.containsKey(getName(element)) &&
+           ("fmt".equals(containingFile.getImportPath()) ||
+            "log".equals(containingFile.getImportPath()));
+  }
+
+  @Nullable
+  private static String getName(@NotNull GoFunctionOrMethodDeclaration element) {
+    String name = element.getName();
+    if (name == null) return null;
+
+    return name.toLowerCase(Locale.getDefault());
+  }
+
+  private static int getPlaceholders(@Nullable GoExpression argument) {
+    if (argument == null) return -1;
+
+    String placeholderText = resolve(argument);
+    if (placeholderText == null) return -1;
+
+
+    Matcher placeholders = PLACEHOLDER_PATTERN.matcher(placeholderText);
+    Matcher countPlaceholders;
+
+    int indexCount = 0;
+    int maxIndexCount = 0;
+
+    while (placeholders.find()) {
+      indexCount++;
+
+      countPlaceholders = INDEXED_PLACEHOLDER_PATTERN.matcher(placeholders.group());
+      while (countPlaceholders.find()) {
+        int index = StringUtil.parseInt(countPlaceholders.group().replaceAll("\\[|\\]", ""), -1);
+        if (index > 0 && index != indexCount) {
+          indexCount = index;
+        }
+        if (maxIndexCount < indexCount) {
+          maxIndexCount = indexCount;
+        }
+      }
+      if (maxIndexCount < indexCount) {
+        maxIndexCount = indexCount;
+      }
+    }
+
+    return maxIndexCount;
+  }
+
+  private static boolean isStringPlaceholder(GoExpression argument) {
+    GoType goType = argument.getGoType(null);
+    if (goType == null || goType.getTypeReferenceExpression() == null) return false;
+
+    return goType.getTypeReferenceExpression().getText().equals("string");
+  }
+
+  @Nullable
+  private static String resolve(GoExpression argument) {
+    if (argument instanceof GoStringLiteral) return argument.getText();
+
+    if (argument.getReference() == null) return null;
+
+    PsiElement resolved = argument.getReference().resolve();
+    if (resolved instanceof GoVarDefinition) {
+      return getValue(((GoVarDefinition)resolved).getValue());
+    }
+    else if (resolved instanceof GoConstDefinition) {
+      return getValue(((GoConstDefinition)resolved).getValue());
+    }
+
+    return null;
+  }
+
+  @Nullable
+  private static String getValue(@Nullable GoExpression expression) {
+    if (expression instanceof GoStringLiteral) {
+      return expression.getText();
+    }
+    else if (expression instanceof GoAddExpr) {
+      String sum = getValue((GoAddExpr)expression);
+      return sum.isEmpty() ? null : sum;
+    }
+
+    return null;
+  }
+
+  @NotNull
+  private static String getValue(@Nullable GoAddExpr expression) {
+    if (expression == null) return "";
+    String result = "";
+    for (GoExpression expr : expression.getExpressionList()) {
+      if (expr instanceof GoStringLiteral) {
+        result += expr.getText();
+      }
+      else if (expr instanceof GoAddExpr) {
+        result += getValue(expr);
+      }
+    }
+
+    return result;
+  }
+}
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -232,6 +232,9 @@
     <localInspection language="go" displayName="Malformed struct tag" groupPath="Go"
                      groupName="Probable bugs" enabledByDefault="true" level="WARNING"
                      implementationClass="com.goide.inspections.GoStructTagInspection"/>
+    <localInspection language="go" displayName="Printf/Logf placeholder handler" groupPath="Go"
+                     groupName="Probable bug" enabledByDefault="true" level="WARNING"
+                     implementationClass="com.goide.inspections.GoPlaceholderCountInspection"/>
     <!-- /probable bugs -->
 
     <!-- control flow issues -->
--- a/resources/inspectionDescriptions/GoPlaceholderCount.html
+++ b/resources/inspectionDescriptions/GoPlaceholderCount.html
@@ -1,0 +1,36 @@
+<!--
+  ~ Copyright 2013-2015 Sergey Ignatov, Alexander Zolotov, Florin Patan
+  ~
+  ~ Licensed under the Apache License, Version 2.0 (the "License");
+  ~ you may not use this file except in compliance with the License.
+  ~ You may obtain a copy of the License at
+  ~
+  ~ http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing, software
+  ~ distributed under the License is distributed on an "AS IS" BASIS,
+  ~ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  ~ See the License for the specific language governing permissions and
+  ~ limitations under the License.
+  -->
+
+<html>
+<body>
+Checks the placeholders count in functions / methods.
+<!-- tooltip end -->
+<br/><br/>
+Known functions are (case-insensitive):<br/>
+<ul>
+  <li>fmt.Printf</li>
+  <li>fmt.Sprintf</li>
+  <li>fmt.Errorf</li>
+  <li>fmt.Fprintf</li>
+  <li>fmt.Fscanf</li>
+  <li>fmt.Scanf</li>
+  <li>fmt.Sscanf</li>
+  <li>log.Printf</li>
+  <li>log.Fatalf</li>
+  <li>log.Panicf</li>
+</ul>
+</body>
+</html>
--- a/testData/highlighting/placeholderCount.go
+++ b/testData/highlighting/placeholderCount.go
@@ -1,0 +1,87 @@
+package placeholderCount
+
+import "fmt"
+import "log"
+
+const (
+	myFormatConst      = "%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d"
+	myWrongFormatConst = "%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d"
+)
+
+func printf(placeholder string, args ...int) {
+
+}
+
+var s string = "%d"
+const s1 = s1
+const s2 = 1
+
+func _() {
+	fmt.Printf(<warning descr="Got 1 placeholder(s) for 2 arguments(s)">s</warning>, 1, 2)
+
+	fmt.Printf(<warning descr="Value used for formatting text does not appear to be a string">s1</warning>, 1, 2)
+
+	fmt.Printf(<warning descr="Value used for formatting text does not appear to be a string">s2</warning>, 1, 2)
+
+	fmt.Errorf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Fprintf(_, "%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Fscanf(_, "%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Printf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Scanf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Sprintf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	fmt.Sscanf(_, "%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	log.Fatalf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
+		1, 2, 3, 4, 5, 6, 7, 8, 9,
+	)
+	log.Panicf("%d %d %#[1]x %#x %2.f %d %2.2f %.f %.3f %[9]*.[2]*[3]f %d %f %#[1]x %#x %[2]d %v % d",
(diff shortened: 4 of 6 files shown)


dlsniper:
Currently numbered parameters are not correctly supported (as they will add to the count of existing params instead of reuse them).
Also, the `getText` from the call expression and from go expression should be replaced with something better (I think the parser needs to be improved for this).

dlsniper:
The inspection now handles the indexed parameters (even if it doesn't restrict them to fmt.(Printf, Sprintf, Fprintf) only).
The last step is to improve the usage of `getText()` on the GoExpression type and this should be ready for review.

dlsniper:
So, I've tried to make the GoExpression a NamedElement but I've got stuck, see the commit: https://github.com/dlsniper/google-go-lang-idea-plugin/commit/df2990d71a330aada12766f58a936ebef1e74a08
The plugin compiles, then it starts and it indexes files but I get errors like this:

```
[ 194914]  ERROR - aemon.impl.PassExecutorService - Not a stub type: UNARY_EXPR in class com.goide.psi.impl.GoUnaryExprImpl 
java.lang.AssertionError: Not a stub type: UNARY_EXPR in class com.goide.psi.impl.GoUnaryExprImpl
	at com.intellij.extapi.psi.StubBasedPsiElementBase.getElementType(StubBasedPsiElementBase.java:354)
	at com.intellij.psi.stubs.DefaultStubBuilder.buildStubTreeFor(DefaultStubBuilder.java:57)
	at com.intellij.psi.stubs.DefaultStubBuilder.buildStubTree(DefaultStubBuilder.java:36)
	at com.intellij.psi.stubs.StubTreeBuilder.buildStubTree(StubTreeBuilder.java:86)
	at com.intellij.psi.stubs.StubTreeLoaderImpl.readOrBuild(StubTreeLoaderImpl.java:61)
	at com.intellij.psi.impl.source.PsiFileImpl.getStubTree(PsiFileImpl.java:674)
	at com.intellij.psi.impl.source.PsiFileImpl.getStub(PsiFileImpl.java:654)
	at com.goide.psi.GoFile.getStub(GoFile.java:431)
	at com.goide.psi.GoFile.getImports(GoFile.java:173)
	at com.goide.psi.GoFile.hasCPathImport(GoFile.java:435)
	at com.goide.util.GoBuildMatcher.matchFile(GoBuildMatcher.java:56)
	at com.goide.util.GoBuildMatcher.matchFile(GoBuildMatcher.java:45)
	at com.goide.util.GoUtil.allowed(GoUtil.java:74)
	at com.goide.util.GoUtil$4.compute(GoUtil.java:233)
	at com.intellij.psi.impl.PsiCachedValueImpl.doCompute(PsiCachedValueImpl.java:49)
	at com.intellij.util.CachedValueBase.getValueWithLock(CachedValueBase.java:222)
	at com.intellij.psi.impl.PsiCachedValue.getValueWithLock(PsiCachedValue.java:66)
	at com.intellij.psi.impl.PsiCachedValueImpl.getValue(PsiCachedValueImpl.java:38)
	at com.intellij.util.CachedValuesManagerImpl.getCachedValue(CachedValuesManagerImpl.java:84)
	at com.goide.util.GoUtil.getAllPackagesInDirectory(GoUtil.java:227)
	at com.goide.psi.GoFile.getImportMap(GoFile.java:235)
	at com.goide.psi.impl.GoReference.processImports(GoReference.java:331)
	at com.goide.psi.impl.GoTypeReference.processUnqualifiedResolve(GoTypeReference.java:133)
	at com.goide.psi.impl.GoTypeReference.processResolveVariants(GoTypeReference.java:103)
	at com.goide.psi.impl.GoTypeReference.resolveInner(GoTypeReference.java:59)
	at com.goide.psi.impl.GoTypeReference.access$000(GoTypeReference.java:39)
	at com.goide.psi.impl.GoTypeReference$1.resolve(GoTypeReference.java:52)
	at com.goide.psi.impl.GoTypeReference$1.resolve(GoTypeReference.java:48)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolve(ResolveCache.java:144)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolveWithCaching(ResolveCache.java:167)
	at com.intellij.psi.impl.source.resolve.ResolveCache.resolveWithCaching(ResolveCache.java:159)
	at com.goide.psi.impl.GoTypeReference.multiResolve(GoTypeReference.java:85)
	at com.intellij.psi.PsiPolyVariantReferenceBase.resolve(PsiPolyVariantReferenceBase.java:46)
	at com.goide.GoAnnotator.annotate(GoAnnotator.java:54)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.runAnnotators(DefaultHighlightVisitor.java:162)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.visit(DefaultHighlightVisitor.java:103)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.runVisitors(GeneralHighlightingPass.java:354)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.access$100(GeneralHighlightingPass.java:65)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass$3.run(GeneralHighlightingPass.java:283)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.analyzeByVisitors(GeneralHighlightingPass.java:308)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.access$200(GeneralHighlightingPass.java:65)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass$4.run(GeneralHighlightingPass.java:314)
	at com.intellij.codeInsight.daemon.impl.DefaultHighlightVisitor.analyze(DefaultHighlightVisitor.java:87)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.analyzeByVisitors(GeneralHighlightingPass.java:311)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.collectHighlights(GeneralHighlightingPass.java:280)
	at com.intellij.codeInsight.daemon.impl.GeneralHighlightingPass.collectInformationWithProgress(GeneralHighlightingPass.java:219)
	at com.intellij.codeInsight.daemon.impl.ProgressableTextEditorHighlightingPass.doCollectInformation(ProgressableTextEditorHighlightingPass.java:82)
	at com.intellij.codeHighlighting.TextEditorHighlightingPass.collectInformation(TextEditorHighlightingPass.java:67)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass$1$1.run(PassExecutorService.java:444)
	at com.intellij.openapi.application.impl.ApplicationImpl.tryRunReadAction(ApplicationImpl.java:1169)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass$1.run(PassExecutorService.java:435)
	at com.intellij.openapi.progress.impl.CoreProgressManager.registerIndicatorAndRun(CoreProgressManager.java:446)
	at com.intellij.openapi.progress.impl.CoreProgressManager.executeProcessUnderProgress(CoreProgressManager.java:392)
	at com.intellij.openapi.progress.impl.ProgressManagerImpl.executeProcessUnderProgress(ProgressManagerImpl.java:54)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.doRun(PassExecutorService.java:432)
	at com.intellij.codeInsight.daemon.impl.PassExecutorService$ScheduledPass.run(PassExecutorService.java:408)
	at com.intellij.concurrency.JobLauncherImpl$VoidForkJoinTask$1.exec(JobLauncherImpl.java:206)
	at jsr166e.ForkJoinTask.doExec(ForkJoinTask.java:260)
	at jsr166e.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:858)
	at jsr166e.ForkJoinPool.scan(ForkJoinPool.java:1687)
	at jsr166e.ForkJoinPool.runWorker(ForkJoinPool.java:1642)
	at jsr166e.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:108)
[ 194915]  ERROR - aemon.impl.PassExecutorService - IntelliJ IDEA 15.0  Build #IC-143.381.42 
[ 194915]  ERROR - aemon.impl.PassExecutorService - JDK: 1.8.0_66 
[ 194915]  ERROR - aemon.impl.PassExecutorService - VM: Java HotSpot(TM) 64-Bit Server VM 
[ 194915]  ERROR - aemon.impl.PassExecutorService - Vendor: Oracle Corporation 
[ 194915]  ERROR - aemon.impl.PassExecutorService - OS: Linux 
```

Please advise how to proceed with this. Thank you.

zolotov:
> The plugin compiles, then it starts and it indexes files but I get errors like this:

All element types that corresponds to stub element should implement IStubElementType. Since you've made GoExpression stub GoUnaryExpr became stub too, obviously. Hence com.goide.GoTypes#UNARY_EXPR should be IStubElementType.

Anyway, making GoExpression namedStubElement is quite strange (e.g. what is the name for `1 + 2` expression?). In spite of incredible increasing index size, what's the point? What do you want to achieve?

dlsniper:
> In spite of incredible increasing index size, what's the point? What do you want to achieve?

In the first commit there are two points where I need to call `getText()` on the GoExpression:
- https://github.com/dlsniper/google-go-lang-idea-plugin/commit/a656d4921ae24737123490a2d9fd33785a8e6ebc#diff-069550141dffa62d28ef9e36887aea77R60
- https://github.com/dlsniper/google-go-lang-idea-plugin/commit/a656d4921ae24737123490a2d9fd33785a8e6ebc#diff-069550141dffa62d28ef9e36887aea77R99

I've realized that making a stub out of other elements which is not desirable.
What I want to achieve is a way to get the node value, for which function is called and for which arguments, so that I can perform the name matching for the function itself and then run the regex on the argument which holds the placeholders. Ideally I want to do that without triggering an index rebuild in background (iirc that's what `getText()` does, no?).

Or maybe there's a better way to do this and I don't realize it now?

dlsniper:
Should I remove the commit that changes the parser do that's easier to work with this PR? Thank you.

zolotov:
> Ideally I want to do that without triggering an index rebuild in background (iirc that's what getText() does, no?).

No, getText doesn't trigger reindex, nothing but changing file triggers reindex (or broken indecies which triggers full reindex for all projects). 
If you invoke `getText` on stub then stub will be converted to psi (will be reparsed). Since you already has non-stub element in inspection, you have the real PSI, so reparsing won't be invoked.

> Should I remove the commit that changes the parser do that's easier to work with this PR?

Yes, that changes look senseless.

dlsniper:
Ok. Thank you for the explanation. I've removed the previous commit. Ready for review I guess.

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+import org.jetbrains.annotations.Nullable;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GoPlaceholderCountInspection extends GoInspectionBase {
+  static final Pattern placeholderPattern =
```
zolotov: make private and rename according to Java code style: PLACEHOLDER_PATTERN

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+    Pattern.compile("(%((#)?(\\[\\d+\\]|\\*?)?\\.?)*((\\d*)(\\.\\d*))*\\s?(v|T|t|b|c|d|o|q|x|X|U|b|e|E|f|F|g|G|s|q|x|X|p))");
+
+  static final Pattern indexedPlaceholderPattern = Pattern.compile("(?:\\[(\\d+)\\])");
```
zolotov: private & rename

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+
+  // This holds the name of the known formatting functions and position of the string to be formatted
+  private Map<String, Integer> formattingFunctions = new HashMap<String, Integer>();
```
zolotov: final static.
You can fill in in initialization block 
```
new HashMap() {{
put(...)
}}
```
Or use ContainerUtil#newHashMap
```
  private static final Map<String, Integer> FORMATTING_FUNCTIONS = ContainerUtil.newHashMap(
    Pair.pair("printf", 0),
    Pair.pair("sprintf", 0),
    Pair.pair("errorf", 0),
    Pair.pair("fprintf", 1),
    Pair.pair("fscanf", 1),
    Pair.pair("scanf", 0),
    Pair.pair("sscanf", 1),
    Pair.pair("fatalf", 0),
    Pair.pair("panicf", 0));
```

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+    formattingFunctions.put("panicf", 0);
+  }
+
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitCallExpr(@NotNull GoCallExpr o) {
+        String name = o.getExpression().getText().toLowerCase();
```
zolotov: Please extract method `isFormattingFunction`

+

I think it's better to check real function declarations instead of call expressions.
To do that, resolve expression and then check whether it's a function with particular name from particular package. E.g.:

```java
func = callExpression.getExpression().getReference().resolve()
check func.getName
check func.getContainingFile.getPackage or func.getCointaningFile.getImportPath
```

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+          if (placeholdersCount < parameterCount) {
+            element = arguments.get(position);
+          }
+          holder.registerProblem(element, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
+        }
+      }
+    };
+  }
+
+  private static Integer getPlaceholders(@Nullable GoExpression argument) {
```
zolotov: set `int` return type

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+          return;
+        }
+
+        Integer position = formattingFunctions.get(name);
+        List<GoExpression> arguments = o.getArgumentList().getExpressionList();
+        if (arguments.size() <= position) {
+          return;
+        }
+
+        Integer placeholdersCount = getPlaceholders(arguments.get(position));
```
zolotov: int

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+        Integer parameterCount = arguments.size() - position - 1;
```
zolotov: int

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+
+        if (!placeholdersCount.equals(parameterCount)) {
```
zolotov: compare with `==`

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+          PsiElement element = o.getExpression();
+          String message = String.format("Got %d placeholder(s) for %d arguments(s)", placeholdersCount, parameterCount);
+          if (placeholdersCount < parameterCount) {
```
zolotov: retrieving reporting element is separated by constructing a message. Please keep similar things togother

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+      }
+    };
+  }
+
+  private static Integer getPlaceholders(@Nullable GoExpression argument) {
+    if (argument == null) {
+      return 0;
+    }
+
+    Matcher m = placeholderPattern.matcher(argument.getText());
```
zolotov: Give more descriptive names please

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+
+    Integer indexCount = 0;
+    Integer maxIndexCount = 0;
+
+    while (m.find()) {
+      indexCount++;
+
+      cm = indexedPlaceholderPattern.matcher(m.group());
+      while (cm.find()) {
+        Integer index = Integer.parseInt(cm.group().replaceAll("\\[|\\]", ""));
```
zolotov: Integer#parseInt throws an exception on invalid number, use StringUtil.parseInt instead

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,125 @@
...
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GoPlaceholderCountInspection extends GoInspectionBase {
+  static final Pattern placeholderPattern =
+    Pattern.compile("(%((#)?(\\[\\d+\\]|\\*?)?\\.?)*((\\d*)(\\.\\d*))*\\s?(v|T|t|b|c|d|o|q|x|X|U|b|e|E|f|F|g|G|s|q|x|X|p))");
```
zolotov: Too many groups. At least hash-sign group and whole regex group look redundant.

src/com/goide/inspections/GoPlaceholderCountInspection.java
dlsniper: I would say it's better only to resolve it only. There are some other packages out there, for example https://github.com/kr/pretty, that provide `fmt` or `log` like functionality so I'd rather try to keep them working if possible (not sure if that's the best choice but at least we can see how many people will report issues with other packages / see if we can expand the functionality on this).

dlsniper:
Done, thank you for feedback.

dlsniper:
Optional feature, is there any way to highlight the placeholder this matches when the cursor is over an argument? As that would be the final feature I see this for this for a while (until we can do type inference and match the type with the argument at least).

zolotov:
Sure, it could be implemented in the same way as highlighting exit points: GoHighlightExitPointsHandlerFactory

dlsniper:
I was just looking over that. Thank you (and thank you for your patience with the review :) )

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,142 @@
...
+
+    return resolved.getText().toLowerCase();
+  }
+
+  private static int getPlaceholders(@Nullable GoExpression argument) {
+    if (argument == null) {
+      return 0;
+    }
+
+    Matcher placeholders = PLACEHOLDER_PATTERN.matcher(argument.getText());
```
zolotov: What if there is not string argument? E.g. variable
```go
fmt.Printf(myStringWithPattern, value1, value2)
```

src/com/goide/inspections/GoPlaceholderCountInspection.java
zolotov: That's the point! I can use some library with following function:

```go
func Printf(s title, s message, a ...interface{}) 
```

And I'll get red code :(

src/com/goide/inspections/GoPlaceholderCountInspection.java
dlsniper: Here I bump yet again into the whole resolving to the final type issue that I have with other inspections as well (see https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1668 for example).

I'm not sure how to do it efficiently (not to mention that at some point I'll get stuck into how to do the resolving in case there's more than one step involved) and doing the whole resolving until the final type / value for each inspection makes me think that there has to be a better way to handle this (unless I don't understand how IntelliJ Platform / Java Runtime optimizations work to make this efficient).

(stupid question but) Would it be possible to have the GoType / Value attached to the nodes so that resolving / value retrieval doesn't happen every time?

zolotov: > I'm not sure how to do it efficiently (not to mention that at some point I'll get stuck into how to do the resolving in case there's more than one step involved) and doing the whole resolving until the final type / value for each inspection makes me think that there has to be a better way to handle this (unless I don't understand how IntelliJ Platform / Java Runtime optimizations work to make this efficient).

It's premature optimization. Just do it somehow – while { resolve } – nothing special

> Would it be possible to have the GoType / Value attached to the nodes so that resolving / value retrieval doesn't happen every time?

Resolving doesn't happen every time, it's caching

zolotov: And more important question. Why do you need this in this PR? What do you want to do?

dlsniper: > Why do you need this in this PR? What do you want to do?

Sorry, can you please rephrase, I didn't understood to what the question refers to. Thank you.

dlsniper:
@zolotov I've implemented the first part of the resolving function, please let me know if that's a good start for it as I'll continue to add the remaining use-cases to it if so. (quick link: https://github.com/dlsniper/google-go-lang-idea-plugin/commit/3e92e2a63e313b0d257ae47b9bcf91058bb5650c#diff-069550141dffa62d28ef9e36887aea77R170) Thank you.

zolotov:
@dlsniper when I ask what if there will be vairable I didn't mean that we should validate them :) I can suggest more scenarios that will fail the inspection: concatenation, string variable reassingment, changing string by pointer, etc. We just can't handle all of these even if we'll implement proper control-flow analyses. And I don't think we should. Also I don't think we should validate printf functions with variables, because we out of context, it's just overkill.

The thing that bothers be more is the algorithm of choosing functions to validate. I don't like the idea to validate all function with name printf. It leeds to many false positive reports. It's ok for weak warning inspections but not for error-inspections. And this is why I didn't merge the PR in the first place.

dlsniper:
> @dlsniper when I asked what if there will be vairable I didn't mean that we should validate them :) I can suggest more scenarios that will fail the inspection: concatenation, string variable reassingment, changing string by pointer, etc. We just can't handle all of these even if we'll implement proper control-flow analyses. And I don't think we should. Also I don't think we should validate printf functions with variables, because we out of context, it's just overkill.

Sure, so in that case should just enable the inspection for maybe the two cases: direct string or constant string (constant strings should be easier, unless they have concatenations. 

> The thing that bothers be more is the algorithm of choosing functions to validate. I don't like the idea to validate all functions with name printf. It leads to many false positive reports. It's ok for weak warning inspections but not for error-inspections. And this is why I didn't merge the PR in the first place.

I've understood that, and changed per feedback to inspect only the Go SDK packages for `fmt` and `log` which should be ok.

Should I tidy up the PR then?

zolotov:
> Sure, so in that case should just enable the inspection for maybe the two cases: direct string or constant string (constant strings should be easier, unless they have concatenations.

Exactly. Just make sure that inspection doesn't fail on string variables and concatenations

> I've understood that, and changed per feedback to inspect only the Go SDK packages for fmt and log which should be ok.

Sorry. I've missed that. Yes, looks good. 

> Should I tidy up the PR then?

Yes. But you may wait a little, I have a couple comments to write.

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,212 @@
...
+          }
+          holder.registerProblem(element, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
+        }
+      }
+    };
+  }
+
+  private static boolean isInCorrectPackage(@NotNull GoFunctionOrMethodDeclaration element) {
+    GoFile containingFile = element.getContainingFile();
+    GoPackageClause filePackage = containingFile.getPackage();
```
zolotov: There is GoFile#getPackageName. Btw, I'm not sure that it's request to check package name. Maybe import path is enough. What do you think?

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,212 @@
...
+    if (filePackage == null ||
+        filePackage.getIdentifier() == null || containingFile.getImportPath() == null) {
+      return false;
+    }
+
+    String packageName = filePackage.getIdentifier().getText();
+    String packageImportPath = containingFile.getImportPath();
+
+    return (packageName.equals("fmt") && packageImportPath.equals("fmt")) ||
```
zolotov: it's easier to check "fmt".equals(packageImportPath), then you don't have to check whether packageImportPath is null

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,212 @@
...
+           (packageName.equals("log") && packageImportPath.equals("log"));
+  }
+
+  private static boolean isFormattingFunction(@NotNull GoFunctionOrMethodDeclaration element) {
+    return FORMATTING_FUNCTIONS.containsKey(getName(element));
+  }
+
+  @Nullable
+  private static String getName(@NotNull GoFunctionOrMethodDeclaration element) {
+    PsiElement resolved = element.getIdentifier();
```
zolotov: ? element.getName()?

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        PsiElement resolved = psiReference.resolve();
+
+        if (!(resolved instanceof GoFunctionOrMethodDeclaration)) {
+          return;
+        }
+
+        if (!isFormattingFunction((GoFunctionOrMethodDeclaration)resolved)) return;
+        if (!isInCorrectPackage((GoFunctionOrMethodDeclaration)resolved)) return;
```
zolotov: I believe that isInCorrectPackage is a part of isFormattingFunction

dlsniper:
@zolotov I've updated the PR with support for resolving to string variables and consts. I'll add support for concatenation if you think it's needed but for now I think this would be enough. One problem that the current resolving has is that the following code

```
int constPosition = constSpec.getConstDefinitionList().indexOf(element);
return resolve(constSpec.getExpressionList().get(constPosition));
```

Might break if by some chance the indexOf returns a proper string but the position is in fact wrong. I'm not sure how to handle this for now but I'll investigate.

dlsniper:
I've fixed the wrong test. As for the comment above you can ignore it, I've better understood how the PSI tree looks like.

dlsniper:
@zolotov this should be ready to be merged now, unless there are some other code issues.

zolotov:
Didn't check it yet. Code looks good but I want to test some more cases in production.

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,209 @@
...
+      return null;
+    }
+
+    PsiReference resolved = argument.getReference();
+    if (resolved == null) {
+      return null;
+    }
+
+    PsiElement element = resolved.resolve();
+    while (element != null &&
```
zolotov: I don't understand what it stands for. Any test example? It doesn't covered at the moment: https://teamcity.jetbrains.com/viewLog.html?buildId=638588&buildTypeId=IntellijIdeaPlugins_Go_Test&tab=coverage_idea

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,209 @@
...
+      element = resolved.resolve();
+    }
+
+    if (element == null) {
+      return null;
+    }
+
+    if (element.getReference() != null) {
+      resolved = element.getReference();
+      if (resolved instanceof GoVarReference) {
```
zolotov: Doesn't work for 

```
var s string = "%d"
  	fmt.Printf(s, 1, 2)
```

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,209 @@
...
+        if (resolved.getElement().getParent() instanceof GoShortVarDeclaration) {
+          element = resolved.getElement();
+          GoShortVarDeclaration shortVarDeclaration = (GoShortVarDeclaration)resolved.getElement().getParent();
+          int varPosition = shortVarDeclaration.getVarDefinitionList().indexOf(element);
+          return resolve(shortVarDeclaration.getExpressionList().get(varPosition));
+        }
+        return "";
```
zolotov: Sometimes you return empty string and something – null. What's the point?

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -0,0 +1,209 @@
...
+      }
+    }
+    else if (element instanceof GoConstDefinition) {
+      if (element.getParent() instanceof GoConstSpec) {
+        GoConstSpec constSpec = (GoConstSpec)element.getParent();
+        int constPosition = constSpec.getConstDefinitionList().indexOf(element);
```
zolotov: Avoid using indecies without checks. Here is the code, it produces an exception:
```
const s
  	fmt.Printf(s, 1, 2)
```

zolotov: And stack overflow
```go
const s = s
fmt.Printf(s, 1, 2)
```

dlsniper:
I'll work on this after #1668 gets merged.

dlsniper:
I've reworked this and hopefully it's a bit more clean now.
I'll use it on the Go SDK and on my GOPATH to see what happens and if I get any issues with it. I've also  fixed all the issues pointed out earlier and aligned (hopefully) the formatting better with the one used in the rest of the code.

PR #2035, opened 2015-11-23
Added const folding option (fixes #2015)

### LINKED ISSUE #2015: const sections cannot be collapsed
Unlike `var` sections, `const` sections cannot be collapsed. This seems like inconstent behavior.

For example this `var` section can be collapsed:

```go
var (
    something = "1"
    // other variables...
)
```

But this `const` section cannot be collapsed:

```go
const (
    something = "1"
    // other constants...
)
```

dlsniper:


mdwhatcott:
Thanks @dlsniper and @zolotov !

PR #2042, opened 2015-11-26
Exit point highlighting - highlight break statements

ilinum:


src/com/goide/highlighting/exitpoint/BreakStmtExitPointHandler.java
```diff
@@ -0,0 +1,133 @@
...
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.intellij.util.Consumer;
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+class BreakStmtExitPointHandler extends HighlightUsagesHandlerBase<PsiElement> {
```
zolotov: Go-prefix, public

src/com/goide/highlighting/exitpoint/FunctionExitPointHandler.java
```diff
@@ -0,0 +1,83 @@
...
+import com.intellij.psi.PsiFile;
+import com.intellij.psi.util.PsiTreeUtil;
+import com.intellij.util.Consumer;
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+
+class FunctionExitPointHandler extends HighlightUsagesHandlerBase<PsiElement> {
```
zolotov: Go-prefix, public

ilinum: I'll add the prefix

Why public? This class used to be internal, so I'm assuming we don't want anyone to know it even exists.

zolotov: package-visibility it's too confusing, every time it ends with CNFE eventually in runtime

+ fill nullability annotations, there is an action for that

src/com/goide/highlighting/exitpoint/BreakStmtExitPointHandler.java
```diff
@@ -0,0 +1,133 @@
...
+  private final PsiElement myTarget;
+  @Nullable private final GoBreakStatement myBreakStmt;
+  @Nullable private PsiElement myOwner;
+
+  private BreakStmtExitPointHandler(Editor editor,
```
zolotov: From IDEA code style:
`Do not shorten names, e.g. Cnd/Cond for Condition, Bpt for Breakpoint`

src/com/goide/highlighting/exitpoint/FunctionExitPointHandler.java
ilinum: Okay, fair enough

src/com/goide/highlighting/exitpoint/FunctionExitPointHandler.java
```diff
@@ -0,0 +1,83 @@
...
+  private FunctionExitPointHandler(Editor editor, PsiFile file, PsiElement target, GoTypeOwner function) {
+    super(editor, file);
+    myTarget = target;
+    myFunction = function;
+  }
+
+  @NotNull
+  @Override
+  public List<PsiElement> getTargets() {
+    return ContainerUtil.newSmartList(myTarget);
```
zolotov: singleton list is more appropriate

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -1131,4 +1131,19 @@ public static boolean isUnaryBitAndExpression(@Nullable PsiElement parent) {
...
+  @Nullable
+  public static PsiElement getBreakStatementOwner(@NotNull GoBreakStatement breakStmt) {
+    GoLabelRef label = breakStmt.getLabelRef();
+    if (label != null) {
+      return label.getReference().resolve();
+    }
+    GoCompositeElement breaksOutOf =
+      PsiTreeUtil.getParentOfType(breakStmt, GoSwitchStatement.class, GoForStatement.class, GoSelectStatement.class,
+                                  GoFunctionOrMethodDeclaration.class);
+    if (breaksOutOf instanceof GoFunctionOrMethodDeclaration) {
```
zolotov: is GoFunctionOrMethodDeclaration kind of optimization?

src/com/goide/highlighting/exitpoint/FunctionExitPointHandler.java
ilinum: well, SmartList seems valid as well:

From com.intellij.util.SmartList:
```
/**
 * A List which is optimised for the sizes of 0 and 1,
 * in which cases it would not allocate array at all.
 */
```

src/com/goide/psi/impl/GoPsiImplUtil.java
ilinum: Well, yes. We need to stop searching when we hit GoFunctionOrMethodDeclaration, no point looking higher up the tree. 

According to spec https://golang.org/ref/spec#Break_statements A "break" statement terminates execution of the innermost "for", "switch", or "select" statement within the same function.

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -1131,4 +1131,19 @@ public static boolean isUnaryBitAndExpression(@Nullable PsiElement parent) {
     PsiElement grandParent = parent != null ? parent.getParent() : null;
     return grandParent instanceof GoUnaryExpr && ((GoUnaryExpr)grandParent).getBitAnd() != null;
   }
+
+  @Nullable
+  public static PsiElement getBreakStatementOwner(@NotNull GoBreakStatement breakStmt) {
```
zolotov: I guess this method will be needed only for break statement highlighting for a while, so it makes sense to move it to handler

src/com/goide/highlighting/exitpoint/BreakStmtExitPointHandler.java
ilinum: Will fix that

src/com/goide/psi/impl/GoPsiImplUtil.java
zolotov: Hm, I believe they meant GoFunctionLit, there is no nested GoFunctionDeclaraions in golang.

ilinum: Would a top level function be GoFunctionLit and GoFunctionOrMethodDeclaration?
Because we need to stop at either of those

zolotov: I think we don't need to stop on GoFunctionOrMethodDeclaration. Only in optimisation purposes, but two getParent invocations doesn't matter much.

ilinum: I guess you're right. It's not worth it.

What do you think about GoFunctionLit? Anonymous functions may be pretty far down the tree. 
Do you think it's worth doing it for GoFunctionLit?

zolotov: Sure, we have to. It's about correctness. I should recheck this, but I think that in following code nothing should be highlighted.

```go
func _() {
  for () {
    func() { bre<caret>ak; }
  }()
}
```

ilinum: You're right, we do have to stop. The code you provided doesn't compile

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
```diff
@@ -0,0 +1,149 @@
...
+      PsiElement parent = breakStmtOwner.getParent();
+      if (parent instanceof GoCompositeElement) {
+        visitor.visitCompositeElement(((GoCompositeElement)parent));
+      }
+    }
+  }
+
+
+  @Nullable
+  private static PsiElement getBreakStatementOwner(@NotNull GoBreakStatement breakStmt) {
```
zolotov: I've just think of where I can reuse this function :) Will move later then

zolotov:
thanks

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
ilinum: Was it the error highlighting: `break is not in a loop`?

Because it's not highlighted now and it probably should be.

ilinum: Also, I am curious, why do you use inspections to do error highlighting? Why not just do it in GoAnnotator? Is there some benefit in doing that?

zolotov: In order to have an ability to disable them or suppress

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        super.visitBreakStatement(o);
+      }
+
+      @Override
+      public void visitSwitchStatement(@NotNull GoSwitchStatement o) {
+        if (o == breakStmtOwner) {
+          GoSwitchStart switchStart = o.getSwitchStart();
+          if (switchStart != null) {
+            addOccurrence(switchStart.getSwitch());
```
zolotov: We shouldn't go deeper if it's our owner

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
```diff
@@ -0,0 +1,149 @@
...
+          }
+        }
+        super.visitSwitchStatement(o);
+      }
+
+      @Override
+      public void visitForStatement(@NotNull GoForStatement o) {
+        if (o == breakStmtOwner) {
+          addOccurrence(o.getFor());
```
zolotov: We shouldn't go deeper if it's our owner

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
```diff
@@ -0,0 +1,149 @@
...
+        }
+        super.visitForStatement(o);
+      }
+
+      @Override
+      public void visitSelectStatement(@NotNull GoSelectStatement o) {
+        if (o == breakStmtOwner) {
+          addOccurrence(o.getSelect());
```
zolotov: We shouldn't go deeper if it's our owner

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
```diff
@@ -0,0 +1,149 @@
...
+    }
+    return breaksOutOf;
+  }
+
+  @Nullable
+  public static GoBreakStatementExitPointHandler createForElement(@NotNull Editor editor, @NotNull PsiFile file, @NotNull PsiElement element) {
+    PsiElement target =
+      PsiTreeUtil.getParentOfType(element, GoBreakStatement.class, GoSwitchStatement.class, GoSelectStatement.class, GoForStatement.class);
+    if (target == null) return null;
+    if (target instanceof GoBreakStatement) {
```
zolotov: I don't understand why do you need GoBreakStatement. Then you use it only for retrieving breakStatementOwner

zolotov: Oh, I see, you need it for proper highlighting labeled breakStatement. Anyway, looks dirty, I don't see any sense not to pass owner to the handler in the first place.

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
ilinum: oh, I see. I was hoping this would give a performance benefit and we could use it in Scala plugin. Oh well :(

zolotov: Nope, actually annotator is faster, it triggers earlier. Actually I think we should reimplement couple of our inspection with annotator. Some generic inspection like misplaced continue

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
ilinum: Yeah, I could not pass it, but it would be slightly worse performance-wise (very slightly, one extra call to `getBreakStatementOwner`)

resources/META-INF/plugin.xml
```diff
@@ -121,7 +121,7 @@
     <typeDeclarationProvider implementation="com.goide.editor.GoTypeDeclarationProvider"/>
     <codeInsight.parameterInfo language="go" implementationClass="com.goide.editor.GoParameterInfoHandler"/>
     <lang.smartEnterProcessor language="go" implementationClass="com.goide.editor.smart.GoSmartEnterProcessor"/>
-    <highlightUsagesHandlerFactory implementation="com.goide.highlighting.GoHighlightExitPointsHandlerFactory"/>
+    <highlightUsagesHandlerFactory implementation="com.goide.highlighting.exitpoint.GoHighlightExitPointsHandlerFactory"/>
```
zolotov: Oops, missed that. I believe it's more about `usages` than `exitpoints`. Maybe we'll highlight something more in future

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
ilinum: You could also add an option to turn highlighting off.
In Scala plugin it's called type-aware highlighting

ilinum: Well, I understand that, sometimes inspections take a while to appear. But I was thinking whether doing a lot of work in Annotator slows down typing, for instance

resources/META-INF/plugin.xml
ilinum: It is called `ExitPointsHandlerFactory` though :)

ilinum: Oh, aren't usages highlighted already?

zolotov: Directly resolved usages – yes. But there are another possible cases. Look at java/groovy implementations: 
screenshot 2015-11-27 02 12 13

And base class is called HighlightingUsagesHandlerBase :)

src/com/goide/highlighting/exitpoint/GoBreakStatementExitPointHandler.java
ilinum: You mean we shouldn't go deeper if it's NOT our owner?

ilinum: And even if it is not our owner, we should go deeper, in case our owner is a `GoLabelDefinition` statement

zolotov: It's questionable. Should we even highlight all of these?

screenshot 2015-11-27 02 28 39

Of even that

screenshot 2015-11-27 02 28 29

zolotov: If so, that we actually don't care where break is places, we can just search all reference to label

ilinum: Okay, so here's my proposal for how break statements with labels should be highlighted:

1) Only break statements within the same function should be highlighted (which should fix the second screenshot you provided)
2) When caret is on the break statement with a label, we should highlight everything that it breaks out of: any `for`, `select`, and `switch` statement, as well as the label. However, my concern is that it may be confusing if the statements are far apart.
3) When caret is on `for`, `select`, or `switch`, we should not highlight any labelled break statements, because it is complicated to see whether the label they refer to is outside or inside the owner.

@zolotov what do you think?

ilinum: Actually, 1) seems to be fixed, because it's underlined red after you added the inspection and is not highlighted as a usage.

PR #2050, opened 2015-11-28
Introduce comment start inspection (fixes #1386)

### LINKED ISSUE #1386: Inspection to check if the the function/method/package documentation starts with the correct name
godoc relies on this and we should get the same rules as it has (to check golint).
This applies to packages, structs, interfaces, methods, functions. Special cases exists for each, covered in the golint as well.
Usually applies to exported elements only, optional toggle for private elements as well?

```go
// Demo does things  -> correct
func Demo() {
}

// Dem does other things -> incorrect
func Demo2() {
}
```

We could also have a quick fix for the above that generates the missing doc headers.

dlsniper:


src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -0,0 +1,109 @@
...
+      @Override
+      public void visitComment(PsiComment comment) {
+        PsiElement commentParent = comment.getParent();
+        if ((!(commentParent instanceof GoFile)) &&
+            (!(commentParent instanceof GoTopLevelDeclaration))) {
+          return;
+        }
+
+        String elementName = null;
+        PsiElement nextElement = PsiTreeUtil.skipSiblingsForward(comment, PsiWhiteSpace.class);
```
zolotov: And what if?

```go
// Demo does things  -> correct
//
// Deprecated: use other thing
func Demo() {
	Demo2()
}

File: src/com/goide/inspections/GoCommentStartInspection.java [hunk 2]
```diff
@@ -0,0 +1,109 @@
...
+        }
+        else if (nextElement instanceof GoVarDeclaration) {
+          if (((GoVarDeclaration)nextElement).getVarSpecList().size() == 0) {
+            return;
+          }
+          elementName = ((GoVarDeclaration)nextElement).getVarSpecList().get(0).getVarDefinitionList().get(0).getName();
+        }
+
+        if (elementName == null ||
+            (!(nextElement instanceof GoPackageClause)) && !StringUtil.isCapitalized(elementName)) {
```
zolotov: isPublic method for named elements

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -0,0 +1,109 @@
...
+        // +4 stands for //<space>Element_Name<space>
+        else if (commentText.length() <= elementName.length() + 4) {
+          holder.registerProblem(comment, "Comment should be meaningful or it should be removed", ProblemHighlightType.WEAK_WARNING);
+        }
+      }
+    };
+  }
+
+  private static boolean correctCommentStart(String commentText, String elementName) {
+    return commentText.startsWith("// " + elementName) ||
```
zolotov: And what if there are two spaces between slash and name?

dlsniper:
@zolotov what do you think? Merge this and have the quick fix later? I'm not really sure how to do the text processing for the fix right now and I'd rather have the functionality for the inspection merged first so that we can mirror the other tools better.

zolotov:
@dlsniper here are couple questions yet, not about quickfix. But since you asking, I don't think that this inspection need any quickfix.

src/com/goide/inspections/GoCommentStartInspection.java
dlsniper: Anything that's not starting with the correct space / prefix and so on is considered to be incorrect by golint.

zolotov: Does they have a special inspection for that?

dlsniper: I've also linked it in in the class comment, forgot to do it: https://github.com/golang/lint/blob/32a87160691b3c96046c0c678fe57c5bef761456/lint.go#L744

src/com/goide/inspections/GoCommentStartInspection.java
dlsniper: No all elements here are of type `GoNamedElement` that's why I've done this.

zolotov: Which one is not?

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -0,0 +1,152 @@
...
+
+        if (elementName == null ||
+            (!(nextElement instanceof GoPackageClause)) && !StringUtil.isCapitalized(elementName)) {
+          return;
+        }
+
+        String commentText = comment.getText();
+        if (nextElement instanceof GoPackageClause) {
+          if (!elementName.equals("main") && !commentText.startsWith("// Package " + elementName)) {
+            String description = "Package comment should be of the form \"Package " + elementName + " ...\"\n";
```
zolotov: Why do you need \n?

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+        else if (!correctCommentStart(commentText, elementName)) {
+          holder.registerProblem(comment, "Comment should start with '" + elementName + "'", ProblemHighlightType.WEAK_WARNING);
+        }
+        // +4 stands for //<space>Element_Name<space>
+        else if (commentText.length() <= elementName.length() + 4) {
+          holder.registerProblem(comment, "Comment should be meaningful or it should be removed", ProblemHighlightType.WEAK_WARNING);
+        }
+
+        if (hadMoreThanOneSpace) {
```
zolotov: It's not about comment start

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        if (elementName == null ||
+            (!(nextElement instanceof GoPackageClause)) && !StringUtil.isCapitalized(elementName)) {
+          return;
+        }
+
+        String commentText = comment.getText();
+        if (nextElement instanceof GoPackageClause) {
+          if (!elementName.equals("main") && !commentText.startsWith("// Package " + elementName)) {
```
zolotov: main package isn't covered with test

zolotov:
merged manually

PR #2051, opened 2015-11-28
Exported element should have its own declaration (fixes #2049)

### LINKED ISSUE #2049: exported var/const XYZ should have its own declaration
```go
package demo

const Hello1, hello2 = 1, 2 // This is valid

const Hello3, Hello4 = 3, 4 // exported const Hello4 should have its own declaration

var Hello5, hello6 int // This is valid

var Hello7, Hello8 int // exported var Hello8 should have its own declaration
```

dlsniper:


src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,75 @@
...
+      }
+    };
+  }
+
+  public static void checkElement(ProblemsHolder holder, String type, GoNamedElement namedElement, int index) {
+    if (index < 1) {
+      return;
+    }
+
+    if (!StringUtil.isCapitalized(namedElement.getName())) {
```
zolotov: namedElement.isPublic

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,75 @@
...
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitConstDeclaration(@NotNull GoConstDeclaration o) {
+        if (!(o.getParent() instanceof GoFile)) {
+          return;
+        }
+
+        for (int i = 0; i < o.getConstSpecList().size(); i++) {
+          int index = 0;
```
zolotov: So this is a warning?

```go
const (
	Foo = 1
	Bar = 1
)
```

zolotov: I believe we should use testdata form golint for testing all inspections that we grab from there.

https://github.com/golang/lint/tree/master/testdata

dlsniper:
@zolotov where should I add a link to the golint inspection? Would the class comment be ok?

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
dlsniper: > So this is a warning?

```go
// valid code

const (
	Foo = 1
	Bar = 1
)

// warning code

const (
	Foo = 1
	Bar, Baz = 1 // Baz should be declared in it's own block
)
```

dlsniper: > I believe we should use testdata form golint for testing all inspections that we grab from there.

Unfortunately there's no test data for that particular check, I've searched for the the message in the testdata folder and nothing.

zolotov: I know, that's why I suggest to use all testdata

zolotov: So where is the test for 

```go
// valid code

const (
    Foo = 1
    Bar = 1
)

// warning code

const (
    Foo = 1
    Bar, Baz = 1 // Baz should be declared in it's own block
)
```

I didn't check, but I think your implementation fails on this code

dlsniper: There's already a test for this: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2051/files#diff-35e2e75d1a64cfc694a3df6f5bbe69daR13 It's not on `const` but on `var` which has the same implementation. I can add a second case for a `const` as well if you think it makes sense,

zolotov:
@dlsniper yep, class commit is ok.

Btw, what about quickfix?

dlsniper:
> Btw, what about quickfix?

Is there any way to hook-up with the extract functionality of the IDE? Since there are already two extract types, var and const, I think it should be handled by that functionality, if possible.

Otherwise I guess I can write the quick fix to extract the elements into a separate element definition.

zolotov:
> Is there any way to hook-up with the extract functionality of the IDE? Since there are already two extract types, var and const, I think it should be handled by that functionality, if possible.

I don't think that extract refactorings may give you a lot to reuse. The quickfix for this inspection is much simpler than refactoring and it's not really similar to introducing something.

> Otherwise I guess I can write the quick fix to extract the elements into a separate element definition.

Would be great. If you want I can merge this PR first and keep the PR open until quickfix is ready.

dlsniper:
I'd rather have this in one Go as it would be a nicer user experience to also get the quick fix when the IDE will report the issue.

dlsniper:
I've added the quick fix for this. Is there any way to trigger the code formatting after the quick fix runs?

zolotov:
As I can see on `variation4.go` test, extracted code is reformated after quickfix

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+            GoType elemType = ((GoVarSpec)element.getParent()).getType();
+            if (elemType != null) {
+              elementType = elemType.getText();
+            }
+          }
+          else {
+            return;
+          }
+
+          if (elementValueList.size() != 0 &&
```
zolotov: isempty

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+              elementValueList.size() <= index) {
+            return;
+          }
+
+          PsiElement elementParent = element.getParent();
+          PsiElement newElement;
+          PsiElement afterElement;
+          if (elementValueList.size() > 0) {
```
zolotov: isempty

tests/com/goide/quickfix/GoExportedOwnDeclarationQuickFixTest.java
```diff
@@ -0,0 +1,43 @@
...
+    myFixture.enableInspections(GoExportedOwnDeclarationInspection.class);
+  }
+
+  @NotNull
+  @Override
+  protected String getBasePath() {
+    return "quickfixes/exported-own-declaration";
+  }
+
+  public void testVariation1() { doTest(GoExportedOwnDeclarationInspection.QUICK_FIX_NAME); }
```
zolotov: vartiantionN is not a very descriptive name. Test name should describe what exactly you're testing.

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+      PsiElement elementParent = element.getParent();
+
+      if (element instanceof GoConstDefinition) {
+        if (((GoConstSpec)elementParent).getExpressionList().size() > 0) {
+          return ((GoConstSpec)elementParent).getExpressionList()
+            .get(((GoConstSpec)elementParent).getExpressionList().size() - 1);
+        }
+
+        return ((GoConstSpec)elementParent).getType() == null
+               ? ((GoConstSpec)elementParent).getConstDefinitionList()
```
zolotov: ContainerUtil#getLastItem

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+
+    public MyLocalQuickFixBase(int index) {
+      super(QUICK_FIX_NAME);
+      this.index = index;
+    }
+
+    @Override
+    public void applyFix(@NotNull final Project project, @NotNull ProblemDescriptor descriptor) {
+      final PsiElement element = descriptor.getPsiElement();
+      if (element == null) return;
```
zolotov: check isValid

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+            if (elemType != null) {
+              elementType = elemType.getText();
+            }
+          }
+          else {
+            return;
+          }
+
+          if (elementValueList.size() != 0 &&
+              elementValueList.size() <= index) {
```
zolotov: Actually, at this moment index might be different already, you have to recalculate it.

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+            return;
+          }
+
+          PsiElement elementParent = element.getParent();
```
zolotov: What is parent? Statement?

src/com/goide/inspections/GoExportedOwnDeclarationInspection.java
```diff
@@ -0,0 +1,207 @@
...
+              newElement = GoElementFactory.createVarSpec(project, element.getText(), elementType, "");
+            }
+          }
+
+          afterElement = getAfterElement(element);
+          PsiElement grandParent = elementParent.getParent();
+          elementParent.addAfter(newElement, afterElement);
+          elementParent.addAfter(GoElementFactory.createNewLine(project), afterElement);
+
+          if (elementParent instanceof GoConstSpec) {
```
zolotov: So you didn't know what elementParent until now, but you've already added some elements there, this is looks very brittle

zolotov:
Sorry, but quickfix code is unreadable for me. There are many type assumptions and assumptions about PSI, playing with list indices and quite few checks. It looks brittle, I'm afraid to merge it, sorry.

Please make test names more sensible and I'll try to cleanup the quick fix later.

zolotov:
Thanks! Will try to do something with it. If I fail then just I merge your implementation

dlsniper:
The code would be much more easier to read / write if the methods on the (Const|Var)(Spec|Definition|Declaration) wouldn't have different names names or extend/inherit the same type. It would reduce the dance with `if const else if var`.

As for the code itself, I'll try to find a nicer way to improve it as well. I guess one way to do it would be simply use `GoElementFactory.create*Declaration` and preprocess the elements added to the specList.

zolotov:
> The code would be much more easier to read / write if the methods on the (Const|Var)(Spec|Definition|Declaration) wouldn't have different names names or extend/inherit the same type. It would reduce the dance with if const else if var.

Agree. That's exactly I want to fix.

Issue #1588, opened 2015-04-29
Not an expression as function parameter should be highlighted


zolotov:
```go
package main

func main() {
	_ := []string
	a := []string{}
	println(a)
	
	println([]string) // should be error
	println([]string{})
}
```

sjamesr:
@zolotov in your example ` _ := []string` should be an error too (invalid syntax). For example:

```go
package main

func main() {
	[]string
}

prog.go:4: syntax error: unexpected semicolon or newline, expecting {
```

Right now the Go plugin parser accepts this syntax and yields the following PSI tree:

```
GO_FILE(0,43)
  PACKAGE_CLAUSE(0,12)
    PsiElement(package)('package')(0,7)
    PsiWhiteSpace(' ')(7,8)
    PsiElement(identifier)('main')(8,12)
  IMPORT_LIST(12,12)
    <empty list>
  PsiWhiteSpace('\n\n')(12,14)
  FUNCTION_DECLARATION(14,42)
    PsiElement(func)('func')(14,18)
    PsiWhiteSpace(' ')(18,19)
    PsiElement(identifier)('main')(19,23)
    SIGNATURE(23,25)
      PARAMETERS(23,25)
        PsiElement(()('(')(23,24)
        PsiElement())(')')(24,25)
    PsiWhiteSpace(' ')(25,26)
    BLOCK(26,42)
      PsiElement({)('{')(26,27)
      PsiWhiteSpace('\n')(27,28)
      PsiWhiteSpace('    ')(28,32)
      SIMPLE_STATEMENT(32,40)
        LEFT_HAND_EXPR_LIST(32,40)
          LITERAL_TYPE_EXPR(32,40)
            ARRAY_OR_SLICE_TYPE(32,40)
              PsiElement([)('[')(32,33)
              PsiElement(])(']')(33,34)
              TYPE(34,40)
                TYPE_REFERENCE_EXPRESSION(34,40)
                  PsiElement(identifier)('string')(34,40)
      PsiWhiteSpace('\n')(40,41)
      PsiElement(})('}')(41,42)
  PsiWhiteSpace('\n')(42,43)
```

I'm trying to compare go.bnf to the go.y grammar from the go source but perhaps you have an idea what's wrong.

zolotov:
@sjamesr you're right, thanks.

> I'm trying to compare go.bnf to the go.y grammar from the go source but perhaps you have an idea what's wrong.

[...]

PR #2062, opened 2015-12-04
Introduce ExpressionOrType production

sjamesr:
This change eliminates literal type expressions from the Expression production
and adds them to a new ExpressionOrType production. This means that the parser
now correctly reports syntax errors like the following:

```go
package main

func main() {
    []string
}
```

### DIFF
--- a/gen/com/goide/parser/GoParser.java
+++ b/gen/com/goide/parser/GoParser.java
@@ -377,7 +377,7 @@ public class GoParser implements PsiParser, LightPsiParser {
   }
 
   /* ********************************************************** */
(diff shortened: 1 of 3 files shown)


sjamesr:
@zolotov here's an attempt at addressing the issues in https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1588. The tests pass, but I'm not sure if it's 100% correct. Also the error messaging isn't ideal:

screen shot 2015-12-04 at 9 30 24 am

screen shot 2015-12-04 at 9 29 36 am

It would be nice if the whole type expression was highlighted rather than the whitespace in the previous line.

Issue #2053, opened 2015-11-29
Enhance support for go get inside the plugin


dlsniper:
In order for the plugin to properly download all the dependencies of a package, it should execute the `go get` update with something like this `go get <package-name>/...` which will traverse the whole package after downloading and get the dependencies.

Also, it could be useful to add the `-t` switch to `go get` so that the test dependencies are also downloaded.

`go get` has a `-v` switch which allows it to display what happens when it's downloading the package. For example:

```
go get -v -u github.com/koding/kite/...
github.com/koding/kite (download)
github.com/cenkalti/backoff (download)
github.com/dgrijalva/jwt-go (download)
github.com/gorilla/mux (download)
github.com/gorilla/context (download)
github.com/gorilla/websocket (download)
# cd /home/florin/golang/src/github.com/gorilla/websocket; git pull --ff-only
error: Your local changes to the following files would be overwritten by merge:
        examples/chat/conn.go
Please, commit your changes or stash them before you can merge.
Aborting
Updating ab5b3a6..361d4c0
package github.com/gorilla/websocket: exit status 1
github.com/juju/ratelimit (download)
github.com/koding/cache (download)
github.com/mitchellh/mapstructure (download)
github.com/koding/logging (download)
github.com/satori/go.uuid (download)
Fetching https://golang.org/x/crypto/ssh/terminal?go-get=1
Parsing meta tags from https://golang.org/x/crypto/ssh/terminal?go-get=1 (status code 200)
get "golang.org/x/crypto/ssh/terminal": found meta tag main.metaImport{Prefix:"golang.org/x/crypto", VCS:"git", RepoRoot:"https://go.googlesource.com/crypto"} at https://golang.org/x/crypto/ssh/terminal?go-get=1
get "golang.org/x/crypto/ssh/terminal": verifying non-authoritative meta tag
Fetching https://golang.org/x/crypto?go-get=1
Parsing meta tags from https://golang.org/x/crypto?go-get=1 (status code 200)
golang.org/x/crypto (download)
Fetching https://gopkg.in/igm/sockjs-go.v2/sockjs?go-get=1
Parsing meta tags from https://gopkg.in/igm/sockjs-go.v2/sockjs?go-get=1 (status code 200)
[...]

PR #2065, opened 2015-12-05
Run go get for test packages and sub-packages

dlsniper:
As a start for fixing https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2053, this allows for running `go get` for test packages and sub-packages as well.


zolotov:
Where I can read about `importPath/...` option?

dlsniper:
This is the docs page: https://golang.org/cmd/go/#hdr-Description_of_package_lists 

>An import path is a pattern if it includes one or more "..." wildcards, each of which can match any string, including the empty string and strings containing slashes. Such a pattern expands to all package directories found in the GOPATH trees with names matching the patterns. As a special case, x/... matches x as well as x's subdirectories. For example, net/... expands to net and packages in its subdirectories.
>
> An import path can also name a package to be downloaded from a remote repository. Run 'go help importpath' for details.

Also, this is how `go get` tool behaves when the `/...` is present:

snapshot157

PR #2066, opened 2015-12-05
Add an annotator check for type ref used as expr

sjamesr:
Checks the following scenario:

```go
func foo() {
    string     // error: Type string is not an expression.
}
```

Partially addresses https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1588.

### DIFF
--- a/src/com/goide/highlighting/GoAnnotator.java
+++ b/src/com/goide/highlighting/GoAnnotator.java
@@ -26,6 +26,7 @@ import com.intellij.psi.PsiDirectory;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.util.PsiTreeUtil;
 import org.jetbrains.annotations.NotNull;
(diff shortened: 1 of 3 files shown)


sjamesr:
This implementation is not general enough, working on a better one..

sjamesr:
I think it's OK for review now. I tried to implement the same checks as https://github.com/golang/go/blob/master/src/cmd/compile/internal/gc/typecheck.go#L306. I don't understand Go compiler internals, so it might be that the check isn't identical. The tests pass though! :)

ignatov:
Please rebase it.

sjamesr:
Done, please take a look.

ignatov:
Thanks, please see my cleanup commit: 1f79f53bfd056180b054753b1710b3900b0d5871. BTW, we decided not to use dots for annotation messages: 1501e0d3468d8fb892457382b8259e8ef5d50008.

Issue #1420, opened 2015-03-08
Usage of package XYZ without selector


dlsniper:
The third parameter to ```fmt.Printf()``` should be flagged with an error saying: "Usage of package time without selector"

```go
package main

import (
	"fmt"
	"time"
)

func main() {
	fmt.Printf("%s - %s\n", time.Now().Format(time.RFC3339), time)
}
```

PR #2067, opened 2015-12-05
Add check for package used without selector

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1420.


ignatov:
Thanks.

PR #2068, opened 2015-12-06
add check that LHS of expr.(T) is interface type

sjamesr:
A type assertion in the form

expr.(T)

requires that expr is of interface type. This change adds an annotator check
for this requirement.

### DIFF
--- a/src/com/goide/highlighting/GoAnnotator.java
+++ b/src/com/goide/highlighting/GoAnnotator.java
@@ -68,5 +68,16 @@ public class GoAnnotator implements Annotator {
         }
       }
     }
+    else if (element instanceof GoTypeAssertionExpr) {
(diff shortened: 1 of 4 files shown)


sjamesr:
This check isn't quite right:

```go
package main

import "fmt"

type interf1 interface{}
type interf2 interf1
type interf3 interf2

func main() {
	var y interf3
	fmt.Println("Hello, playground")
	if _, ok := y.(string); ok {}  // plugin reports error, should not
}
```

sjamesr:
This should detect situations like the one described above, please take a look!

ignatov:
Thanks, merged manually due to conflicts.

ignatov:
671bbb6eb65b9de86c27eda79b574be4e26216d0

Issue #1422, opened 2015-03-08
Show const values in tooltips


dlsniper:
It should be nice to display a cost value in the tooltip, besides the ```const``` value that we have right now.

![snapshot17](https://cloud.githubusercontent.com/assets/607868/6545710/6dc1c556-c595-11e4-9170-14fd71fe13b5.png)

PR #2071, opened 2015-12-06
add definition annotation for const ref

sjamesr:
If a GoReferenceExpression is a reference to a const, an annotation will be
created with the text of the constant's definition.

Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1422.

### DIFF
--- a/src/com/goide/GoDocumentationProvider.java
+++ b/src/com/goide/GoDocumentationProvider.java
@@ -189,6 +189,22 @@ public class GoDocumentationProvider extends AbstractDocumentationProvider {
       String name = ((GoTypeSpec)element).getName();
       return StringUtil.isNotEmpty(name) ? "type " + name : "";
     }
+    if (element instanceof GoConstDefinition) {
+      if (element.getParent() instanceof GoConstSpec) {
+        GoConstSpec spec = (GoConstSpec) element.getParent();
+        StringBuilder result = new StringBuilder();
+        result.append(StringUtil.join(spec.getConstDefinitionList(), GoPsiImplUtil.GET_TEXT_FUNCTION, ","));
+        result.append(" ");
+        result.append(getTypePresentation(spec.getType()));
+        if (spec.getAssign() != null) {
+          result.append(" ");
+          result.append(spec.getAssign().getText());
+          result.append(" ");
+        }
+        result.append(StringUtil.join(spec.getExpressionList(), GoPsiImplUtil.GET_TEXT_FUNCTION, ","));
+        return result.toString();
+      }
+    }
     if (!(element instanceof GoSignatureOwner)) return "";
 
     PsiElement identifier = null;
--- a/testData/doc/constants.go
+++ b/testData/doc/constants.go
@@ -1,0 +1,13 @@
+package a
+
+type FooBar string
+
+const (
+	x FooBar = "something"
+)
+
+func foo(FooBar) {}
+
+func _() {
+	foo(x<caret>)
+}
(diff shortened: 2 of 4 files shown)


sjamesr:
It looks like this:

screen shot 2015-12-06 at 9 07 00 am

dlsniper:
I might be wrong but iirc, `getText` shouldn't be called in the `GoHighlightingAnnotator` (something to do with triggering a reparsing of the tree but I might be very wrong). 

Also, what happens if the const is a concatenation of other consts or `iota`? Should we deal with those?

I'd rather see this resolved in the ctrl+mouse-over from GoDocumentationProvider part.

sjamesr:
I'll rewrite using the GoDocumentationProvider, I think that is a better spot.

If the constant definition is some constant expression, should we try evaluating it? I suspect this is going to be a lot of work if it isn't already done somewhere else in the plugin.

How about something like this: if we have

```go
const (
    x = 4
    y = 1
    z = x + y
)
```

then the quick documentation could be:

```go
Constant z = x + y
x = 4
y = 1
```

For `iota`, we'd have to implement the rules to figure out its value. I think this shouldn't be too much work.

dlsniper:
I guess we could skip `iota` and concatenations to start with and add them later, based on feedback from users. The advantage of implementing this via GoDocProvider would be that people could potentially navigate to `x` and `y` on their own, which might not be ideal but it's better than nothing. What do you think?

sjamesr:
Oh that's a good idea, will do that!

sjamesr:
Ok, here's the next iteration:

screen shot 2015-12-06 at 10 25 29 am

dlsniper:
Nice! Thank you for your work!

src/com/goide/GoDocumentationProvider.java
```diff
@@ -173,6 +180,21 @@ private static String getSignature(PsiElement element) {
       String name = ((GoTypeSpec)element).getName();
       return StringUtil.isNotEmpty(name) ? "type " + name : "";
     }
+    if (element instanceof GoConstDefinition) {
+      if (element.getParent() instanceof GoConstSpec) {
+        GoConstSpec spec = (GoConstSpec) element.getParent();
+        StringBuilder result = new StringBuilder("const ");
+        result.append(StringUtil.join(spec.getConstDefinitionList(), ELEMENT_TO_TEXT_FUNC, ", ") + " ");
```
zolotov: reuse GoPsiImplUtil.GET_TEXT_FUNCTION

zolotov: get rid of yellow code please

src/com/goide/GoDocumentationProvider.java
```diff
@@ -173,6 +180,21 @@ private static String getSignature(PsiElement element) {
...
+        if (spec.getType() != null) {
+          result.append(getTypePresentation(spec.getType()));
```
zolotov: it's ok to pass null to `getTypePresentation`

src/com/goide/GoDocumentationProvider.java
sjamesr: Done and done

src/com/goide/GoDocumentationProvider.java
sjamesr: Fixed, thanks

sjamesr:
Comments addressed and rebased onto master

zolotov:
Thanks!

PR #2072, opened 2015-12-06
[WIP] Introduce the Go IDE into the repository

dlsniper:
# DO NOT USE THIS YET

### PLEASE DON'T +1 OR SIMILAR. THANK YOU!

-----

This PR aims to integrate the build scripts necessary to build an IDE for Go only.

The scripts are based on the ones used in the previous version of the plugin and developed by @mtoader but I foresee major changes to them. They provide a good enough start for now.

The current state is: the IDE is compiled and building but the plugin is not present in it.

Further steps include:
- [ ] better branding
- [ ] finding a place to host this (I guess I could do it over AppEngine)
- [ ] contacting @mtoader about the domain go-ide.com
- [ ] fixing everything else
- [ ] release into the wild

@ignatov / @zolotov I might need some help on this one but I plan to use https://devnet.jetbrains.com/thread/475879 as much as possible. If you could suggest who to ping for quicker responses that would be awesome.


zolotov:
Personally I don't see much value in it, so I won't participate in this PR. But I recommend to implement it from scratch, avoid using shell-scripts and to ask @ignatov and @cmf for help with it, since they have an experience with this.

ignatov:
@dlsniper I agree with @zolotov, it's better to avoid shell scripting and use our gant layouts for this stuff.

dlsniper:
Yup, will do. @cmf was awesome and compiled a pretty good list to start from here: https://devnet.jetbrains.com/message/5565633#5565633

I'll review them, and try to solve this with gant and so on. I'll update here when I make some progress on this, for now I have to wait a bit more before I can properly start on it.

Issue #497, opened 2014-01-06
plugin should show error,when missing argument to type conversion


bronze1man:
http://play.golang.org/p/GMP5gsVWSN
```go
package main

import "sort"

func main() {
	sort.IntSlice()
}
```

```
prog.go:6: missing argument to conversion to sort.IntSlice: sort.IntSlice()
 [process exited with non-zero status]
```
same need as #475

dlsniper:
This still happens under 1.0

PR #2074, opened 2015-12-07
add check to ensure that type conversion has 1 arg

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/497.


ignatov:
Thanks

Issue #2079, opened 2015-12-09
map literal syntax not enforced


sjamesr:
The plugin accepts the following code even though it contains an error:

```go
package main

import "fmt"

func main() {
	foo := map[string]string{"1", "2"}   // missing key in map literal
	fmt.Printf("%#v\n", foo)
}
```

The map literal is missing a key (probably the `,` should be a `:`).

PR #2080, opened 2015-12-09
add check for missing key in map literal

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2079.


zolotov:
Thanks.

PR #2087, opened 2015-12-11
extend assignment count check to short var decl

sjamesr:
Previously the plugin would not complain when doing something like

```go
x := 1, 2
```

whereas

```go
var x int = 1, 2
```

would produce the correct error.

### DIFF
--- a/src/com/goide/inspections/GoVarDeclarationInspection.java
+++ b/src/com/goide/inspections/GoVarDeclarationInspection.java
@@ -83,7 +83,8 @@ public class GoVarDeclarationInspection extends GoInspectionBase {
 
   @NotNull
   private static Pair<List<? extends GoCompositeElement>, List<GoExpression>> getPair(@NotNull GoVarSpec varDeclaration) {
-    PsiElement assign = varDeclaration.getAssign();
+    PsiElement assign =
+      varDeclaration instanceof GoShortVarDeclaration ? ((GoShortVarDeclaration)varDeclaration).getVarAssign() : varDeclaration.getAssign();
     if (assign == null) {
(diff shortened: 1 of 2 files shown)


dlsniper:
@sjamesr have you also looked into this: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1668 ?
I'll try and rebase it right now

dlsniper:
Do you think we should merge the effort in that PR or keep them separated?

sjamesr:
Hey @dlsniper, I think you should incorporate my patch into your work. There's definitely more work to be done, e.g.

```go
var z chan int
x, ok, y := <-z
```

is invalid but is accepted by the plugin even after the patch, but I believe the patch is correct given the functionality the plugin offers today.

dlsniper:
Right now this is already covered, see this test: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1668/files#diff-c6f77f9a80aaaa5817dbce7cb837bd88R3 but I do have a problem with resolving certain function calls / aliases.
If you'd like to help me get this into a better shape, that would be awesome. Thank you.

sjamesr:
@dlsniper I'll admit that I have not looked at your patch in detail, but if you're wrestling with type aliases you might want to look at https://github.com/sjamesr/go-lang-idea-plugin/commit/3b3e7c68eda58be8ee8221f4670f0e459960dce5

Basically it moves `GoFieldNameReference.findTypeFromRef` to a public method `GoPsiImplUtil.findBaseTypeFromRef`, which iteratively resolves type references until it finds a "concrete" type.

dlsniper:
Cool. Yes, I'm wrestling with it as the situations themselves can become quite complex. I'll wait for @ignatov to merge that one and then continue on #1668 then.

ignatov:
Thanks

Issue #2088, opened 2015-12-12
Inspection: make([]string) is not marked as an error.


vburenin:
make([]string) - this is definitely wrong, since it is missing at least length. However, code inspection doesn't treat it as an error.

https://play.golang.org/p/ecFXUXieoY

```
prog.go:4: missing len argument to make([]string)
```

IntelliJ: 15.0.2
GoLang plugin: 0.10.797

zolotov:
Works fine for me. Could you provide details?

vburenin:
![image](https://cloud.githubusercontent.com/assets/4350891/11761052/af8fc14c-a067-11e5-92c1-21389ce46952.png)

Compiler error is: testutils/memdbservice.go:112: missing len argument to make([]string)

dlsniper:
Hi,I've updated the description with a link to the playground. I've also tested it in the plugin and it indeed is not flagged as an error.

PR #2092, opened 2015-12-12
add checks to calls to builtin make func

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2088.

### DIFF
--- a/src/com/goide/highlighting/GoAnnotator.java
+++ b/src/com/goide/highlighting/GoAnnotator.java
@@ -19,6 +19,7 @@ package com.goide.highlighting;
 import com.goide.psi.*;
 import com.goide.psi.impl.GoPsiImplUtil;
 import com.goide.quickfix.GoReplaceWithReturnStatementQuickFix;
+import com.google.common.collect.Sets;
 import com.intellij.lang.annotation.Annotation;
 import com.intellij.lang.annotation.AnnotationHolder;
 import com.intellij.lang.annotation.Annotator;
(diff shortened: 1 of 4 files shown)


dlsniper:
Hi. Thanks for the PR. However, I think there's a problem with calling `getText()` from the `Annotator` implementations and I think this should be fixed in a different place. At least, if I recall this correctly, every time when calling `getText()` from it it triggers a PSI tree refresh, which wouldn't be a good idea. @zolotov / @ignatov please confirm I'm not remembering this wrong. If it's wrong, sorry for the noise.

sjamesr:
If this is indeed a problem, I can move it to its own inspection.

ignatov:
It's not a problem to get a text of local PSI element, but it's too expensive for stub element. In your case it looks good.

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -80,13 +96,110 @@ else if (element instanceof GoTypeAssertionExpr) {
...
+
+    if (type instanceof GoSpecType) {
+      type = ((GoSpecType)type).getType();
+    }
+
+    return type;
+  }
+
+  private static boolean isIntegerType(GoType type) {
+    return intTypeNames.contains(type.getText());
```
ignatov: Type can be a stub, so it's definitely not a good idea to use getText.

sjamesr: Is there another way to find the type's name? Is this check simply not suitable for the GoAnnotator?

ignatov: Please use `GoPsiImplUtil#getTypeSpecSafe`

sjamesr: Can you give me an example of how I might use it here? For example, in the test I have

```go
type stringType string

make([]stringType, ...)
```

I determine the type of the first argument is a `GoArrayOrSliceType`, calling `getTypeSpecSafe` on that gives me `null`.

I'm unsure how to proceed.

sjamesr:
@ignatov I think I've removed all calls to getText on stubbed elements. The error messages don't match Go's exactly, but I think it's fine.

ignatov:
BTW, we have a special stub with text: com.goide.stubs.StubWithText

ignatov:
I'm going to take a look to the PR today or tomorrow.

sjamesr:
@ignatov regarding StubWithText, does this mean it's now safe to call getText on a GoType in GoAnnotator without triggering any lengthy re-parsing?

ignatov:
For sure, calling getText on remote (resolved) element is potentially dangerous.

sjamesr:
Ok, I was doing plenty of that previously. :) Hopefully this latest pull request is ready to go.

ignatov:
Merged manually, thanks a lot for your help!

PR #2113, opened 2015-12-18
Go 1.6 SDK detection

dlsniper:
As reported by @stuartcarnie on Gitter, Go 1.6 has moved `zversion.go`; now found in `runtime/internal/sys/zversion.go`

This PR addresses that.

May I also suggest releasing the plugin on all channels (nightly, alpha and stable) with this change? I know from previous experience about the number of tickets that the old plugin didn't supported Go 1.4 (or was it 1.5?) when that appeared and changed the way this works.

### DIFF
--- a/src/com/goide/sdk/GoSdkUtil.java
+++ b/src/com/goide/sdk/GoSdkUtil.java
@@ -59,9 +59,9 @@ import java.util.regex.Pattern;
 import static com.intellij.util.containers.ContainerUtil.newLinkedHashSet;
 
 public class GoSdkUtil {
-  private static final Pattern GO_VERSION_PATTERN = Pattern.compile("theVersion\\s*=\\s*`go([\\d.]+\\w+(\\d+)?)`");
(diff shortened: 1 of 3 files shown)


stuartcarnie:
:+1:

src/com/goide/sdk/GoSdkUtil.java
```diff
@@ -59,9 +59,9 @@
 import static com.intellij.util.containers.ContainerUtil.newLinkedHashSet;
 
 public class GoSdkUtil {
-  private static final Pattern GO_VERSION_PATTERN = Pattern.compile("theVersion\\s*=\\s*`go([\\d.]+\\w+(\\d+)?)`");
-  private static final Pattern GAE_VERSION_PATTERN = Pattern.compile("theVersion\\s*=\\s*`go([\\d.]+)( \\(appengine-[\\d.]+\\))?`");
-  private static final Pattern GO_DEVEL_VERSION_PATTERN = Pattern.compile("theVersion\\s*=\\s*`(devel.*)`");
+  private static final Pattern GO_VERSION_PATTERN = Pattern.compile("(?:t|T)heVersion\\s*=\\s*`go([\\d.]+\\w+(\\d+)?)`");
```
zolotov: why not just [tT]?

PR #2117, opened 2015-12-20
Ignore embedded types when finding duplicate methods

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2099.

This should be considered a short-term fix until general embedded type support is implemented.


dlsniper:
lgtm

Issue #1968, opened 2015-11-05
Inspect arguments in nested struct methods


bictorman:
```go
package main

type Foo struct {
	bar Bar
}

type Bar struct {}
func (b *Bar) Args(a string) {}
func (b *Bar) NoArgs() {}

func main() {
	f := new(Foo)
	f.bar.Args() // should be an error
	f.bar.NoArgs("argument") // should be an error
}
```

```
./main.go:13: not enough arguments in call to f.bar.Args
./main.go:14: too many arguments in call to f.bar.NoArgs
```

PR #2119, opened 2015-12-20
Rework parameter count inspection (fixes #2115)

### LINKED ISSUE #2115: Rework count mismatch inspection
```go
package main

import (
	"fmt"
	"reflect"
)

type b func(intParam int)
type a b

func c() {
	aFunc := generateFunc()
	if f, ok := aFunc.(a); ok {
		fmt.Println(reflect.TypeOf(f))
		f(<caret>) // should be not enough arguments in call to f
	}
}

func generateFunc() interface{} {
	return a(func (intParam int) { })
}

func main() {
	c()
}
```

See com.goide.editor.GoParameterInfoHandler#showParameterInfo

dlsniper:
@ignatov not sure if this is what you had in mind but let me know if I need to rework anything else

### DIFF
--- a/testData/highlighting/funcCall.go
+++ b/testData/highlighting/funcCall.go
@@ -144,7 +144,7 @@ func bar_m() (int, int) {
 }
 
 func <warning>main127</warning>() {
-	foo_m<error>(bar_m())</error>
+	<error descr="too many arguments in call to foo_m">foo_m</error>(bar_m())
 }
 
 type AnInterface interface {
@@ -166,7 +166,7 @@ func (x *mystruct) getAnInterface() AnInterface {
 var getAnInterface = (*mystruct).getAnInterface
 
 func _() {
-	fmt.Println(getAnInterface(&mystruct{}).MyMethod())
+	fmt.Println(<error descr="too many arguments in call to getAnInterface">getAnInterface</error>(&mystruct{}).MyMethod())
 }
 
 type (
--- a/tests/com/goide/inspections/GoHighlightingTest.java
+++ b/tests/com/goide/inspections/GoHighlightingTest.java
@@ -146,6 +146,7 @@ public class GoHighlightingTest extends GoCodeInsightFixtureTestCase {
   public void testPlaceholderCount()          { doTest(); }
   public void testPlaceholderCountVet()       { doTest(); }
   public void testTypeConversion()            { doTest(); }
+  public void testParameterCount()            { doTest(); }
   public void testInit()                      { doTest(); }
   public void testMainWithWrongSignature()    { doTest(); }
   public void testChan()                      { doTest(); }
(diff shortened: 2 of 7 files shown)


.idea/codeStyleSettings.xml
```diff
@@ -225,4 +225,4 @@
     </option>
     <option name="USE_PER_PROJECT_SETTINGS" value="true" />
   </component>
-</project>
\ No newline at end of file
+</project>
```
ignatov: Hmm, it's definitely from another change list.

gradle.properties
```diff
@@ -26,4 +26,4 @@ sources=true
 publishUsername=user
 publishPassword=password
 publishChannel=
-org.gradle.jvmargs=-Xms256m -Xmx1024m
\ No newline at end of file
+org.gradle.jvmargs=-Xms256m -Xmx1024m
```
ignatov: And this one.

dlsniper:
Removed unrelated misc file changes

dlsniper:
Once this gets merged I plan to work on fixing: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1968 as currently this PR doesn't cover it. However I want to keep the PR smaller than the one with assignment count mismatch so I'd rather handle one issue at a time as it appears this time around. Thank you

dlsniper:
Updated to be mergeable again (I'll update it as soon as it changes state if again).

dlsniper:
Updated

dlsniper:
Updated

dlsniper:
Updated this as well.

testData/highlighting/funcCall.go
```diff
@@ -143,8 +143,8 @@ func bar_m() (int, int) {
 	return 0, 0
 }
 
-func <warning>main127</warning>() {
-	foo_m<error>(bar_m())</error>
+func <warning descr="Unused function 'main127'">main127</warning>() {
```
ignatov: Please don't add such markup without need, it's very hard to review.

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -135,6 +135,7 @@ protected String getBasePath() {
   public void testEmbeddedInterfacePointer()  { doTest(); }
   public void testPlaceholderCount()          { doTest(); }
   public void testTypeConversion()            { doTest(); }
+  public void testGh2115()                    { doTest(); }
```
ignatov: Better to create a test with a meaningful name, not a issue number.

dlsniper: I'll update it, yes.

testData/highlighting/funcCall.go
dlsniper: I'll try and remove those, I've understood meanwhile why they appear in the tests since I've wrote this :)

dlsniper:
Done, I've fixed the previous comments, added the extra test cases from the other two related issues and made the suggestion @zolotov had regarding the position of where the error is reported, it's now on the function name not the the whole arguments list (which makes it more visible when there are no arguments).

dlsniper:
The only thing that's different is the error message vs the Go compiler in some cases but I think it's more readable this way (I can use the compiler messages if needed).

PR #2120, opened 2015-12-20
Introduce cyclic definition inspection (fixes #2096)

### LINKED ISSUE #2096: Do not resolve const/var to itself
```go
package main

var (
	Foo int = 1 + Foo
)
const (
	Bar int = 1 + Bar
)
```

Both cases are initialization loops. Foo and Bar shouldn't be resolved. Might be covered with annotator but I think it's better do not resolve it at all

dlsniper:


dlsniper:
As there's a lot of duplication here I wonder if we could (should) improve the PSI structure to support both types from a base one.

dlsniper:
I've updated this. Unfortunately I still don't know how to unify this into a single function.

zolotov:
Move to annotator please

src/com/goide/inspections/GoCyclicDefinitionInspection.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+
+public class GoCyclicDefinitionInspection extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull final LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitReferenceExpression(@NotNull GoReferenceExpression o) {
+        super.visitReferenceExpression(o);
```
zolotov: You shouldn't check all reference expressions. You should check only expressions inside constDefinition value or varDefinition value.

src/com/goide/inspections/GoCyclicDefinitionInspection.java
```diff
@@ -0,0 +1,82 @@
...
+  private static boolean check(@NotNull GoReferenceExpression o, GoConstDefinition constDefinition) {
+    GoConstSpec constSpec = PsiTreeUtil.getParentOfType(o, GoConstSpec.class);
+    if (constSpec == null) return false;
+
+    List<GoExpression> expressionList = constSpec.getExpressionList();
+    List<GoConstDefinition> definitionList  = constSpec.getConstDefinitionList();
+    PsiElement element = PsiTreeUtil.getTopmostParentOfType(o, GoExpression.class);
+    if (element == null) return false;
+
+    int index = expressionList.indexOf((GoExpression)element);
```
zolotov: GoConstDefinition has getValue method

dlsniper:
@zolotov I've moved this to the `GoAnnotator` and improved based on the previous feedback as well.

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -67,6 +67,16 @@ else if (element instanceof GoReferenceExpression) {
       if (resolvedReference instanceof GoTypeSpec && isIllegalUseOfTypeAsExpression(reference)) {
         holder.createErrorAnnotation(element, "Type " + element.getText() + " is not an expression");
       }
+      if (PsiTreeUtil.getParentOfType(element, GoConstDeclaration.class) != null &&
+          resolvedReference instanceof GoConstDefinition &&
+          resolvedReference.getParent() instanceof GoConstSpec) {
+        checkSelfReference((GoReferenceExpression)element, (GoConstDefinition)resolvedReference, holder);
```
zolotov: there is already casted `reference` variable

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -67,6 +67,16 @@ else if (element instanceof GoReferenceExpression) {
       if (resolvedReference instanceof GoTypeSpec && isIllegalUseOfTypeAsExpression(reference)) {
         holder.createErrorAnnotation(element, "Type " + element.getText() + " is not an expression");
       }
+      if (PsiTreeUtil.getParentOfType(element, GoConstDeclaration.class) != null &&
```
zolotov: it makes sense to move this predicate at the end, since it most expensive call

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -186,6 +196,24 @@ private static boolean isIntegerType(@Nullable GoType type) {
...
+    if (constDefinition.getValue() == null) return;
+    PsiElement element = PsiTreeUtil.getTopmostParentOfType(o, GoExpression.class);
+    if (element == null) return;
+    if (constDefinition.getValue().equals(element)) {
+      holder.createErrorAnnotation(o, "Cyclic definition detected");
+    }
+  }
+
+  private static void checkSelfReference(@NotNull GoReferenceExpression o, GoVarDefinition varDefinition, AnnotationHolder holder) {
+    if (varDefinition.getValue() == null) return;
```
zolotov: extract value variable please

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -186,6 +196,24 @@ private static boolean isIntegerType(@Nullable GoType type) {
     return GoPsiImplUtil.builtin(type) && INT_TYPE_NAMES.contains(ref.getText());
   }
 
+  private static void checkSelfReference(@NotNull GoReferenceExpression o, GoConstDefinition constDefinition, AnnotationHolder holder) {
+    if (constDefinition.getValue() == null) return;
```
zolotov: extract value variable please

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -186,6 +196,24 @@ private static boolean isIntegerType(@Nullable GoType type) {
...
+    PsiElement element = PsiTreeUtil.getTopmostParentOfType(o, GoExpression.class);
+    if (element == null) return;
```
zolotov: this check is redundant

src/com/goide/highlighting/GoAnnotator.java
zolotov: actually you can do something like:
```
    GoExpression value = constDefinition.getValue();
    if (value != null && value.equals(PsiTreeUtil.getTopmostParentOfType(o, GoExpression.class))) {
      holder.createErrorAnnotation(o, "Cyclic definition detected");
    }
```

dlsniper:
Done and simplified the code a bit as well.

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -186,6 +196,22 @@ private static boolean isIntegerType(@Nullable GoType type) {
...
+    if (definition instanceof GoVarDefinition) {
+      value = ((GoVarDefinition)definition).getValue();
+    }
+    else if (definition instanceof GoConstDefinition) {
+      value = ((GoConstDefinition)definition).getValue();
+    }
+    if (value == null) return;
+
+    PsiElement element = PsiTreeUtil.getTopmostParentOfType(o, GoExpression.class);
+    if (element == null) return;
```
zolotov: This check is still redundant

dlsniper: I've seen the comment after pushing it, updated already. Thanks.

PR #2125, opened 2015-12-22
Introduce inspection for structure initialization (fixes #2094)

### LINKED ISSUE #2094: Inspection for initializing structs without using explicit struct field names
Given the following struct:
```go
struct {
	a string
	b int32
	c string
}
```
While initializing the struct can be done without explicitly using the field names, it's recommended to do so as any changes in the struct layout could potentially affect the functionality of the program.

With the example above, the following change can potentially be error prone:
```go
struct {
	c string
	b int32
	a string
}
```

Also, the initializing part can suffer as well if a field is changed.

dlsniper:


src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Implements <a href="https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1983"/>, an
+ * inspector that warns if a go StructTag is not well-formed according to Go language conventions.
+ */
+public class GoStructInitializationInspection extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder,
+                                     @SuppressWarnings({"UnusedParameters", "For future"}) @NotNull LocalInspectionToolSession session) {
```
zolotov: please delete @SuppressWarnings annotation

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -0,0 +1,61 @@
...
+        GoType type = o.getType();
+        if (!(type instanceof GoStructType) &&
+            !(GoPsiImplUtil.findBaseTypeFromRef(o.getTypeReferenceExpression()) instanceof GoSpecType)) {
+          return;
+        }
+
+        boolean hasUnitializedFieldInitialization = false;
+        for (GoElement element : o.getLiteralValue().getElementList()) {
+          if (element.getKey() == null) {
+            hasUnitializedFieldInitialization = true;
```
zolotov: so if getKey is null then you should registerProblem and return, right? Why do you need `hasUnitializedFieldInitialization` then?

dlsniper:
Updated.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -0,0 +1,56 @@
...
+        if (PsiTreeUtil.getParentOfType(o, GoReturnStatement.class, GoShortVarDeclaration.class, GoAssignmentStatement.class) == null) return;
+        GoType type = o.getType();
+        if (!(type instanceof GoStructType) &&
+            !(GoPsiImplUtil.findBaseTypeFromRef(o.getTypeReferenceExpression()) instanceof GoSpecType)) {
+          return;
+        }
+
+        for (GoElement element : o.getLiteralValue().getElementList()) {
+          if (element.getKey() == null) {
+            holder.registerProblem(o, "Structure has unnamed field initialization", ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
```
zolotov: Maybe it's better to highlight the element without key instead of whole composite literal? How do you think?

Also, maybe quickfix will be useful here? E.g. it can add key based on field order

dlsniper: I was also thinking about the quickfix and there are two possibilities here:
- either remove the pointer
- introduce a name for the field

zolotov: I think both are valid

sjamesr: There's a similar missing key check in the GoAnnotator:

https://github.com/go-lang-plugin-org/go-lang-idea-plugin/commit/aa6b2b92a3427486ddf9bd3a8f224ffc6ee2487f#diff-3c03889681be146ba35095e41c206df9R66

We should be consistent between the two.

dlsniper: In that case, the compiler will throw an error.
In this case things will work, but if there's a refactoring done, they will break. That's why I've made this as a weak warning and not an warning / error. 
I've also mixed-up the PRs. for this a quickfix would be to add the field names, however since one can only partially initialize the struct, we don't know which field to use (maybe display a list of fields is a good idea then).

zolotov: > maybe display a list of fields is a good idea then

It's a good idea

dlsniper:
Updated

testData/highlighting/unitializedStructInitialization.go
```diff
@@ -0,0 +1,30 @@
...
+func _() (struct {
+	x string
+}) {
+	_ = struct{ x string }{x: "demo"}
+	_ = <weak_warning descr="Structure has unnamed field initialization">demo{"demo"}</weak_warning>
+	b, _ := <weak_warning descr="Structure has unnamed field initialization">demo{"demo"}</weak_warning>, 1
+	_ = <weak_warning descr="Structure has unnamed field initialization">demo{"demo", 1, "demo"}</weak_warning>
+	_ = demo{a: "demo"}
+	_ = demo{a: "demo", b: 1}
+	_ = <weak_warning descr="Structure has unnamed field initialization">demo{
```
zolotov: I believe it's not passed now. Also it's not `structure has field` anymore, it's `unnamed field`

dlsniper: > I believe it's not passed now. 

Not sure what it means, the tests passed on last run.

> Also it's not structure has field anymore, it's unnamed field

Actually that case should be an error: `mixture of field:value and value initializers` now that I think of it, see: https://play.golang.org/p/qZpTKwV6q0 

Should I introduce a different inspection for that and abort this inspection if it would match for the other case?

zolotov: Sorry, I've missed that you didn't change reporting element. As @sjamesr said, it's good to keep inspection in consistent state, so if `missing key` error is reported on GoElement then this inspection also should be reported on GoElement too, not on whole literal

zolotov: > Should I introduce a different inspection for that and abort this inspection if it would match for the other case?

Yes, I think it's different case with different inspection level and should be covered by another inspection

dlsniper:
I've updated this to emit the warning on the fields and not the struct itself.
I'll create a separate issue for the other inspection.
I've created the issue for the quickfix for this one.

PR #2126, opened 2015-12-22
Introduce method defined on non-local type inspection (fixes #2031)

### LINKED ISSUE #2031: Inspection for defining methods on non-local types
Incorrect version:
```go
package main

func (s *string) Hello() { // cannot define new methods on non-local type string

}
```

QuickFix version:

```go
package main

type myString *string

func (s myString) Hello() {

}
```

dlsniper:


dlsniper:
I need a bit of help with this one because currently the inspection test will fail as the file which contains the method returns `null` and I don't really know how to handle this case. Should the inspection automatically fail then or not? I lean towards yes but I'm not sure I'm reading the code correctly. Thank you.

zolotov:
Do you mean that getImportPath returns null for file? Indeed, you need some preparation for this, please take a look at `GoHighlightingTest#testPackageWithTestPrefix` test

dlsniper:
Updated. I hope it's ok how I've solved creating the file, it felt strange to copy so much text inside the test function.

PR #2127, opened 2015-12-22
Introduce inspection for range iteration on illegal type (fixes #1847)

### LINKED ISSUE #1847: For iteration on a pointer slice is not valid
This is a continuation of #1837. The difference is that in that example `data` is initialized via a call to a function, here it's initialized directly 

```go
package main

type TestStruct struct {
	SomeId int
}

type DataSlice []*TestStruct

func _() {
	data := &DataSlice{}
	for _, element := range data {  // here we should have an error: cannot range over data (type *DataSlice)
		if element.SomeId > 20 {
			println("some text")
		}
	}
}

func _() {
	data := DataSlice{}
	for _, element := range &data {  // here we should have an error: cannot range over cannot range over &data (type *DataSlice)
		if element.SomeId > 20 {
			println("some text")
		}
	}
}
```

dlsniper:


dlsniper:
I've omitted accurate checks for all the situations as I'd first want to see how this works in real code. I've inspected the Go SDK and I couldn't find any issues with this. Hope that's ok.

src/com/goide/inspections/GoRangeIterationOnIllegalTypeInspection.java
```diff
@@ -0,0 +1,80 @@
...
+
+  private static boolean isValidType(GoType goType) {
+    if (goType instanceof GoSpecType) {
+      goType = ((GoSpecType)goType).getType();
+    }
+
+    return goType instanceof GoArrayOrSliceType ||
+           goType instanceof GoMapType ||
+           goType instanceof GoChannelType ||
+           goType.getText().equals("string");
```
zolotov: please check whether it's builtin `string`

dlsniper:
Updated.

src/com/goide/inspections/GoRangeIterationOnIllegalTypeInspection.java
```diff
@@ -0,0 +1,80 @@
...
+
+    return goType instanceof GoArrayOrSliceType ||
+           goType instanceof GoMapType ||
+           goType instanceof GoChannelType ||
+           (goType.textMatches("string") && GoPsiImplUtil.builtin(goType));
+  }
+
+  private static boolean isValidTypeRefExpression(GoType goType) {
+    if (goType instanceof GoParType) {
+      goType = ((GoParType)goType).getType();
```
zolotov: what if there is a parType inside parType?

dlsniper: I would like to keep this for now and when we have a better way to get to the GoType of an element / reference use that. @ignatov said he'd like to write such a method so we can either wait until that happens or merge this and then do a pass on all the inspections and have them use it.

I know there are other cases that are not covered as well, and not just in this inspection, like type casting for example.

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -357,7 +358,7 @@ public void testCGOImportInNonTestFile() {
     myFixture.configureByText("a.go", "package a; import \"C\"");
     myFixture.checkHighlighting();
   }
-  
+
```
ignatov: The same

src/com/goide/inspections/GoRangeIterationOnIllegalTypeInspection.java
ignatov: So, will wait for a such method, right?

dlsniper: Depends when you think it will be ready to be used. I can rework it for now but I do need some direction for it :)

dlsniper:
I've updated this as ParType is now unwrapped until the base type automatically (thanks Sergey). There's a test for this in `testData/highlighting/ranges.go` (see `(((c)))` which is  channel).

zolotov:
merged manually. please see `GoRangeIterationOnIllegalTypeInspection`

Btw, why did you introduce `expressions.size == 1` check?

dlsniper:
> merged manually. please see GoRangeIterationOnIllegalTypeInspection

Thank you, I've seen it, looks much better.

> Btw, why did you introduce expressions.size == 1 check?

It was introduced initially when we didn't had a way to differentiate between the various expression types of the for loop (iirc you've recently added this feature).

zolotov:
> It was introduced initially when we didn't had a way to differentiate between the various expression types of the for loop (iirc you've recently added this feature).

Ah, I see, thanks

PR #2130, opened 2015-12-23
Introduce declared but not used inspection for literals (fixes #1310)

### LINKED ISSUE #1310: Function cannot be declared in another function body without being used
The continuation of #1308 
```go
func demo() {
    func(){} //func literal evaluated but not used
}
```

dlsniper:


dlsniper:
I've fixed the previous issues with code not being annotated.
@ignatov I think these are the specs, if I understand the question correctly: https://golang.org/ref/spec#Expressions
The issue is that for example, one can't define a function inside another function/method and not call it.

mtoader:
That spec doesn't seems to cover this inspection. The expression spec only specifies what an expression is. Maybe a pointer to the actual messages that the go compiler is firing when this happens ?

dlsniper:
I could offer this playground example https://play.golang.org/p/KRT_O0TgVI which also shows the current implementation state.

dlsniper:
Updated.

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -110,6 +110,46 @@ else if (element instanceof GoBuiltinCallExpr) {
...
+    if (parent == null) return;
+    if (parent instanceof GoArgumentList) return;
+    if (parent instanceof GoLeftHandExprList &&
+        ((GoLeftHandExprList)parent).getExpressionList().size() != 1) {
+      return;
+    }
+
+    if (element instanceof GoFunctionLit) {
+      element = ((GoFunctionLit)element).getFunc();
+    } else {
```
ignatov: Code style

tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -357,7 +358,7 @@ public void testCGOImportInNonTestFile() {
     myFixture.configureByText("a.go", "package a; import \"C\"");
     myFixture.checkHighlighting();
   }
-
```
ignatov: Just an advice: please try not to change unnecessary lines – you PR looks dirty after that.

dlsniper: I know but that's strange it happens. Shouldn't the code style be applied from the plugin repository? I haven't done anything special for this and I have the other settings such as always trim whitespaces and so on applied because of the other projects I'm working on.
Would it be ok if I run formatting on the whole project once when you guys aren't developing anything on it so I don't interfere?

src/com/goide/highlighting/GoAnnotator.java
dlsniper: Done.

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -110,6 +110,47 @@ else if (element instanceof GoBuiltinCallExpr) {
...
+    if (parent instanceof GoArgumentList) return;
+    if (parent instanceof GoLeftHandExprList &&
+        ((GoLeftHandExprList)parent).getExpressionList().size() != 1) {
+      return;
+    }
+
+    if (element instanceof GoFunctionLit) {
+      element = ((GoFunctionLit)element).getFunc();
+    }
+    else if (parent.getParent().getParent() instanceof GoSwitchStatement) {
```
ignatov: What does it mean? And do we have a test for this branch?

dlsniper: This is useful to solve cases like this:
```go
	switch 1 {
		case 1: println("1")
	}
```

I've added a test for it as well.

ignatov: Is it possible to use literal in the same positions in other statements?

ignatov: Please run annotator on your code base and check that we haven't any false positive cases. 
Use Ctrl+Alt+Shift+I: annotator 
screenshot 2015-12-24 12 19 39

dlsniper: No false positives as far as I can see on my GOPATH / Go SDK.
While inspecting the Go SDK however I've got: https://ea.jetbrains.com/browser/ea_reports/1010007 but I'm not sure if it's related or not.
Also, it does take quite a bit to perform all the inspections as it seems, do you think I should move this one up to it's own inspection?

ignatov:
Merged manually, thanks

PR #2131, opened 2015-12-23
Introduce inspection for embedded interface pointer (fixes #1374)

### LINKED ISSUE #1374: Should be an error: embedded type cannot be a pointer to interface
```go
package main

import "fmt"

func main() {
    c := Parent{&Child{}, &impl{}}
    c.Get(1,1,1)
}

type Parent struct {
    *Child
    *Interface // embedded type cannot be a pointer to interface
}
type Child struct { Unimplemented }
type Interface interface { Get(...int) }
type impl struct { }
func (r impl) Get(h ...int) { fmt.Println("HERE") }
type Unimplemented interface { Get(s string) }
```

dlsniper:


src/com/goide/inspections/GoEmbeddedInterfacePointerInspection.java
```diff
@@ -0,0 +1,44 @@
...
+import org.jetbrains.annotations.NotNull;
+
+public class GoEmbeddedInterfacePointerInspection extends GoInspectionBase {
+  @NotNull
+  @Override
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitAnonymousFieldDefinition(@NotNull GoAnonymousFieldDefinition o) {
+        if (o.getMul() == null) return;
```
zolotov: please add test when mul == null

dlsniper: Done

zolotov:
thanks!

Issue #2132, opened 2015-12-23
Wrong triggering error "Non-integer size argument to make"


linniksa:
```go
package main

import "fmt"

func main() {
	type test uint32
	foo := make([]int, test(2))
	fmt.Println(foo)
}
```

```go
package main

import "fmt"

func main() {
	foo := make([]int, uint64(2))
	fmt.Println(foo)
}
```

ignatov:
@sjamesr Please take a look.

sjamesr:
@linniksa thank you for the report, the fix is in review now

uvwild:
still happeing in 2018.1 EAP in an example file:
https://github.com/manuelkiessling/go-cleanarchitecture/blob/96a022b6127638236963ed8f00acc499300a7b70/src/usecases/usecases.go

i disabled all inspections... its still there.... very frustrating.... 
when I cannot disable false positives  the inspections  become a bit useless....

dlsniper:
@uvwild I cannot replicate your problem with 2018.1 EAP. It is likely a problem generated by your configuration of the GOPATH, as that repository is "special" in terms of handling that and the author decided to roll his own layout.

Please observe the correct configuration in the IDE for the GOPATH and open the project root directory in the same place as the GOPATH is pointing out.

Furthermore, please open any further issues or comments on: https://youtrack.jetbrains.com/issues/Go as this tracker is not the correct place to do so.

![2018-02-10](https://user-images.githubusercontent.com/607868/36067480-0ed4eb82-0eb6-11e8-9e65-03bc267a35f8.png)

PR #2133, opened 2015-12-23
Fix builtin make integer type check

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2132.

### DIFF
--- a/src/com/goide/highlighting/GoAnnotator.java
+++ b/src/com/goide/highlighting/GoAnnotator.java
@@ -35,7 +35,7 @@ import java.util.Set;
 
 public class GoAnnotator implements Annotator {
   private static final Set<String> INT_TYPE_NAMES = Sets.newHashSet(
-    "int", "int8", "int16", "int32", "int64", "uint", "uint8", "uint16", "uint32", "uintptr"
+    "int", "int8", "int16", "int32", "int64", "uint", "uint8", "uint16", "uint32", "uint64", "uintptr"
   ); // todo: unify with DlvApi.Variable.Kind
 
   @Override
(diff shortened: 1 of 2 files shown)


dlsniper:
@sjamesr does this work for:
```go
	var b interface{} = 1
	_ = make(chan int, b.(int))
```
as well?

sjamesr:
@dlsniper I believe so, I just added a test for it

.idea/codeStyleSettings.xml
```diff
@@ -133,7 +133,6 @@
           <option name="ALIGN_MULTILINE_THROWS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_EXTENDS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_PARENTHESIZED_EXPRESSION" value="true" />
-          <option name="ALIGN_SUBSEQUENT_SIMPLE_METHODS" value="true" />
```
zolotov: ?

sjamesr: reverted, sorry

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -183,7 +186,8 @@ private static boolean isIntegerType(@Nullable GoType type) {
     if (type == null) return false;
     GoTypeReferenceExpression ref = type.getTypeReferenceExpression();
     if (ref == null) return false;
-    return GoPsiImplUtil.builtin(type) && INT_TYPE_NAMES.contains(ref.getText());
+    PsiElement resolve = ref.getReference().resolve();
+    return GoPsiImplUtil.builtin(resolve) && INT_TYPE_NAMES.contains(ref.getText());
```
zolotov: it makes sense to filter out references by name before resolving

sjamesr: Done

zolotov:
Very fast response. Thanks :)

PR #2134, opened 2015-12-23
[WIP] Introduce package should be named as containing directory inspection (fixes #1087)

### LINKED ISSUE #1087: Inspection for "Package name should be set to"
When the folder name is "demo" and the package in file "file.go" is named "package alpha" then the inspection flags it as an error when it should be just an info.

dlsniper:


dlsniper:
This is a bit work in progress as I need to fix some of the tests.

zolotov:
Do you think it's a valid inspection? I didn't see such recommendation about package naming. Also directory name is not always a valid package name

dlsniper:
From my point of view yes, but I'd like to hear others as well on this. @sjamesr @stuartcarnie @linniksa any other Go users, any input for this?

dlsniper:
I'll close this for now.

Issue #2139, opened 2015-12-24
False-positive for invalid type assertion inspection


dlsniper:
```go
package demo

type (
	int64Val int64
	boolVal  bool
	Value    interface {
		Demo()
	}
)

func _(x, y Value) {
	switch x := x.(type) {
	case boolVal:
		y := y.(boolVal)
		_, _ = x, y
	case int64Val:
		b := int64(y.(int64Val)) // this is seen as the check on boolVal not on int64Val
		_ = b
	}

}
```

sjamesr:
Actually I think this is a bug in type resolution:

In the line `b := int64(y.(int64Val))`, calling `getGoType` on the expression `y` yields `boolVal`, whereas it should be `Value`.

@ignatov depending on how complicated this is to fix, we should consider disabling the invalid type assertion check until getType is fixed.

sjamesr:
Looking further, resolving the reference `y` yields the var definition `y := y.(boolVal)`, not the parameter `y`. Since the var definition is in a separate case branch, it should not be considered.

PR #2163, opened 2015-12-24
Fix decls in case clause "leaking" to next clauses

sjamesr:
Previously, the plugin did not consider a case statement to be a separate scope
for reference resolution purposes.

Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2139.


ignatov:
Thanks

Issue #2140, opened 2015-12-24
False positive for non-integer size argument for make


dlsniper:
```go
package demo

import "C"

func _() {
	i := C.int(1)
	_ = make([]*C.char, i) // C references should be considered valid (at least until we can interop with C (via CLion?)
}
```

PR #2166, opened 2015-12-24
Whitelist C types from make args typecheck

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2140.


sjamesr:
The other solution may be to remove the isBuiltin check, but obviously this leads to false negatives when people alias builtin types.

PR #2167, opened 2015-12-24
Go lang std template initial support

jhsx:
Hi everyone, this pull request includes an initial support for the text/template package.

### DIFF
--- a/src/com/goide/psi/impl/GoPsiImplUtil.java
+++ b/src/com/goide/psi/impl/GoPsiImplUtil.java
@@ -534,7 +534,8 @@ public class GoPsiImplUtil {
 
   @Nullable
   private static GoType findTypeInVarSpec(@NotNull GoVarDefinition o, @Nullable ResolveState context) {
-    GoVarSpec parent = (GoVarSpec)o.getParent();
+    GoVarSpec parent = (GoVarSpec)PsiTreeUtil.getStubOrPsiParent(o);
+    if (parent == null) return null;
     GoType commonType = parent.getType();
     if (commonType != null) return commonType;
     List<GoVarDefinition> varList = parent.getVarDefinitionList();


.idea/codeStyleSettings.xml
```diff
@@ -133,7 +133,6 @@
           <option name="ALIGN_MULTILINE_THROWS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_EXTENDS_LIST" value="true" />
           <option name="ALIGN_MULTILINE_PARENTHESIZED_EXPRESSION" value="true" />
-          <option name="ALIGN_SUBSEQUENT_SIMPLE_METHODS" value="true" />
```
dlsniper: Please revert this.

.idea/vcs.xml
```diff
@@ -12,5 +12,6 @@
   </component>
   <component name="VcsDirectoryMappings">
     <mapping directory="$PROJECT_DIR$" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
```
dlsniper: Please revert this

dlsniper:
Please remove: JFlex.jar from the PR, it's not needed.

resources/META-INF/plugin.xml
```diff
@@ -93,8 +93,17 @@
     <psi.fileReferenceHelper implementation="com.goide.psi.impl.imports.GoImportReferenceHelper"/>
     <fileContextProvider implementation="com.goide.psi.impl.imports.GoFileContextProvider"/>
 
-    <lang.parserDefinition language="go" implementationClass="com.goide.GoParserDefinition"/>
+
+    <lang.fileViewProviderFactory language="tpl" implementationClass="com.goide.gotemplate.file.GoTemplateFileViewProviderFactory" />
```
dlsniper: I'm not sure if the language name is tpl, maybe gotpl to be even more specific?

src/com/goide/GoFileType.java
```diff
@@ -62,8 +62,4 @@ public String getCharset(@NotNull VirtualFile file, @NotNull byte[] content) {
     return CharsetToolkit.UTF8;
   }
 
-  @Override
-  public Charset extractCharsetFromFileContent(@Nullable Project project, @Nullable VirtualFile file, @NotNull CharSequence content) {
```
dlsniper: Why remove this from the Go files? They should always be UTF-8 per language specs (or do I miss something?)

dlsniper:
One final comment from my understanding of how the project is structured would be to convert this to a gradle module, like `google-app-engine` for example. @stuartcarnie might be able to help you out on this one.

Otherwise, thank you for this, it's really nice to have it! I'll let @ignatov and @zolotov properly review this!

Another thing to have in mind would be the `html/package` package which is basically `text/template` but for html usage.

jhsx:
@dlsniper Sorry for the mess, i will clean up the tree and push it again, some of those files i don't remember to editing they.
the Pr is not yet in the state for merging.

I do have few issues for discussion that i will list following.
The language name as you mentioned in your comment.
Default File extension and Icon.
Syntax Coloring in a different PR.
Should i create a another module for this ? but the generated source will be shared?

idea-flex.skeleton
```diff
@@ -0,0 +1,251 @@
+  /** initial size of the lookahead buffer */
```
ignatov: There's no need to have a skeleton file, please see `gen_lexer.sh`.

src/com/goide/gotemplate/GoTemplateLanguage.java
```diff
@@ -0,0 +1,59 @@
...
+import com.intellij.openapi.fileTypes.FileType;
+import com.intellij.openapi.fileTypes.LanguageFileType;
+import com.intellij.openapi.fileTypes.StdFileTypes;
+import com.intellij.psi.templateLanguages.TemplateLanguage;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Created by josesantos on 21/12/15.
+ */
```
ignatov: Please remove such comment with author according the code style.

src/com/goide/lexer/_Go_templateLexer.java~
```diff
@@ -0,0 +1,728 @@
+/* The following code was generated by JFlex 1.4.3 on 24-12-2015 16:47 */
```
ignatov: Do we actually need this?

src/com/goide/gotemplate/parser/GoTemplateParserUtil.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package com.goide.gotemplate.parser;
+
+import com.intellij.lang.parser.GeneratedParserUtilBase;
+
+/**
+ * Created by josesantos on 21/12/15.
+ */
+public class GoTemplateParserUtil extends GeneratedParserUtilBase {
```
ignatov: It's possible to use a basic `com.intellij.lang.parser.GeneratedParserUtilBase` as a parser util class in the bnf file.

ignatov:
> but the generated source will be shared?
Yes, it should be shared.

jhsx:
Sorry close by mistake

PR #2171, opened 2015-12-24
Improve declared but not used inspection (fixes #2170)

### LINKED ISSUE #2170: (chan int)(iAreaId) evaluated but not used
```go
func foo(c interface{}) {
	iAreaId := c.(chan int)
	chan int(iAreaId) // should be (chan int)(iAreaId) evaluated but not used
}
```

dlsniper:


dlsniper:
Should I move this to it's own inspection? Does it start to become heavy for the annotator to handle?

PR #2182, opened 2015-12-27
Fix evaluated but not used (fixes #2176)

### LINKED ISSUE #2176: Inspection for unused numeric values
Using nightly build of the plugin: `0.10.951` , using IDEA Community Edition `15.0.2`. 

Consider the following code:
```go
s.indices[i+1] = uint16(j)
```

The `1` part, results in the inspection error: `Numeric value evaluated but not used`. This is non-sense. This bug did not exist a few days ago (in `0.10.935`). 

Also the case for the numbers `800` and `16`:
```go
if space.Position.X > (800 - 16) {
```

---------

But not for `400` and `16` in an assignment:
```go
space.Position.X = 400 - 16
```
And not for `0` when there's no math operator involved:
```go
if space.Position.Y < 0 {
```

dlsniper:


dlsniper:
With this patch (but I've seen it without it as well) I get the out of memory exception I've mentioned on Gitter. The logs and the project I've used the scan with `Annotator` on are here: https://drive.google.com/folderview?id=0B9dwUHiIo-0PSDJ3bGM0Wl9NNDA&usp=sharing (it's basically a Kubernetes snapshot with running `Annotator` inspection by name).

src/com/goide/highlighting/GoAnnotator.java
```diff
@@ -130,7 +130,7 @@ else if (element instanceof GoConversionExpr) {
   private static void checkLiteralEvaluatedButNotUsed(@NotNull AnnotationHolder holder, @NotNull PsiElement e, @NotNull String message) {
     PsiElement parent = e.getParent();
     if (e instanceof GoFunctionLit && parent instanceof GoCallExpr) return;
-    if ((e instanceof GoStringLiteral || e instanceof GoLiteral) && (parent instanceof GoIndexOrSliceExpr || parent instanceof GoConditionalExpr)) return;
+    if ((e instanceof GoStringLiteral || e instanceof GoLiteral) && PsiTreeUtil.getParentOfType(e, GoIndexOrSliceExpr.class, GoConditionalExpr.class) != null) return;
```
ignatov: Please add a test

ignatov:
I've just reverted all your changes about this inspection. Please find a spec or a source code in go compiler. 
Relates to #2176

dlsniper:
Ok, I'll redo the PR to better handle this and I'll try to find the Go source for all the cases where this happens. I'll close this for now.

PR #2192, opened 2015-12-29
Added folding for switch, select and case statements

dlsniper:


testData/folding/switchSelectCaseStatement.go
```diff
@@ -0,0 +1,41 @@
+package foo
+
+func main() <fold text='{...}'>{
+	for <fold text='{...}'>{
```
zolotov: why do you need this here?

testData/folding/switchSelectCaseStatement.go
```diff
@@ -0,0 +1,41 @@
...
+
+	switch true <fold text='{...}'>{
+	case 1 == 2:
+		println("demo")
+
+	case 2 == 3:<fold text='{...}'>
+		println("demo")
+		println("demo")</fold>
+
+	default:<fold text='{...}'>
```
zolotov: bad placeholder text

testData/folding/switchSelectCaseStatement.go
```diff
@@ -0,0 +1,41 @@
...
+		println("demo")
+		println("demo")
+		println("demo")</fold>
+	}</fold>
+
+	switch true {
```
zolotov: why here is no any folding?

testData/folding/switchSelectCaseStatement.go
dlsniper: I felt this is more appropriate than `...` as in Go it's optional to use the curly braces for the case / default statements. But can change it if it makes more sense.

testData/folding/switchSelectCaseStatement.go
dlsniper: It's because of the clause here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/blob/16bfaadc38887de75d7d6c102271abef1b993470/src/com/goide/editor/GoFoldingBuilder.java#L45 I could extrat the size limit as a parameter if you think it makes more sense.

testData/folding/switchSelectCaseStatement.go
dlsniper: Bad copy-paste, removed, thanks.

testData/folding/switchSelectCaseStatement.go
zolotov: I think this shouldn't work for {} blocks like GoSwitchStatement

testData/folding/switchSelectCaseStatement.go
zolotov: Then it should be `{...}` when braces are present and `...` otherwise

dlsniper:
Updated according to comments. Thank you.

src/com/goide/editor/GoFoldingBuilder.java
```diff
@@ -130,22 +130,60 @@ protected void buildLanguageFoldRegions(@NotNull final List<FoldingDescriptor> r
...
+                                    switchStatement.getExprCaseClauseList().size(), 1);
+      if (range != null) {
+        result.add(new FoldingDescriptor(switchStatement, range));
+      }
+    }
+
+    for (GoCommClause commClause : PsiTreeUtil.findChildrenOfType(file, GoCommClause.class)) {
+      if (commClause.getColon() == null) continue;
+      TextRange range = processList(commClause.getColon().getNextSibling(),
+                                    ContainerUtil.getLastItem(commClause.getStatementList()),
```
zolotov: I asked not to use this logic only for `{...}` blocks

`getNextSiblign` looks not good here

src/com/goide/editor/GoFoldingBuilder.java
```diff
@@ -41,8 +41,8 @@
 
 public class GoFoldingBuilder extends CustomFoldingBuilder implements DumbAware {
   @Nullable
-  private static TextRange processList(@Nullable PsiElement left, @Nullable PsiElement right, int size) {
-    if (left == null || right == null || size < 2) {
+  private static TextRange processList(@Nullable PsiElement left, @Nullable PsiElement right, int size, int minSize) {
```
zolotov: You have only two cases here: always fold something or not to fold something if size > 2. Extracting minSize looks redundant.

zolotov:
merged manually

Issue #2210, opened 2016-01-04
Plugin is not compatible with IDEA EAP 16


obourgain:
Is there a way I can help on this topic ?

ignatov:
You can provide a patch for Idea 16. Right now we have a bunch of compilation errors in the debugger module: https://teamcity.jetbrains.com/viewLog.html?buildId=664247&tab=buildResultsDiv&buildTypeId=IntellijIdeaPlugins_Go_TestsOnIdea144Snapshot

Please set up ideaVersion = 144.2608.2 in gradle.properties.

dlsniper:
I've pushed https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2211 I'm testing it right now to see how things work and if there are any issues with the tests.

obourgain:
I also got a PR, #2212 , it looks a lot similar to yours but I also fixed a compilation issue in GoHighlightingTest: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2212/files#diff-544004967bdfc48d2a248a107830982bR298

dlsniper:
Right, I've ended up doing the same changes but still delve doesn't work, so more investigation is needed there.

zolotov:
Compatible with IDEA 16 EAP plugin had been uploaded to the nightly channel.

PR #2212, opened 2016-01-04
Fix compilation errors for with Intellij 16 EAP

obourgain:
This PR fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2210
The code changes will probably break compatibility with previous versions of IJ.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -15,7 +15,7 @@
 #
 
 version = 0.10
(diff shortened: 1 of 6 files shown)


zolotov:
Thanks! Your changes were merged manually into `master` (partially) and `144` branch.

PR #2225, opened 2016-01-12
Run Gotype command on folder instead of specific file

aagahi:
Gotype on specific file is quiet useless. Instead I changed the code to perform the gotype on the directory (an in fact I presume that was the expected behaviour as the gotype command is invoked with -a parameter).
I've successfully tested this change on IJ 15.0.2

### DIFF
--- a/src/com/goide/dlv/DlvDebugProcess.java
+++ b/src/com/goide/dlv/DlvDebugProcess.java
@@ -221,7 +221,7 @@ public final class DlvDebugProcess extends DebugProcessImpl<RemoteVmConnection> 
 
   @NotNull
   @Override
-  public List<Location> getLocationsForBreakpoint(@NotNull XLineBreakpoint<?> breakpoint, boolean b) {
+  public List<Location> getLocationsForBreakpoint(@NotNull XLineBreakpoint<?> breakpoint) {
     return Collections.emptyList();
   }
 


zolotov:
What's the point? As far as I know you can run gotype on directory without this pull request

aagahi:
I see...
When you perform gotype while the current file is edited with a kb shortcut (and btw I use a macro) it is useless... Or it would be useful to be able to choose between file or directory.

zolotov:
Then I believe it's better to provide one more action for running gotype on parent directory.

dlsniper:
Just as a thought about this: gotype will only work if you have the package "installed" via go install, which might not be the case if you use the IDE workflows only and create new packages for your project (see https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/1991)

aagahi:
Correct you also have to do go install in order to install all related package (but that could be manageable).
Yeah you could have 2 actions. Meanwhile I believe the most common use case is gotype on directory as gotype on single file is only useful for 1 file project (and I don't the point of using IJ in such case ;) ).

zolotov:
Ok. Then please rename action according to this PR

src/com/goide/actions/tool/GoTypeFileAction.java
```diff
@@ -38,6 +40,11 @@ protected boolean isAvailableOnFile(VirtualFile file) {
   protected GoExecutor createExecutor(@NotNull Project project, @Nullable Module module, @NotNull String title, @NotNull String filePath) {
     VirtualFile executable = getExecutable(project, module);
     assert executable != null;
-    return GoExecutor.in(project, module).withExePath(executable.getPath()).withParameters("-e", "-a", "-v", filePath).showOutputOnError();
+    File file = new File(filePath);
```
zolotov: Here is convertation virtualFile -> filePath -> File that looks lame. Please extract another createExecutor method in base class with VirtualFile parameter instead of filepath

aagahi:
Ok

aagahi:
Hope it would match your requirement - anyhow I'll be happy to adjust accordingly

zolotov:
Looks good but I expected that you didn't change `GoImportsFileAction` and `GoFmtFileAction`. Could you please revert those files and keep the old `createExecutor` method and introduce the new one:

```
GoExecutor createExecutor(... on virtual file ...) {
  String filePath = virtualFile.getCanonicalPath();
  assert filePath != null;
  createExecturo(... on file path ....)
}
```

aagahi:
Ok

zolotov:
closed?

PR #2226, opened 2016-01-12
Update gotype external tool and refactor GoExternalToolsAction create…

aagahi:
Following up with #2225 with more simple changeset.


zolotov:
Thanks! Will be available in the next nightly

aagahi:
You're welcome. Glad to help.

Issue #1995, opened 2015-11-12
Support benchmark running


zolotov:

sjamesr:
I'm going to have a go at this by adding a new framework type (there will be gotest, gocheck and gobench). Please stop me if you don't think this is the right approach :)

zolotov:
@sjamesr I guess it's ok

sjamesr:
I think this is good to go now, the output looks really nice unless the benchmark emits messages that confuse the output processor.

PR #2254, opened 2016-01-16
add go bench framework

sjamesr:
Fixes https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/1995.

Please take a look at the event converter class. It's not complete but I don't want to keep working on it if there's an easier way. The main reason I can't use `processServiceMessage` is that the go test -bench output is not line-delimited.

### DIFF
--- a/src/com/goide/runconfig/testing/GoTestRunConfigurationProducerBase.java
+++ b/src/com/goide/runconfig/testing/GoTestRunConfigurationProducerBase.java
@@ -94,8 +94,6 @@ public abstract class GoTestRunConfigurationProducerBase extends RunConfiguratio
         }
         else {
           GoFunctionOrMethodDeclaration function = findTestFunctionInContext(contextElement);
-          if (shouldSkipContext(function)) return false;
-
           if (function != null) {
             configuration.setName(getFunctionConfigurationName(function, getFileConfigurationName(file.getName())));
             configuration.setPattern("^" + function.getName() + "$");
@@ -115,8 +113,6 @@ public abstract class GoTestRunConfigurationProducerBase extends RunConfiguratio
 
     return false;
   }
-
-  protected abstract boolean shouldSkipContext(@Nullable GoFunctionOrMethodDeclaration context);
 
   @NotNull
   protected String getFileConfigurationName(@NotNull String fileName) {
@@ -167,7 +163,7 @@ public abstract class GoTestRunConfigurationProducerBase extends RunConfiguratio
   }
 
   @Nullable
-  private static GoFunctionOrMethodDeclaration findTestFunctionInContext(@NotNull PsiElement contextElement) {
+  protected GoFunctionOrMethodDeclaration findTestFunctionInContext(@NotNull PsiElement contextElement) {
     GoFunctionOrMethodDeclaration function = PsiTreeUtil.getNonStrictParentOfType(contextElement, GoFunctionOrMethodDeclaration.class);
     return function != null && GoTestFinder.isTestOrExampleFunction(function) ? function : null;
   }
--- a/src/com/goide/runconfig/testing/frameworks/gobench/GobenchEventsConverter.java
+++ b/src/com/goide/runconfig/testing/frameworks/gobench/GobenchEventsConverter.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2013-2016 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.runconfig.testing.frameworks.gobench;
+
+import com.goide.runconfig.testing.GoTestLocationProvider;
+import com.google.common.base.Stopwatch;
+import com.intellij.execution.testframework.TestConsoleProperties;
+import com.intellij.execution.testframework.sm.runner.OutputToGeneralTestEventsConverter;
+import com.intellij.execution.testframework.sm.runner.events.TestFailedEvent;
+import com.intellij.execution.testframework.sm.runner.events.TestFinishedEvent;
+import com.intellij.execution.testframework.sm.runner.events.TestOutputEvent;
+import com.intellij.execution.testframework.sm.runner.events.TestStartedEvent;
+import com.intellij.openapi.util.Key;
+import com.intellij.openapi.util.text.StringUtil;
+import jetbrains.buildServer.messages.serviceMessages.TestFailed;
+import org.jetbrains.annotations.NotNull;
+
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+public class GobenchEventsConverter extends OutputToGeneralTestEventsConverter {
+  private static final Pattern RUN = Pattern.compile("^(Benchmark[^ (\n\t\r]+)");
+  private static final Pattern FAIL = Pattern.compile("^--- FAIL: (Benchmark[^ (\n\t\r]+)");
+  private static final Pattern OK = Pattern.compile("^ok  \t");
+
+  public GobenchEventsConverter(TestConsoleProperties properties) {
+    super(GobenchFramework.NAME, properties);
+  }
+  private Stopwatch testStopwatch = Stopwatch.createUnstarted();
+  private String currentBenchmark = "<benchmark>";
+  private boolean benchmarkFailing = false;
+
+  @Override
+  public void process(String text, Key outputType) {
+    Matcher matcher;
+
+    if (OK.matcher(text).find()) {
+      maybeFinishCurrentTest();
+    } else if ((matcher = RUN.matcher(text)).find()) {
+      maybeFinishCurrentTest();
+      testStopwatch.start();
+      currentBenchmark = StringUtil.notNullize(matcher.group(1), "<benchmark>");
+      TestStartedEvent event = new TestStartedEvent(currentBenchmark, testUrl(currentBenchmark));
+      getProcessor().onTestStarted(event);
+    } else if ((matcher = FAIL.matcher(text)).find()) {
+      currentBenchmark = StringUtil.notNullize(matcher.group(1), "<benchmark>");
+      benchmarkFailing = true;
+    }
+
+    super.process(text, outputType);
+  }
+
+  private void maybeFinishCurrentTest() {
+    if (testStopwatch.isRunning()) {
+      testStopwatch.stop();
+      if (benchmarkFailing) {
+        TestFailedEvent event = new TestFailedEvent(currentBenchmark, "", null, true, null, null);
+        getProcessor().onTestFailure(event);
+        benchmarkFailing = false;
+      }
+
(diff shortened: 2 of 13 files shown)


sjamesr:
Here's an example screenshot. We have a custom test main function that prints a histogram. I've updated this change so that the event converter does its best to capture test output.

bench

It's easy to construct test output that will put the output converter in a bad state, but this is a problem beyond the scope of the plugin for now.

dlsniper:
This might be a _very_ stupid idea but if we know which benchmarks will be executed, could we workaround possible out issues and run them one at a time?

sjamesr:
It seems difficult to implement for the case where the user wants to benchmark a package or directory. I don't think it's worth hacking around either, better would be to wait until Go has test output that's better for tooling (e.g. https://github.com/golang/go/issues/2981). I also think the same problems with this implementation also affect unit test support in the plugin.

src/com/goide/runconfig/testing/GoTestRunConfigurationProducerBase.java
```diff
@@ -167,7 +167,7 @@ public boolean isConfigurationFromContext(@NotNull GoTestRunConfiguration config
   }
 
   @Nullable
-  private static GoFunctionOrMethodDeclaration findTestFunctionInContext(@NotNull PsiElement contextElement) {
+  protected GoFunctionOrMethodDeclaration findTestFunctionInContext(@NotNull PsiElement contextElement) {
```
zolotov: in this case `shouldSkipContext()` method is no needed. Please consider removing it

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
```diff
@@ -0,0 +1,53 @@
...
+import java.util.Collection;
+
+public class GobenchRunningState extends GoTestRunningState {
+  public GobenchRunningState(ExecutionEnvironment env, Module module, GoTestRunConfiguration configuration) {
+    super(env, module, configuration);
+  }
+
+  @NotNull
+  @Override
+  protected String buildFilePattern(GoFile file) {
```
zolotov: Please rename base method, it's supposed to be `buildFilterPattern`

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
```diff
@@ -0,0 +1,53 @@
...
+    for (GoFunctionDeclaration function : file.getFunctions()) {
+      ContainerUtil.addIfNotNull(benchmarkNames, GoTestFinder.isBenchmarkFunction(function) ? function.getName() : null);
+    }
+    return "^" + StringUtil.join(benchmarkNames, "|") + "$";
+  }
+
+  @Override
+  protected void addFilterParameter(@NotNull GoExecutor executor, String pattern) {
+    executor.withParameters("-bench", StringUtil.isEmpty(pattern) ? "." : pattern);
+    executor.withParameters("-run", "^$");
```
zolotov: Why do you add empty filter here? Does it mean that I cannot run single benchmark function?

sjamesr: The `-run ^$` ensures that no regular test methods are run along with the benchmark. One of the more experienced Go programmers in my team told me this was common practice.

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
```diff
@@ -0,0 +1,53 @@
...
+    Collection<String> benchmarkNames = ContainerUtil.newLinkedHashSet();
+    for (GoFunctionDeclaration function : file.getFunctions()) {
+      ContainerUtil.addIfNotNull(benchmarkNames, GoTestFinder.isBenchmarkFunction(function) ? function.getName() : null);
+    }
+    return "^" + StringUtil.join(benchmarkNames, "|") + "$";
+  }
+
+  @Override
+  protected void addFilterParameter(@NotNull GoExecutor executor, String pattern) {
+    executor.withParameters("-bench", StringUtil.isEmpty(pattern) ? "." : pattern);
```
zolotov: This method is supposed to patch command line with filter patterns only. Please in order not to break the semantics add separate method for adding `-bench` parameter or reuse patchExecutor method if it's possible.

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
zolotov: It will be `-run ^$` if user keeps `Pattern` field empty, otherwise user know better what he want in filter.
If benchmarks will be run on file or package, your `buildFilePatterns` gives the guarantees that there won't be any functions but benchmarks

zolotov: I just think that following scenario works incorrectly (didn't check it):

1. open file with several benchmark functions
2. click with right mouse button on single benchmark funciton name
3. click `run gobench...`
4. single benchmark should be run. As I understand – all benchmark functions will be started.

sjamesr: the `-bench` flag has the same meaning for benchmarks as `-run` does for unit tests. If the user specifies gobench framework, we want to always specify `-run ^$` and `-bench <user pattern or result of buildFilePatterns>`

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
zolotov: Or maybe I meant something like buildFilterPatternForFile, not sure :)

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunningState.java
zolotov: Ahh, I see, thanks for details. It's ok then

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunConfigurationProducer.java
```diff
@@ -0,0 +1,70 @@
...
+  @Override
+  protected boolean shouldSkipContext(@Nullable GoFunctionOrMethodDeclaration context) {
+    return context != null && context instanceof GoMethodDeclaration;
+  }
+
+  @Override
+  protected boolean setupConfigurationFromContext(@NotNull GoTestRunConfiguration configuration,
+                                                  ConfigurationContext context,
+                                                  Ref sourceElement) {
+    return super.setupConfigurationFromContext(configuration, context, sourceElement);
```
zolotov: Redundant?

src/com/goide/runconfig/testing/frameworks/gobench/GobenchEventsConverter.java
```diff
@@ -0,0 +1,87 @@
...
+  private void maybeFinishCurrentTest() {
+    if (testStopwatch.isRunning()) {
+      testStopwatch.stop();
+      if (benchmarkFailing) {
+        TestFailedEvent event = new TestFailedEvent(currentBenchmark, "", null, true, null, null);
+        getProcessor().onTestFailure(event);
+        benchmarkFailing = false;
+      }
+
+      TestFinishedEvent event = new TestFinishedEvent(currentBenchmark, testStopwatch.elapsed(TimeUnit.MILLISECONDS));
```
zolotov: Do you really need to measure time manually? I thought that `gobench` provide this information in output.

sjamesr: As far as I can tell the only information you get for sure is nanoseconds per operation, not total benchmark running time.

src/com/goide/runconfig/testing/GoTestRunConfigurationProducerBase.java
sjamesr: Quite right, done

src/com/goide/runconfig/testing/frameworks/gotest/GotestRunConfigurationProducer.java
```diff
@@ -16,19 +16,10 @@
...
-import com.goide.psi.GoFunctionOrMethodDeclaration;
-import com.goide.psi.GoMethodDeclaration;
 import com.goide.runconfig.testing.GoTestRunConfigurationProducerBase;
-import org.jetbrains.annotations.Nullable;
-
 
 public class GotestRunConfigurationProducer extends GoTestRunConfigurationProducerBase implements Cloneable {
   public GotestRunConfigurationProducer() {
     super(GotestFramework.INSTANCE);
   }
```
zolotov: This producer still should ignore GoMethodDeclarations but now it can do it overriding `findTestFunctionInContext` method

src/com/goide/runconfig/testing/frameworks/gocheck/GocheckRunConfigurationProducer.java
```diff
@@ -43,11 +43,6 @@ protected String getFunctionConfigurationName(@NotNull GoFunctionOrMethodDeclara
            : super.getFunctionConfigurationName(function, fileName);
   }
 
-  @Override
-  protected boolean shouldSkipContext(@Nullable GoFunctionOrMethodDeclaration context) {
```
zolotov: This producer still should ignore GoFunctionDeclarations but now it can do it overriding `findTestFunctionInContext` method

sjamesr: Oops, should be fixed now, please take a look

src/com/goide/runconfig/testing/frameworks/gotest/GotestRunConfigurationProducer.java
sjamesr: Oops, should be fixed now, please take a look

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunConfigurationProducer.java
```diff
@@ -0,0 +1,50 @@
...
+  @NotNull
+  @Override
+  protected String getFileConfigurationName(@NotNull String fileName) {
+    return "gobench " + super.getFileConfigurationName(fileName);
+  }
+
+  @Nullable
+  @Override
+  protected GoFunctionOrMethodDeclaration findTestFunctionInContext(@NotNull PsiElement contextElement) {
+    GoFunctionOrMethodDeclaration function = PsiTreeUtil.getNonStrictParentOfType(contextElement, GoFunctionOrMethodDeclaration.class);
```
zolotov: I think that GoMethodDeclaration is not appliable for `gobench`, so something like 

```
type T int

function (T) B<caret>enchmarkName() {
}
```

shouldn't produce `gobench` run configuration, right? If so then it's better to replace with

```
GoFunctionDeclaration function = PsiTreeUtil.getNonStrictParentOfType(contextElement, GoFunctionDeclaration.class);
```

sjamesr: You're right of course, fixed! :)

src/com/goide/runconfig/testing/frameworks/gobench/GobenchRunConfigurationProducer.java
```diff
@@ -0,0 +1,50 @@
...
+  protected String getPackageConfigurationName(@NotNull String packageName) {
+    return "gobench package '" + packageName + "'";
+  }
+
+  @NotNull
+  @Override
+  protected String getFileConfigurationName(@NotNull String fileName) {
+    return "gobench " + super.getFileConfigurationName(fileName);
+  }
+
```
zolotov: What about getFunctionConfigurationName?

sjamesr: Added, please take a look. It's nice to be able to tell the difference between running a test and a benchmark!

zolotov:
Looks good to me. Thank you. Let's try it in nightly

PR #2255, opened 2016-01-16
Improve comment start inspection

dlsniper:
The comment start inspection had some issues about it's usage and this patch attempts to fix it.

### DIFF
--- a/src/com/goide/GoDocumentationProvider.java
+++ b/src/com/goide/GoDocumentationProvider.java
@@ -81,7 +81,7 @@ public class GoDocumentationProvider extends AbstractDocumentationProvider {
   }
 
   @NotNull
-  private static List<PsiComment> getCommentsInner(@Nullable PsiElement element) {
+  public static List<PsiComment> getCommentsInner(@Nullable PsiElement element) {
     if (element == null) {
       return ContainerUtil.emptyList();
     }
--- a/src/com/goide/inspections/GoCommentStartInspection.java
+++ b/src/com/goide/inspections/GoCommentStartInspection.java
@@ -18,14 +18,14 @@ package com.goide.inspections;
 
 import com.goide.GoConstants;
 import com.goide.GoDocumentationProvider;
-import com.goide.psi.GoCompositeElement;
-import com.goide.psi.GoNamedElement;
-import com.goide.psi.GoPackageClause;
-import com.goide.psi.GoVisitor;
+import com.goide.psi.*;
+import com.goide.runconfig.testing.GoTestFinder;
 import com.intellij.codeInspection.LocalInspectionToolSession;
 import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.codeInspection.ProblemsHolder;
 import com.intellij.psi.PsiComment;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.util.PsiTreeUtil;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.List;
@@ -42,10 +42,10 @@ public class GoCommentStartInspection extends GoInspectionBase {
 
       @Override
       public void visitPackageClause(@NotNull GoPackageClause o) {
+        super.visitPackageClause(o);
         String packageName = o.getName();
-        if (GoConstants.MAIN.equals(packageName)) {
-          return;
-        }
+        if (GoConstants.MAIN.equals(packageName)) return;
+
         List<PsiComment> comments = GoDocumentationProvider.getCommentsForElement(o);
         String commentText = GoDocumentationProvider.getCommentText(comments, false);
         if (!comments.isEmpty() && !commentText.isEmpty() && !commentText.startsWith("Package " + packageName)) {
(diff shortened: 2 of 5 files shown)


dlsniper:
Also, I'm not sure if it's wise to have the inspection inspect every `GoCompositeElement` instead of exported elements. Would such a optimization work?  And if so, how should it be added to the `GoVisitor` if possible?

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -55,13 +55,18 @@ public void visitPackageClause(@NotNull GoPackageClause o) {
 
       @Override
       public void visitCompositeElement(@NotNull GoCompositeElement o) {
-        if (!(o instanceof GoNamedElement) || !((GoNamedElement)o).isPublic()) {
+        if (!shouldInspect(o)) {
```
ignatov: Could a one line statement.

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -55,13 +55,18 @@ public void visitPackageClause(@NotNull GoPackageClause o) {
...
           return;
         }
+
         List<PsiComment> comments = GoDocumentationProvider.getCommentsForElement(o);
         String commentText = GoDocumentationProvider.getCommentText(comments, false);
         String elementName = ((GoNamedElement)o).getName();
-        if (elementName != null && !comments.isEmpty() && !commentText.isEmpty()) {
+        if (elementName == null) {
```
ignatov: The same.

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -55,13 +55,18 @@ public void visitPackageClause(@NotNull GoPackageClause o) {
...
+          return;
+        }
+
+        if (!comments.isEmpty() && !commentText.isEmpty()) {
```
ignatov: Lots of negations. Please invert the if statement

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -70,10 +75,23 @@ else if (commentText.length() <= elementName.length() + 1) {
...
+          if (identifier != null) {
+            holder.registerProblem(identifier, String.format("'%s' should have a comment or not be exported", elementName), ProblemHighlightType.WEAK_WARNING);
+          }
+        }
       }
     };
   }
 
+  private static boolean shouldInspect(GoCompositeElement o) {
+    return (o instanceof GoNamedElement) &&
```
ignatov: A bit complicated condition.

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -773,7 +786,7 @@ public static PsiElement getBreakStatementOwner(@NotNull PsiElement breakStateme
                                                            GoSelectStatement.class, GoFunctionLit.class);
     return owner instanceof GoFunctionLit ? null : owner;
   }
-
```
ignatov: Yarrr! Please review your changes before push and provide a separate commit with cleanup if needed in the same PR.

dlsniper: Oh my, I'll clean this up immediately.

dlsniper:
I've updated this but unfortunately I can't run tests using IDEA 16 anymore (I'll have to jump IDEs again) due to https://youtrack.jetbrains.com/issue/IDEA-150731 (not related to this issue anyway).

There's one condition that I explicitly made to fail because it works in golint but not in the plugin.

```go
// This should be correct
const (
	NewConst = 1
	NewConst2 = 1
)
```

Currently the code will take the comment of the block for each of the values but it's not what I want because I don't know where the comment comes from, the declaration or the block, which means I can't flag the declaration as correctly commented if the comment is applied to the block. Any tips for this?

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -86,6 +86,19 @@ public static boolean isPanic(@NotNull GoCallExpr o) {
...
+  public static boolean isTestFunction(@NotNull GoFunctionOrMethodDeclaration o) {
+    String funcName = o.getName();
+    return funcName != null &&
+           isTestFile(o.getContainingFile()) &&
+           (funcName.startsWith(GoConstants.TEST_PREFIX) ||
+            funcName.startsWith(GoConstants.BENCHMARK_PREFIX) ||
+            funcName.startsWith(GoConstants.EXAMPLE_PREFIX));
+  }
+
+  public static boolean isTestFile(GoFile file) {
```
zolotov: GoTestFinder#isTestFile

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -86,6 +86,19 @@ public static boolean isPanic(@NotNull GoCallExpr o) {
     return false;
   }
 
+  public static boolean isTestFunction(@NotNull GoFunctionOrMethodDeclaration o) {
```
zolotov: unused?

src/com/goide/psi/impl/GoPsiImplUtil.java
dlsniper: I'm thinking of moving those here, would it be ok?

zolotov: Why? There are all test-related util methods

dlsniper:
Updated. I've left the failing test still there. Thank you.

testData/highlighting/commentStart.go
```diff
@@ -72,11 +75,11 @@ func Demo6() {}
 // Deprecated: use other thing
 func Demo7() {}
 
-// Demo8 does things  -> correct
+// Demo8 does things  -> incorrect
 //
 // Deprecated: use other thing
 
-func Demo8() {}
+func <weak_warning descr="'Demo8' should have a comment or not be exported">Demo8</weak_warning>() {}
```
zolotov: Why is this incorrect?

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -55,32 +55,45 @@ public void visitPackageClause(@NotNull GoPackageClause o) {
...
+        if (elementName == null) return;
+
+        if (comments.isEmpty() || commentText.isEmpty()) {
+          PsiElement identifier = ((GoNamedElement)o).getIdentifier();
+          if (identifier != null) {
+            holder.registerProblem(identifier, String.format("'%s' should have a comment or not be exported", elementName), ProblemHighlightType.WEAK_WARNING);
           }
         }
+        
+        if (!isCorrectComment(commentText, elementName)) {
```
zolotov: else if

dlsniper:
Test file without the IDEA markup:

```go
// demo
package commentstart

type (
	// help
	Hello struct{}

	// help
	Helloi interface{}
)

// help
type Helllo struct{}

// help
type Hellloi interface{}

const (
	// help
	Helloc = 3
)

// help
const Hellloc, hellloc2 = 1, 2

var (
	// help
	Hello1, Hellow1 int
)

// help
var Hello2 int

// Helllo
func (a Helllo) Hellllo() {
	_ = Hello1
	_ = Hellow1
	_ = Hello2
	_ = Helloc
	_ = Hellloc
	_ = hellloc2
}

// Demo does things  -> correct
func Demo() {
	Demo2()
}

// Demo does other things ->  incorrect
func Demo2() {
	Demo()
	Demo3()
	Demo4()
	Demo5()
	Demo6()
	Demo7()
	Demo8()
	Demo9()
}

// Demo3
func Demo3() {}

// A Demo4 does things  -> correct
func Demo4() {}

// An Demo5 does things  -> correct
func Demo5() {}

// The Demo6 does things  -> correct
func Demo6() {}

// Demo7 does things  -> correct
//
// Deprecated: use other thing
func Demo7() {}

// Demo8 does things  -> incorrect
//
// Deprecated: use other thing

func Demo8() {}

// Demo does things  -> correct
//
// Deprecated: use other thing

// Demo9 demo
func Demo9() {}

var (
	A int
	// demo comment
	C int
)

const (
	DemoDemo = 0

	DemoEmo = 1
)

// This should be correct
const (
	NewConst  = 1
	NewConst2 = 1
)

type (
	Deeeeo struct {
	}
)

// digi
func Gigel() {
	// demo
	B := "Hello"
	_, _, _, _, _ = DemoDemo, DemoEmo, A, B, C
}

func Gigel2() {
	_, _ = NewConst, NewConst2
}

func (d Deeeeo) Gigel() {
	Gigel()
	Gigel2()
}
```

@zolotov regarding: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2255#discussion_r50631822: This is how godoc / golint works currently. And the plugin currently also won't register that comment, which again, is correct.

I've also refactored this so that everything works as expected now. Should I open a `housekeeping` issue to improve handling of documentation to match what godoc does and cleanup this implementation?

testData/highlighting/commentStart.go
zolotov: The question is still actual

src/com/goide/inspections/GoCommentStartInspection.java
```diff
@@ -55,35 +55,80 @@ public void visitPackageClause(@NotNull GoPackageClause o) {
...
       public void visitCompositeElement(@NotNull GoCompositeElement o) {
-        if (!(o instanceof GoNamedElement) || !((GoNamedElement)o).isPublic()) {
-          return;
+        super.visitCompositeElement(o);
+        if (!shouldInspect(o)) return;
+
+        boolean isBlockComment = false;
+        int elementsInBlock = 0;
+        PsiElement specElement = o;
+        if (o instanceof GoVarDefinition || o instanceof GoConstDefinition) {
```
zolotov: What about GoTypeDeclaration?

src/com/goide/GoDocumentationProvider.java
```diff
@@ -85,7 +85,7 @@ else if (element instanceof GoTypeSpec) {
   }
 
   @NotNull
-  private static List<PsiComment> getCommentsInner(@Nullable PsiElement element) {
+  public static List<PsiComment> getCommentsInner(@Nullable PsiElement element) {
```
zolotov: It even contains `inner` word in name, it means that nobody should build its logic depending on this method

dlsniper: I didn't wanted to duplicate the code in the inspection but I don't know how else to prevent getting the comment from the block and not the individual declaration with using the other method. Should I add a flag to that not to bubble up?

testData/highlighting/commentStart.go
dlsniper: The answer to this is buried within the Go parser itself, I've yet to figure out where is the exact part that handles the documentation.

dlsniper: Please observe the documentation that godoc.org provides for the function: https://godoc.org/github.com/dlsniper/u/demo/demo2255#Demo8

src/com/goide/inspections/GoCommentStartInspection.java
zolotov: Still actual

dlsniper: For GoTypeDeclaration we retrieve the comments correctly, whereas for those two, unless we do that change, can't have this handled correctly:
```go
var (
	A int
	// demo comment -> this comment is not correctly handled
	C int
)

type (
	Deeeeo struct {
	}
	// demo stuff -> this comment is handled correctly
	Ehlo struct{
	}
)
```

dlsniper:
By the discussion we have in this PR it feels to me that I should rather rework how we process (get) the documentation first then refactor this. What do you think?

PR #2256, opened 2016-01-16
Read-write access detector (fixes #1332)

### LINKED ISSUE #1332: Change usage from unidentified to proper usage types in the 'Find usages' functionality
Right now, if you do a find usages or a variable, function or anything else, things work but there's no separation between assignment and reading, definition and usage for example.

dlsniper:


dlsniper:
There are two "issues" I have with the current implementation:

- I can't find a sample to do the tests for this
- function calls are marked as `Read` instead of a function call or similar (I haven't found a way to make this possible).

ignatov:
Thanks, merged manually.
> function calls are marked as Read instead of a function call or similar (I haven't found a way to make this possible).
Please clarify.

ignatov:
> I can't find a sample to do the tests for this

The same thing – I can't find any test for such functionality in the whole platform.
@zolotov could you help with it?

dlsniper:
>> function calls are marked as Read instead of a function call or similar (I haven't found a way to make this possible).

> Please clarify.

I was expecting to be able to mark function calls as being function calls not reads. But I've looked in the platform and usually these are marked as "Unclassified usage" so I guess this would have to do it for now. Thank you.

PR #2281, opened 2016-01-21
Improve coverage support for packages (fixes #2276)

### LINKED ISSUE #2276: Run 'Test all' with coverage should not attempt to include sub-packages
When I use the context option: "Run 'All in foo' with coverage", where foo is a package, the command executed is:

```/go test -v ././... -coverprofile=/[...]$All_in__foo_.coverage -covermode=atomic```

When the foo package has sub-packages, this always fails with the error ```cannot use test profile flag with multiple packages```. If the command was instead:

```/go test -v . -coverprofile=/[...]$All_in__foo_.coverage -covermode=atomic```

It would work.

### PR DESCRIPTION
**Pull request by dlsniper**
Technically only directory run configs have this issue (package running doesn't support specifying `...` yet, I'll do a different PR for that if it's deemed of interested, so far no such request tho).

### DIFF
--- a/src/com/goide/runconfig/testing/GoTestRunningState.java
+++ b/src/com/goide/runconfig/testing/GoTestRunningState.java
@@ -40,6 +40,7 @@ import com.intellij.openapi.vfs.LocalFileSystem;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.psi.PsiFile;
 import com.intellij.psi.PsiManager;
+import com.intellij.util.PathUtil;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -83,11 +84,14 @@ public class GoTestRunningState extends GoRunningState<GoTestRunConfiguration> {
         String relativePath = FileUtil.getRelativePath(myConfiguration.getWorkingDirectory(),
                                                        myConfiguration.getDirectoryPath(),
                                                        File.separatorChar);
+        // TODO Once Go gets support for covering multiple packages the ternary condition should be reverted
+        // See https://golang.org/issues/6909
+        String pathSuffix = (myCoverageFilePath == null) ? "..." : ".";
         if (relativePath != null) {
-          executor.withParameters("./" + relativePath + "/...");
+          executor.withParameters("./" + relativePath + "/" + pathSuffix);
         }
         else {
-          executor.withParameters("./...");
+          executor.withParameters("./" + pathSuffix);
           executor.withWorkDirectory(myConfiguration.getDirectoryPath());
         }
         addFilterParameter(executor, myConfiguration.getPattern());

### CODE REVIEW

File: src/com/goide/runconfig/testing/GoTestRunningState.java [hunk 1]
```diff
@@ -83,7 +83,10 @@ protected GoExecutor patchExecutor(@NotNull GoExecutor executor) throws Executio
         String relativePath = FileUtil.getRelativePath(myConfiguration.getWorkingDirectory(),
                                                        myConfiguration.getDirectoryPath(),
                                                        File.separatorChar);
-        if (relativePath != null) {
+        if (relativePath == "." && myCoverageFilePath != null) {
```
zolotov: do not compare objects with `==`

zolotov: write a todo to remove this as soon as multiple packages coverage will be supported with the link to corresponding issue

zolotov: Actually it is quite bad and inconsistent solution, it won't work good if testing directory is not equal to working directory

dlsniper:
Updated. Thank you.

zolotov:
> Actually it is quite bad and inconsistent solution, it won't work good if testing directory is not equal to working directory

Still actual

dlsniper:
I've totally missed that notification, sorry.

dlsniper:
Updated and improved (hopefully).

zolotov:
Looks good, thanks. But now it won't work if testing directory is outside of working directory :)

dlsniper:
> But now it won't work if testing directory is outside of working directory :)

I've tested with the working directory being a parent and even a different leaf of the same parent and they worked both (second one needed "convincing" but still worked). Can you please provide a sample of what's not working? Thank you.

zolotov:
@dlsniper empty dir path, empty working dir path, directories on different drive – anything that lead to null relative path leads to incorrect work of your patch

dlsniper:
Updated.

src/com/goide/runconfig/testing/GoTestRunningState.java
```diff
@@ -83,13 +84,17 @@ protected GoExecutor patchExecutor(@NotNull GoExecutor executor) throws Executio
...
-          executor.withParameters("./" + relativePath + "/...");
+          path += relativePath;
         }
         else {
-          executor.withParameters("./...");
           executor.withWorkDirectory(myConfiguration.getDirectoryPath());
         }
+        // TODO Once Go gets support for covering multiple packages the ternary condition should be reverted
+        // See https://golang.org/issues/6909
+        if (!path.endsWith(File.separator)) path += File.separator;
```
zolotov: Now flow is overcomplicated. Path can end with separator only if relativePath is not null and appending it with File.separator makes sense only if it's not coverage run. As a result you make 3 ifs instead 2.

The only thing you change is path suffix ("..." or "."), you can extract it in variable and append it to executor for both cases (with null relativePath and notnull relativePath)

dlsniper:
Done. It does indeed look much better.

Issue #2137, opened 2015-12-23
Quickfix for adding field names to struct initialization


dlsniper:
After https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2125, it would be nice to have a quickfix that adds the field names to the struct. When the number of values doesn't match the number of fields, then a list of possible fields should be displayed for each field.

PR #2289, opened 2016-01-23
Improve unnamed field initialization inspection (fixes #2267)

### LINKED ISSUE #2267: Unnamed field initialization inspection error in array
```go
package main

type Item struct {
    Name   string
}

type Menu []Item

func main() {
    _ = Menu{
        {Name: "home"}, // flagged as unnamed field initialization error but it's correct
    }
}
```

dlsniper:
fixes #2267

### DIFF
--- a/src/com/goide/inspections/GoStructInitializationInspection.java
+++ b/src/com/goide/inspections/GoStructInitializationInspection.java
@@ -17,13 +17,22 @@
 package com.goide.inspections;
 
 import com.goide.psi.*;
+import com.goide.quickfix.GoReplaceWithNamedStructFieldQuickFix;
 import com.intellij.codeInspection.LocalInspectionToolSession;
 import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.codeInspection.ui.SingleCheckboxOptionsPanel;
 import com.intellij.psi.util.PsiTreeUtil;
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.*;
+import java.util.List;
+
 public class GoStructInitializationInspection extends GoInspectionBase {
+
+  @SuppressWarnings("WeakerAccess")
+  public boolean reportImportedStructs;
+
   @NotNull
   @Override
   protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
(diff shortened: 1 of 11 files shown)


src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -36,8 +36,11 @@ public void visitCompositeLit(@NotNull GoCompositeLit o) {
         if (PsiTreeUtil.getParentOfType(o, GoReturnStatement.class, GoShortVarDeclaration.class, GoAssignmentStatement.class) == null) return;
 
         GoType type = o.getType();
-        if (!(type instanceof GoStructType) &&
-            !(GoPsiImplUtil.findBaseTypeFromRef(o.getTypeReferenceExpression()) instanceof GoSpecType)) {
+        GoType typeFromTypeRef = GoPsiImplUtil.findTypeFromTypeRef(o.getTypeReferenceExpression());
+        if (!(type instanceof GoStructType ||
```
zolotov: why just replacing && with || is not enough?

dlsniper: I'll commit a version with just that change. I don't remember why I've done this and it's too complicated to run the tests right now (I'm running on IDEA 16 EAP)

dlsniper: I think the find method will bubble up too much and resolve to the type ` string ` instead of the type ` Item `. Maybe the change should be done in that function https://teamcity.jetbrains.com/viewLog.html?buildId=693727&buildTypeId=IntellijIdeaPlugins_Go_Test

zolotov: I see, I restore the changes then. Are there any issues about quickfixes for this inspection? Please create one if not.

dlsniper: I'll restore the changes. As for the quickfix, we have: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2137

zolotov: Ok

dlsniper: Updated code to the original PR. Thank you.

dlsniper:
@zolotov this has to suffer some more work. Can you please point me to an inspection with options? In order to match the golint / go vet equivalent, by default the inspection should only be applied to non-local structs. However I'd like to have the option to check all structs, local and non-local as well. Thank you.

zolotov:
https://github.com/JetBrains/intellij-community/blob/daad338196544c0e603edbb4e5ef270e882e8927/plugins/InspectionGadgets/InspectionGadgetsAnalysis/src/com/siyeh/ig/inheritance/AbstractMethodOverridesAbstractMethodInspection.java#L56-L56

dlsniper:
I've refactored this and further improved the inspection for scenarios it didn't previously considered (embedded structs for example). It also has the option to be enabled for local defined types or act like the current alternative does which inspects only non-package local types (default setting).

There's a todo left to handle but I think it can be covered at a later time.

The quick fix might need to be improved but I don't know where to look for which is better between `LocalQuickFixOnPsiElement` and `LocalQuickFixBase`.

zolotov:
I'll cleanup the PR, the comments I write here is just for your information.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -32,17 +41,87 @@ protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNul
...
+
+  @Override
+  public JComponent createOptionsPanel() {
+    return new SingleCheckboxOptionsPanel("Report for local type definitions as well", this, "reportImportedStructs");
+  }
+
+  private boolean canProcess(String myPackage, @NotNull GoStructType type) {
+    if (!(type.getContainingFile() instanceof GoFile)) return false;
+    String typePackage = ((GoFile)type.getContainingFile()).getPackageName();
+    return reportImportedStructs || myPackage == typePackage;
```
zolotov: do not compare objects with `==`

zolotov: comparing package's names is not enough to say that symbols are located in the same package

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -32,17 +41,87 @@ protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNul
...
 
         GoTypeReferenceExpression ref = o.getTypeReferenceExpression();
-        if (o.getType() instanceof GoStructType || ref != null && ref.resolveType() instanceof GoSpecType) {
+        if (ref == null || ref.resolveType() == null) return;
+        GoType refType = ref.resolveType().getUnderlyingType();
+        if (refType instanceof GoStructType) {
+          if (!canProcess(myPackage, (GoStructType)refType)) return;
+          fields = ((GoStructType)refType).getFieldDeclarationList();
+          if (hasAnonymousStructField(fields)) return;
+          processLiteralValue(holder, fields, o.getLiteralValue());
```
zolotov: duplicates GoStructType processing above

zolotov: Hmm, I've replaced all 3 if-statements with the following code:

```
GoType refType = GoPsiImplUtil.getLiteralType(o);
        if (refType instanceof GoStructType) {
          processStructType((GoStructType)refType, o, holder);
        }
```

and all tests are passed

zolotov:
merged manually

PR #2290, opened 2016-01-24
Introduce underscore as value inspection (fixes #2257)

### LINKED ISSUE #2257: Add a simple inspection: cannot use _ as value
```go
package main

func main() {
	println(_, "hello") // cannot use _ as value
	a, b := 1, _  // cannot use _ as value
	_, _ = a, b
}
```

dlsniper:


zolotov:
The solution doesn't look good to me. Inspection ignores a lot of variants expressions and checks only constants, variables and argument lists. I believe it could be simpler and cover more cases.

Some not covered cases:

```go
        c := 1 + _ + (_)
	println(_ + (_)) // cannot use _ as value
	println(c)
```

zolotov:
Also I almost cannot see error-stripe on `_`. Maybe it's better to mark it with `ERROR` instead of `GENERIC_ERROR_OR_WARNING`.

dlsniper:
I'll try rework the solution, thanks for feedback and test scenarios.

> Also I almost cannot see error-stripe on _. Maybe it's better to mark it with ERROR instead of GENERIC_ERROR_OR_WARNING.

Ok, I was used to have this always flagged as that.

dlsniper:
I've updated this, it's much better now I think.
However, there's a problem in determining on which side of the ` range ` clause the expression is.

zolotov:
Fix tests please

dlsniper:
The test that's not working, https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2290/files#diff-2626ab6331a8b9c53ed085de9efadf2dR26, is correct from Go's perspective as it is marked but the inspection is wrong now. I don't know a good method to determine if _ is used in the range or assignment, please advise. That's why I've left the test to fail.

src/com/goide/inspections/GoUnderscoreUsedAsValueInspection.java
```diff
@@ -0,0 +1,47 @@
...
+                                     @SuppressWarnings({"UnusedParameters", "For future"}) @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitReferenceExpression(@NotNull GoReferenceExpression o) {
+        super.visitReferenceExpression(o);
+        if (!o.getIdentifier().textMatches("_")) return;
+        PsiElement parent = o.getParent();
+        if (parent instanceof GoSendStatement) return;
+        if (parent instanceof GoRangeClause) {
+          if (!((GoRangeClause)parent).getExpressionList().contains(o)) return;
```
zolotov: This is always true

dlsniper: I'm not sure how to solve this for the following case:

```
for _, _ = range <error descr="Cannot use '_' as value">_</error>  {
}
```

I've tried to replace this with a call to `getVarDefinitionList` but if I do ` if (((GoRangeClause)parent).getVarDefinitionList().contains(o)) return; ` then everything will be flagged as incorrect in that snippet.

dlsniper: I've rebased the code onto the latest master.

zolotov: `getVarDefinitionList` is completely different thing, first of all it returns var definitions instead of referenceExpression so it will be always `false`. In the second it's better to look at bnf if you're not sure what will be retrieved by particular method on PSI.

I've pushed new method `GoReferenceExpression#getReadWriteAccess`, you can use it to know whether it's read or write operation.

zolotov: And fix the tests please

testData/highlighting/underscoreUsedAsValue.go
```diff
@@ -0,0 +1,28 @@
...
+	println(<error descr="Cannot use '_' as value">_</error>, "hello1")
+	a, b:= 1, <error descr="Cannot use '_' as value">_</error>
+	_, _, _, _ = a, b, demov, democ
+	_, _, _, _ = demov1, demov2, democ1, democ2
+	c := 1 + <error descr="Cannot use '_' as value">_</error> + (<error descr="Cannot use '_' as value">_</error>)
+	println(<error descr="Cannot use '_' as value">_</error> + (<error descr="Cannot use '_' as value">_</error>))
+	println(c)
+
+	select {
+	case _ <- 0:  break;
```
zolotov: should be a error

dlsniper: Excellent catch, thank you!

PR #2292, opened 2016-01-24
Improve highlighting (fixes #1534)

### LINKED ISSUE #1534: Coloring Issues

dlsniper:


dlsniper:
I'm not sure if it's safe to do resolving here but the other option would be to check the text. It would be good if there would be a `isPublic` method for `GoFieldName`.

zolotov:
I found more appropriate fix. Test is reused, thanks

PR #2293, opened 2016-01-24
Introduce support for builtin function highlighting (fixes #1841)

### LINKED ISSUE #1841: Syntax category for built-in methods
Built in methods like new(), delete() and make() etc. should have their own syntax category so they can be independently styled (since they are somewhat special, and somewhat important). There's already a category for built-in types (which inherit from the same module).

dlsniper:
This introduces support for highlighting builtin functions. However to do this, the test class had to be updated which lead to the noise in the PR.

### DIFF
--- a/src/com/goide/highlighting/GoHighlightingAnnotator.java
+++ b/src/com/goide/highlighting/GoHighlightingAnnotator.java
@@ -92,6 +92,7 @@ public class GoHighlightingAnnotator implements Annotator {
   }
 
   private static TextAttributesKey getColor(GoNamedSignatureOwner o) {
+    if (GoPsiImplUtil.builtin(o)) return BUILTIN_FUNCTION;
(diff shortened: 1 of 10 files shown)


src/com/goide/highlighting/GoHighlightingAnnotator.java
```diff
@@ -92,6 +92,7 @@ private static TextAttributesKey getColor(GoVarDefinition o) {
   }
 
   private static TextAttributesKey getColor(GoNamedSignatureOwner o) {
+    if (GoPsiImplUtil.isBuiltinFile(o.getContainingFile())) return BUILTIN_FUNCTION;
```
zolotov: GoPsiImplUtil#builtin

dlsniper: Updated, thank you.

PR #2296, opened 2016-01-24
Initial support for generating interfaces implementations (fixes #1568)

### LINKED ISSUE #1568: Support 'Generate' actions
I'd like to have a "Generate" action that enables me to implement all the methods defined in a interface on a certain struct, just like I can use it in Java to add the boilerplate code to implement a certain method / interface / abstract class.

The way it should work would be like this:
Allow you to search in the list of all existing interfaces, once you've selected which interfaces you want to use generate the method stubs for those interfaces bound to the struct and add a new "constructor" as defined below.

Ideally there should be two "Generate" actions, one to generate an implementation on a pointer struct and one on the concrete struct.

Concrete example:

```go
package awesome

type (
	Fetcher interface {
		Execute(string) string
	}

	myFetcher struct <caret> {
		myDependency string
	}
)

```

Expected result:

```go
package awesome

type (
	Fetcher interface {
		Execute(string) string
	}

	myFetcher struct {
		myDependency string
	}
)

func (mf *myFetcher) Execute(<caret> string) string {
	//TODO Add implementation
}


func NewFetcher() Fetcher {
	return &myFetcher {}
}
```

dlsniper:
This aims to provide a way to generate the mock implementations for interfaces for a struct.

Things left to do:

- [ ] allow this to run on all local types (Go allows defining methods on any local types, not only structs)
- [ ] gather all the interfaces
- [ ] provide a better way to select which interface should be implemented
- [ ] check which method is already implemented by the type before creating the implementation
 - [ ] ask to rename existing methods that don't have the same signature before continuing
- [ ] better positioning for the newly created methods
- [ ] allow better cursor positioning
- [ ] analyze existing methods on the type in check if the receiver should be a pointer or value
- [ ] tests

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -335,6 +335,12 @@
             description="Analyze types for current directory with gotype util">
       <add-to-group group-id="CodeMenu" anchor="last"/>
     </action>
+
+    <action id="implementInterface" class="com.goide.actions.GoImplementInterfaceAction"
+            text="Implement interface"
+            description="Implement interface">
+      <add-to-group group-id="GenerateGroup" anchor="last"/>
+    </action>
   </actions>
 
   <application-components>
--- a/src/com/goide/actions/GoImplementInterfaceAction.java
+++ b/src/com/goide/actions/GoImplementInterfaceAction.java
@@ -1,0 +1,130 @@
+/*
+ * Copyright 2013-2016 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


dlsniper:
I've opened this PR early on so that it doesn't become too complex for review. Please let me know how do you wish to proceed with it.

Furthermore, I would definitely need some help in some places, mainly getting all the interfaces that could be implemented and the UI for better presentation. I'm expecting that there are going to be a large number of interfaces, especially when computing them from the GOPATH.

Thank you.

dlsniper:
I'll close this for now to remove the noise from the PR pool. Will reopen when I have more time for it.

PR #2321, opened 2016-02-02
Improve the placeholder count inspection

### LINKED ISSUE #2194: cannot detect   placeholder  %+v
###### for example,  the typing following code will get inspections warning

log.Printf("a: %+v",a) ` Got 0 placeholder(s) for 1 arguments(s)`
fmt.Printf("a: %+v",a) ` Got 0 placeholder(s) for 1 arguments(s)`

### LINKED ISSUE #2306: Printf placeholders incorrect count for left-justify
Using `Printf` with left-justify formatting ("pad with spaces on the right rather than the left") I get a `Got 0 placeholder(s) for 1 arguments(s)` error. Example:

```
fmt.Printf("%-4d", 999)
```

dlsniper:
fixes #2194, fixes #2306

### DIFF
--- a/src/com/goide/inspections/GoPlaceholderCountInspection.java
+++ b/src/com/goide/inspections/GoPlaceholderCountInspection.java
@@ -35,7 +35,7 @@ import java.util.regex.Pattern;
 
 public class GoPlaceholderCountInspection extends GoInspectionBase {
   private static final Pattern PLACEHOLDER_PATTERN =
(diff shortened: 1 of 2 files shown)


dlsniper:
This will still fail for something like ` %%%v ` but I hope that pattern doesn't come often enough until I get the rework on this inspection done.

dlsniper:
You can ignore the comment above. I do agree that especially after this fix there must be a nicer way to maintain this than regular expressions.

PR #2340, opened 2016-02-11
Added initial simple tests for GoTemplate lexer and parser

dlsniper:
This adds initial simple tests to see if the format is correct.
More to follow once this is merged.


dlsniper:
I've added some tests but I've cheated and got the output from the current parser/lexer, I haven't validated if the output is actually correct or not (I still have to read about testing for these components). Thank you for helping out.

PR #2415, opened 2016-03-25
Run tests via building runner

dlsniper:


dlsniper:
This has problems when running test unfortunately so please merge with care if needed. I don't know why the output with this information:
```
=== RUN   TestDemo
--- PASS: TestDemo (0.00s)
```
doesn't get parsed / sent correctly.

dlsniper:
I might be able to make the PR smaller if I revert the changes to the GoApplicationRunner to include the debugging code again, let me have a look first.

dlsniper:
Done, it's a lot smaller now.

dlsniper:
And I've also fixed the issue with tests not working

dlsniper:
The next step will be to merge the Run File functionality into Go application, then add debugging to tests. As for the multiple file running I don't know how to do that, theoretically the files should be specified somehow to the go tool and we already have a way to send params to the tool so we'll have to see when we get there.

src/com/goide/runconfig/GoBuildingRunner.java
```diff
@@ -79,46 +85,86 @@ public boolean canRun(@NotNull String executorId, @NotNull RunProfile profile) {
...
-      .withParameterString(((GoApplicationRunningState)state).getGoBuildParams())
-      .withParameters("-o", outputFile.getAbsolutePath())
-      .withParameters(((GoApplicationRunningState)state).isDebug() ? new String[]{"-gcflags", "-N -l"} : ArrayUtil.EMPTY_STRING_ARRAY)
-      .withParameters(((GoApplicationRunningState)state).getTarget())
+
+    String goToolCommand;
+    File outputFile;
+    String target;
+
+    if (state instanceof GoApplicationRunningState) {
```
zolotov: The particular state should know how to build itself, there is no need to do these instanceofs in generic builder

dlsniper: Hi, I've tried to do what you've suggested but I have a good model of how this should be done. Can you please provide more detail? If there should be a `GoBuildingRunnerState` who should call that? And is that class inherited anywhere or initialized in `GoBuildingRunner`? Thanks.

dlsniper: Oh, I've missed the comment on this as you edited my question. We already have a `GoRunningState` which is inherited by every class. Should I add a method there which is named `build()` (or a better name)?

zolotov: No, GoRunningState is GoRunningState, it can run. GoBuildRunningState is a State which can be build before running.

dlsniper:
Too many changes to easily rebase. I'll redo this. I still don't know how to introduce the GoBuilding state and how to call it tho. I'm sorry, the design of this eludes me.

src/com/goide/runconfig/GoBuildingRunner.java
dlsniper: Hi @zolotov, I'd like to give this another go but I don't know how to structure the code.

You said that GoBuildRunningState should run before GoRunningState but how should the class be defined? And how should I call GoBuildingRunningState from GoRunningState? Or is there a a running state chain that I should add this to?

Thank you.

zolotov: > You said that GoBuildRunningState should run before GoRunningState

I didn't say that and it shouldn't. There is not any state chain or something.

dlsniper: Sorry, that was my interpretation for "GoBuildRunningState is a State which can be build before running ". Did you meant that it should be build in the code?

zolotov: > Did you meant that it should be build in the code?

What is `it`? Be build in the code? What does this mean?

I don't know how to rephrase this. I just described two entities and their differences: GoRunningState and GoBuildingRunningState

PR #2418, opened 2016-03-26
Run tests one package at a time

horkhe:
By default go tool is testing several packages in parallel. That does not play well with testing progress visualization. Besides if tests use the same test DB or something like that they can affect each other progress resulting in sporadic failures. Fortunately **one package at a time** testing can be forced with `-p 1` flag.

### DIFF
--- a/src/com/goide/runconfig/testing/GoTestRunningState.java
+++ b/src/com/goide/runconfig/testing/GoTestRunningState.java
@@ -98,6 +98,10 @@ public class GoTestRunningState extends GoRunningState<GoTestRunConfiguration> {
   @Override
   protected GoExecutor patchExecutor(@NotNull GoExecutor executor) throws ExecutionException {
     executor.withParameters("test", "-v");
+    // By default go tool is testing several packages in parallel. That does
+    // not play well with testing progress visualization. Therefore one package
+    // at a time is forced.
+    executor.withParameters("-p", "1");
     executor.withParameterString(myConfiguration.getGoToolParams());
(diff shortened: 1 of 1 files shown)


dlsniper:
Hi, thanks for the PR. I would say no to this as the argument can be supplied by the user if desired using the go tool arguments. I would rather leave the defaults Go uses as they are.

dlsniper:
As for the "does not play well with test visualization" part, can you please explain? Thanks.

horkhe:
When go run tests for several packages in parallel it buffers output until tests are completed and spews the buffered output in the end. Such behaviour is very annoying when you have long running tests.

Anyway, you are right I can specify this flag in the go tool arguments, in fact that is what I do now.

However I would really prefer having that enabled by default. So how about adding a checkbox in the Go test configuration **test one package at a time** and have it enabled by default?

zolotov:
Run configuration shouldn't add `-p` parameter implicitly, it will shadow user's `-p` parameter.

> Anyway, you are right I can specify this flag in the go tool arguments, in fact that is what I do now.

Moreover, you can specify this flag for all Go Test Run configurations by default:
screenshot 2016-03-28 13 07 52


> However I would really prefer having that enabled by default. So how about adding a checkbox in the Go test configuration test one package at a time and have it enabled by default?

Checkbox is not good too, if you check it and add `-p 2` in parameters you'll undefined behavior.

The only solution is to set `Go Tool parameters`=`-p 1` by default for all Go Test Configurations: set `myGoParams` in constructor of `GoTestRunConfiguration`

PR #2422, opened 2016-03-30
WIP: New inspection: Obtaining address of a for loop variable

deadok22:
Please, don't merge it now - it's a work in progress.

Today I've spent about an hour looking for a bug introduced by obtaining address of a for loop variable. This inspection could have helped to avoid that bug, or at the very least, it'd have helped me find it =)

Please, let me know if my vision of the problem is incomplete or if this inspection is too loud.


zolotov:
Not sure that it's necessary inspection. As far as I understand it will compain about following code:

```go
package main

import "fmt"

func doSomethingWithPointer(pointer *int) {
	fmt.Println(*pointer)
}

func main() {
	for _, i := range []int{1, 2, 3} {
		doSomethingWithPointer(&i)
	}
}
```

but in my opinion it's more or less valid case and its output looks logical to me:

```
1
2
3
```

deadok22:
Your example is legit. 

But it can be a problem if a pointer is stored somewhere. The pointer is the same on each iteration of the loop and it can be a source of hard-to-find bugs.

zolotov:
My point is that other functions might require pointer-parameters and you have to obtain address in these cases, and I think that IDEA shouldn't complain about it. `gometalinter` doesn't complain about it neither.

ignatov:
BTW, you can add it and disable by default.

deadok22:
@zolotov that's true, that's why I asked for opinions on whether this inspection is too loud.

So, yes, I can either make it disabled by default, or I can use scripting extensions to have it for myself only.

zolotov:
Disabled by default inspection looks good to me

zolotov:
@deadok22 can it be merge now? with disabling by default. If so, please rebase and disable the inspection by default.

deadok22:
No, I still have a TODO there.

PR #2467, opened 2016-04-16
Add inspection for infinite for loop (fixes #2443)

### LINKED ISSUE #2443: Add inspection for infinite loop
Instead of ` for {} ` the plugin should detect it and suggest ` select {} ` for blocking a function / goroutine.
Also this should be a warning as `for {}` will cause the CPU to spin at 100% while `select {}` will simply block the goroutine and the scheduler will be aware of that (so no performance issue).

dlsniper:


src/com/goide/inspections/GoInfiniteForInspection.java
```diff
@@ -0,0 +1,52 @@
...
+  protected GoVisitor buildGoVisitor(@NotNull final ProblemsHolder holder, @NotNull LocalInspectionToolSession session) {
+    return new GoVisitor() {
+      @Override
+      public void visitForStatement(@NotNull GoForStatement o) {
+        super.visitForStatement(o);
+
+        GoBlock forBlock = o.getBlock();
+        GoRangeClause forRangeClause = o.getRangeClause();
+        GoForClause forClause = o.getForClause();
+        if (o.getExpression() != null ||
```
zolotov: please invert the condition. this is really strange to consider all loops as infinite *except* `o.getExpression() != null` or block == nul or ... etc

zolotov: things like 

```
o.getExpression) != null || forRangeClause != null && !forRangeClause.getExpressionList().isEmpty()
```

could be extracted in separate method with sensible name. At the moment whole condition looks like a mess

src/com/goide/quickfix/GoReplaceWithSelectStatementQuickFix.java
```diff
@@ -0,0 +1,63 @@
...
+    return QUICK_FIX_NAME;
+  }
+
+  @Override
+  public void invoke(@NotNull final Project project,
+                     @NotNull PsiFile file,
+                     @Nullable("is null when called from inspection") Editor editor,
+                     @NotNull final PsiElement startElement,
+                     @NotNull PsiElement endElement) {
+    new WriteCommandAction.Simple(project, getName(), file) {
```
zolotov: WriteCommandAction#runWriteCommandAction

zolotov: startElement should be checked before, otherwise you can suddenly replace whole file with your `select` statement

zolotov: Also since you don't really use psi element that you passed to quick fix and don't use editor, it's better to use LocalQuickFixBase

dlsniper:
Updated per feedback

zolotov:
merged manually

PR #2468, opened 2016-04-16
Added custom theme

dlsniper:
I'd like to add the theme that I'm using daily since about a year now, see below. I've spoke briefly about this on Gitter, so hope it's ok.

There are some people that agree we could ship better default themes but I'm not sure how to achieve that right now.

I've looked at a few themes on http://color-themes.com/ (lovely website btw) and I have a few suggestions to add:
- white backgrounds:
 - [Relax Your Eyes](http://color-themes.com/?view=theme&id=563a1a6e80b4acf11273ae76)
 - [Classic Eclipse](http://color-themes.com/?view=theme&id=563a1a6b80b4acf11273ae67)
 - [Codding Horror](http://color-themes.com/?view=theme&id=563a1a6a80b4acf11273ae63)
- dark backgrounds:
 - [Monokai Sublime Text 3](http://color-themes.com/?view=theme&id=563a1a7680b4acf11273ae94) *
 - [Base16 Ocean](http://color-themes.com/?view=theme&id=563a1a7880b4acf11273ae9b) *
 - [Programmer](http://color-themes.com/?view=theme&id=563a1a6880b4acf11273ae5a)

I've marked with * those that some gophers in the slack channel indicated as being most used by them.

I've looked a bit into them and I don't have a full grasp on how to embed them just yet. A possible issue that I can see is that IntelliJ doesn't support color schemes that apply only to certain languages so if we ship a theme it will go out for everyone. This leads to the problem that once we ship a theme as part of the plugin then I'm not sure what will happen to existing themes with the same name and, from what I can see, in order to get it modified it will have to be saved first (which again goes against how this will work for people already using these themes?)

Thank you.

Color scheme preview:
Darcula - dlsniper

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -302,6 +302,10 @@
                      groupName="General" enabledByDefault="true" level="ERROR"
                      implementationClass="com.goide.inspections.GoUnderscoreUsedAsValueInspection"/>
     <!-- /general -->
+
+    <!-- color schemes -->
+    <bundledColorScheme path="colorscheme/Darcula - dlsniper"/>
+    <!-- /color schemes -->
   </extensions>
   <actions>
     <action id="Go.NewGoFile" class="com.goide.actions.file.GoCreateFileAction"
--- a/resources/colorscheme/Darcula - dlsniper.xml
+++ b/resources/colorscheme/Darcula - dlsniper.xml
@@ -1,0 +1,107 @@
+<scheme name="Darcula - dlsniper" version="142" parent_scheme="Darcula">
+  <option name="LINE_SPACING" value="1.0" />
+  <option name="EDITOR_FONT_SIZE" value="14" />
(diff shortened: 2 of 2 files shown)


zolotov:
Monokai Sublime Text 3 is good, let's bundle it too

dlsniper:
> Monokai Sublime Text 3 is good, let's bundle it too

Should I do anything for it? I'm a bit afraid to ship a theme that will affect other languages as well as possibly interfere with themes that users might already have installed.

Also, do you think we should ship the ` Relax Your Eyes ` them as well? It looks good for a white background theme but I'm not a user of such themes, (I like Darcula a lot more)

zolotov:
> Should I do anything for it? 

Bundle it please :) Also it would be nice if you take a look at its implementation in Sublime/Atom for golang, maybe theme from http://color-themes.com/ need to be adjusted for go a little.

> I'm a bit afraid to ship a theme that will affect other languages as well as possibly interfere with themes that users might already have installed.

It won't be default so I guess it's ok.

PR #2469, opened 2016-04-17
Rework placeholder count inspection (fixes #2188)

### LINKED ISSUE #2188: Rework the placeholder detection in the placeholder count detection
As per comment https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2187#issuecomment-167661791 the placeholder detection should be simplified / improved so that we can can get more out of this:

- simplify the process
- get the pair of placeholder / value(s) (a value can be assigned to multiple placeholders due to placeholder supporting indexes)
- check if the placeholder type is valid for the formatting function

dlsniper:


dlsniper:
Better placeholder validation has to be done before this is done as it's currently possible to have the allowed characters in any order as long as they are part of the allowed list.
My best idea right now would be a `map<char, list<char>>` of precedence so that we can check what is allowed to be in front of a specified character. It won't fix everything but should further limit the false positives.

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -55,6 +54,18 @@
     Pair.pair("logf", 0),
     Pair.pair("skipf", 0));
 
+  private static final List<String> INDEXED_PLACEHOLDER_FUNCTIONS = ContainerUtil.newArrayList(
+    "printf", "sprintf", "errorf", "fprintf", "fatalf", "logf", "skipf");
+
+  @SuppressWarnings("FieldCanBeLocal")
```
zolotov: it's a valid inspection report, do not suppress it, fix it

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -63,25 +74,68 @@ private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration
...
+    LinkedHashMap<Integer, String> placeholders = new LinkedHashMap<Integer, String>();
+    while (placeholderStart < placeholderTextLength) {
+      ProgressManager.checkCanceled();
+      if (PLACEHOLDER_START_CHAR != placeholderText.charAt(placeholderStart)) {
+        placeholderStart++;
+        continue;
+      }
+      int placeholderEnd = placeholderStart + 1;
+      while (placeholderEnd < placeholderTextLength &&
+             PLACEHOLDER_ALLOWED_CHARS.indexOf(placeholderText.charAt(placeholderEnd)) != -1 &&
```
zolotov: StringUtil#containsChar

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -63,25 +74,68 @@ private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration
...
+             PLACEHOLDER_END_CHARS.indexOf(placeholderText.charAt(placeholderEnd)) == -1) {
+        placeholderEnd++;
+      }
+      if (placeholderEnd >= placeholderTextLength) break;
+      if (PLACEHOLDER_ALLOWED_CHARS.indexOf(placeholderText.charAt(placeholderEnd)) == -1 &&
```
zolotov: StringUtil#containsChar

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -63,25 +74,68 @@ private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration
...
     String placeholderText = resolve(argument);
-    if (placeholderText == null) return -1;
+    if (placeholderText == null) return null;
+
+    int placeholderTextLength = placeholderText.length();
+    int placeholderStart = 0;
+    LinkedHashMap<Integer, String> placeholders = new LinkedHashMap<Integer, String>();
+    while (placeholderStart < placeholderTextLength) {
+      ProgressManager.checkCanceled();
+      if (PLACEHOLDER_START_CHAR != placeholderText.charAt(placeholderStart)) {
```
zolotov: this actually is `placeholderText.indexOf(PLACEHOLDER_START_CHAR, placeholderStart)`

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -63,25 +74,68 @@ private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration
...
+      placeholderStart = placeholderEnd;
+      placeholderStart++;
+    }
+
+    return placeholders;
+  }
 
-    Matcher placeholders = PLACEHOLDER_PATTERN.matcher(placeholderText);
-    Matcher countPlaceholders;
+  private static boolean isValidPlaceholder(String placeholder) {
```
zolotov: So `%%` is the only valid placeholder?

dlsniper: I've pushed this last night so that I have it. I'll continue working on this. Thanks for feedback. There's much to work on this to make it report all the things correctly (like ` fmt.Printf("%.*s %d %g\n", 3, "hi", 23, 2.3) `, which is part of the go vet tests)

zolotov: Have you seen `govet` implementation https://github.com/golang/go/blob/master/src/cmd/vet/print.go? I think it good to get as much as possible from there

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -63,25 +74,68 @@ private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration
...
     }
+
     return -1;
   }
 
-  private static int getPlaceholders(@Nullable GoExpression argument) {
-    if (argument == null) return -1;
+  @Nullable
+  private static LinkedHashMap<Integer, String> getPlaceholders(@Nullable GoExpression argument) {
+    if (argument == null) return null;
```
zolotov: Do not return `null` instead of empty collections

src/com/goide/inspections/GoPlaceholderCountInspection.java
dlsniper: Funny enough seems I was more strict than Go (`go vet`) is when it comes to parsing the placeholders.

zolotov: it's good reason to contribute in golang

dlsniper:
I'll close this. Needs a different rewrite to match the logic behind Go vet / fmt package which is less strict than the current logic.

PR #2477, opened 2016-04-19
Added two more color themes

dlsniper:
I've added the other popular dark background color scheme.
Also, I've added a new theme with white background.

For the copyright notice I've added something to the README of the project, hope it's enough.


zolotov:
Looks like they don't have any go-specific styles. Don't the same schemes for sublime/atom/etc have some go-specific stuff too?

dlsniper:
No, it seems there's no control over how the theme looks like for individual languages in the other editors. I'd have to dig in the API / source files to see if they (can) change something for this.

PR #2489, opened 2016-04-25
Upgrade to delve API v2

dlsniper:
This PR will migrate the plugin to API v2 of Delve.
It also adds the usage of goroutineId to make it consistent (as a side effect of doing the work on this).
Almost everything works with the new API. However it seems that there are some problems / breaking changes in the API with regards to string values and structs, see: https://github.com/derekparker/delve/issues/513

When that will be fixed, I'll see if any further changes need to be done. Meanwhile I think this should be safe to review.

Thank you.


src/com/goide/dlv/protocol/DlvApi.java
```diff
@@ -94,6 +98,10 @@
     public Function function;
   }
 
+  public static class StacktraceOut {
```
ignatov: Why do wee need these wrappers?

dlsniper: These wrappers are now present in delve as well. If there's a way to access the information directly without the wrapper that would be great (but unfortunately I couldn't see how to do this).

dlsniper: For reference this is how the API looks like right now: https://github.com/dlsniper/delve/blob/alphabetically-sort-server-methods/service/rpc2/server.go (it's not the complete version as more changes where merged after I've made this commit)

dlsniper:
I've sorted out the API issue with delve and now I'll need to finish this. At the moment the only thing that is not working is the `Set Value` functionality which I'll finish tonight.

dlsniper:
Done. I've fixed the remaining `Set` and `Detach` API differences, tested out everything related to to the debugger that I could think of (printing a struct, an int, a string, changing values, stopping the debugger, finishing the debugged process) and everything looked ok.

dlsniper:
Ping for a review on this please :)

kolkov:
Hi! What news about upgrade to API v2? @dlsniper @ignatov

zolotov:
@kolkov hi, any reason to do this?

dlsniper:
I would have to rework it before anything else but there are other things I need to fix / work on before that. @kolkov feel free to pick it up and redo it if you find anything missing right now.

@zolotov the only reason to do this would be to help delve remove the API v1 which is "deprecated". As this fell off the radar it will need another pass at the Delve API to be updated before anything else.

kolkov:
I thought that all improvements such as pretty variable printing etc. will be available only in version API v2.  Ask @derekparker about this please!

dlsniper:
I'm not aware of that, I'll need to check what delve does these days. When I've opened the PR only the public API was different but internally it was calling the same functions.

zolotov:
> @zolotov the only reason to do this would be to help delve remove the API v1 which is "deprecated". 

They free to remove anything they want, we depend on particular delve versions, not on its snapshot

dlsniper:
Yuo, there's no such pressure from them anyway. That's this was on the back-burner.

dlsniper:
Hi @zolotov / @ignatov I've updated this as Delve has had some recent changes.
Now there's a good reason to upgrade to the latest version of delve api: support for the `step out` command. I've also added support for the working directory while at it.

dlsniper:
I've also spent some time and we can't fix #2730 unless this gets merged (and use the API v2 in Delve).

kolkov:
Hi! What news about this PR?

Issue #2499, opened 2016-05-01
Return parameters shouldn't be reported as unused if returned by value


dlsniper:
In the following code

```go
package main

import (
	"fmt"
)

type demo interface {
	Hello(param1 string) (int, string)
}

type dem struct{}

// swap these
func (d dem) Hello(param1 string) (param2 int, param3 string) { // param2 is reported as unused even if it is used in the return
//func (d dem) Hello(param1 string) {
	fmt.Printf("Hello %s", param1)
	param3 = "ok"
	return 1
}

func Holla(d demo) {
	d.Hello("world")
}

func main() {
	d := dem{}
	Holla(d)
}
```

` param2 ` shouldn't be marked as unused since the return statement returns by value and the number of returned values matches the number of expected return values. ` param3 ` should be marked as unused since the return statement return insufficient values (if it would return no value it should not be considered as unused).

zolotov:
param2 shouldn't be marked as unused but not because it's used but because we cannot delete its name.

> param3 should be marked as unused 

param3 shouldn't be mark as unused because it's used

dlsniper:
Maybe I understand incorrectly how the issue should work so please bare with me.

> param3 shouldn't be mark as unused because it's used

While the assignment takes place for `param3` since the return statement returns explicit values the compiler will reject it. So my view on this is that while the assignment takes place, since the return doesn't use it (and it's not a naked return) then we should flag this as unused (or add a quickfix for `return` to send the missing parameters.

zolotov:
> While the assignment takes place for param3 since the return statement returns explicit values the compiler will reject it.

Oh, I see. Well, it makes sense but it is more complicated inspection and it requires proper control flow analyse which we're not plan to implement yet.
Unused inspection is much simpler thing, it should just report return parameter usages that unneeded named.

PR #2501, opened 2016-05-02
Improve unused parameters detection. Added quick fixes (fixes #2294)

### LINKED ISSUE #2294: Unused function arguments inspection
```go
package main

func demo(hello string) { // hello should be flagged as unused
	println("hello")
}

func main() {
	demo("hello")
}
```

----
**Original report:**

When a function doesn't use one of its arguments, it's still the same color as the used arguments. When editing python, java, c++ etc. unused function arguments are greyed out.

dlsniper:
This fixes the remaining case presented in #2294.
It also adds the quickfixes needed for the inspection.

### DIFF
--- a/src/com/goide/inspections/unresolved/GoUnusedParameterInspection.java
+++ b/src/com/goide/inspections/unresolved/GoUnusedParameterInspection.java
@@ -18,6 +18,7 @@ package com.goide.inspections.unresolved;
 
 import com.goide.inspections.GoInspectionBase;
 import com.goide.psi.*;
+import com.goide.quickfix.GoRenameToBlankQuickFix;
 import com.goide.runconfig.testing.GoTestFinder;
 import com.goide.runconfig.testing.GoTestFunctionType;
 import com.intellij.codeInspection.LocalInspectionToolSession;
@@ -28,6 +29,7 @@ import com.intellij.psi.PsiReference;
 import com.intellij.psi.search.searches.ReferencesSearch;
 import com.intellij.util.Query;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import java.util.List;
 
@@ -39,7 +41,7 @@ public class GoUnusedParameterInspection extends GoInspectionBase {
       @Override
       public void visitMethodDeclaration(@NotNull GoMethodDeclaration o) {
         super.visitMethodDeclaration(o);
-        visitDeclaration(o, false);
+        visitDeclaration(holder, o.getSignature(), false);
       }
 
       @Override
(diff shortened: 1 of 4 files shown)


dlsniper:
Please let me know if I should add a test for the quickfixes or not.
As for #2499, as soon as I understand it I'll either close it or look more into it.

src/com/goide/inspections/unresolved/GoUnusedParameterInspection.java
```diff
@@ -67,16 +69,34 @@ private void visitDeclaration(@NotNull GoFunctionOrMethodDeclaration o, boolean
...
+        Query<PsiReference> search;
         for (GoParameterDeclaration parameterDeclaration : parameters) {
+          if (parameterDeclaration.getParamDefinitionList().isEmpty()) {
+            search = ReferencesSearch.search(parameterDeclaration, parameterDeclaration.getUseScope());
+            if (search.findFirst() == null) {
+              holder.registerProblem(
+                parameterDeclaration,
+                "Unused " + what,
+                ProblemHighlightType.LIKE_UNUSED_SYMBOL,
+                new GoDeleteQuickFix("Remove " + what, parameterDeclaration.getElementType()));
```
zolotov: it's not appropriate quick fix, the fix should change all usages as well

zolotov: Also we cannot apply this fix for result parameters

zolotov:
This is how I see unused inspection:

for function parameters: report about any unused parameter (including unnamed) and provide following quickfixes
- rename to `_`
- remove parameter with fixing all usages (may be delayed until proper `change signature` refactoring is implemented)

for method's parameters, method's result parameters and function's result parameters: report about unused parameter only if all of them are unused or blank
quickfixes:
- make them unnamed

Maybe unused function/method parameter and function/method result parameters should be separate inspections, I'm not sure.
Also I'm not sure what we should do with function literals. What do you think?

dlsniper:
> for method's parameters, method's result parameters and function's result parameters: report about unused parameter only if all of them are unused or blank
> quickfixes:
> 
> - make them unnamed

I'm not sure I understand the ` only if all of them are unused or blank `. To implement a method of an interface the parameters can be named anything, as well as blank. Currently the inspection won't report any `_` as unused, so that's fine and it should be ok to rename any parameter from its name to `_`.

> Maybe unused function/method parameter and function/method result parameters should be separate inspections, I'm not sure.

For now I think it's fine to have it like this, like you said in the other issue, we don't have control flow analysis so until we get that, this can remain the same.

> Also I'm not sure what we should do with function literals.

Good finding as well about this, I'll add that to the list. There are two cases as far as I can see:
- one where the function literal is defined and used (either at the definition place or in the same method). For this case we could see them as functions
- one where the function literal is defined and satisfies a type / passed as parameter. For this case we could see them as methods (since there's a lot more logic involved in this).

Overall I'd say we could see function literals as methods to start with and later on improve on this if needed.

Thanks for feedback.

dlsniper:
Done, I've the changes into two commits, one for the function literals, one to allow the inspection to run on methods again.
If you'd like to drop the last one, let me know, else I can squash these.
Thanks.

zolotov:
> I'm not sure I understand the only if all of them are unused or blank.

What exactly didn't you understand? You can make parameters unnamed only if all of them are unused

> Currently the inspection won't report any _ as unused, so that's fine and it should be ok to rename any parameter from its name to _.

And what the point to have all blank parameters if you can make them unnamed?

testData/highlighting/methodExpr.go
```diff
@@ -20,8 +20,8 @@ func (v *Vertex) Abs() float64 {
 type T struct {
 	a int
 }
-func (tv  T) Mv(int) int { return 0 }  // value receiver
-func (tp *T) Mp(float32) float32 { return 1 }  // pointer receiver
+func (tv  T) Mv(<warning descr="Unused parameter">int</warning>) int { return 0 }  // value receiver
```
zolotov: we should report unused parameter on methods since we cannot guarantee this.

src/com/goide/inspections/unresolved/GoUnusedParameterInspection.java
```diff
@@ -67,16 +72,29 @@ private void visitDeclaration(@NotNull GoFunctionOrMethodDeclaration o, boolean
       }
 
       private void visitParameterList(List<GoParameterDeclaration> parameters, String what) {
+        Query<PsiReference> search;
         for (GoParameterDeclaration parameterDeclaration : parameters) {
+          if (parameterDeclaration.getParamDefinitionList().isEmpty()) {
+            search = ReferencesSearch.search(parameterDeclaration, parameterDeclaration.getUseScope());
```
zolotov: what are you trying to find on parameter declaration?

dlsniper:
> What exactly didn't you understand? You can make parameters unnamed only if all of them are unused

What would be the benefit for that? Less noise?

> And what the point to have all blank parameters if you can make them unnamed?

Because I think it's easier to read: ` func(a, _ string, c int) ` rather than ` func (a string, string, c int) `. Same goes for changing all the parameters to unnamed ones. Also it's easier to do this rather than have to split the parameters by type as well (at least until we have the change signature capability like you've mentioned).

zolotov:
> What would be the benefit for that? Less noise?

What is the benefit of whole unused parameter inspection or renaming parameter to `_`?

> Because I think it's easier to read: func(a, _ string, c int) rather than func (a string, string, c int)

`func (a string, string, c int)` is not the same as `func(a, _ string, c int)`. Please read spec or at least bnf before implementing any inspection.

dlsniper:
@zolotov I've change this according to the comments. There are two things that I still need to implement for it:
- report unused parameters on methods when all of them are unused. To do that, I'd like to understand where to report that? Do you think the signature is the correct place for this? The quick fix for this should be presented once, no?
- what do you think about functions with unnamed parameters? Should I create a todo ticket to add the inspection for this when we can do signature changes for functions so that we can suggest removing it?

PR #2505, opened 2016-05-08
Improve the json live template

dlsniper:
I've been using this for a while now and it's easier to use than having the backticks as well.

### DIFF
--- a/resources/liveTemplates/go.xml
+++ b/resources/liveTemplates/go.xml
@@ -89,14 +89,14 @@
       <option name='GO_BLOCK' value='true'/>
     </context>
   </template>
-  <template description='`xml:""`' name='xml' toReformat='true' toShortenFQNames='true'
-            value='`xml:"$END$"`'>
+  <template description='xml:""' name='xml' toReformat='true' toShortenFQNames='true'
+            value='xml:"$END$"'>
     <context>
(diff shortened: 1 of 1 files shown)


zolotov:
1. why?
2. will it be available inside backticks
3. description doesn't match value

dlsniper:
> 1. why?

I usually find myself typing the backtick (since IntelliJ will complete the other one) and then type jso<enter> When that happens I get two backticks then I have to manually fix this. 

> 2. will it be available inside backticks

yes, it seems to be working fine

> 3. description doesn't match value

I've fixed it, sorry.

dlsniper:
If there's a better way to fix the double backtick issue then that would be great (aside from changing how I type it, unfortunately it's in muscle memory by now).

zolotov:
> I usually find myself typing the backtick (since IntelliJ will complete the other one) and then type jso When that happens I get two backticks then I have to manually fix this.

Then change these templates in your IDE. Changing template in plugin affects all users who got used to these templates.

> If there's a better way to fix the double backtick issue then that would be great (aside from changing how I type it, unfortunately it's in muscle memory by now).

I believe there are some methods to fix it, file an issue please.

dlsniper:
Will do, thanks.

PR #2506, opened 2016-05-08
Added inspection for mixed named and unnamed parameters (fixes #2494)

### LINKED ISSUE #2494: New inspection: "Return value name required"
This code doesn't compile:
```go
func mixNamedAndUnnamedReturnValues() (a string, error) {
	return
}
```

I'd like the plugin to highlight this code red. 

I like it that ```error``` is parsed as type here (the compiler parses it as variable name) - it allows navigating to types while working on implementation. So I propose adding an inspection which would complain on unnamed parameters.

Note that this code is valid, though:
```go
func mixNamedAndUnnamedReturnValues() (string, a json.Decoder) {
	return
}
```

Plugin version: 0.11.1316

dlsniper:
The inspection might need a better reporting target. As for a quick fix, I'm not sure how to proceed on this as there are different issues with each approach (make everything named, make everything unnamed).

### DIFF
--- a/src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
+++ b/src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2013-2016 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.*;
+import com.goide.runconfig.testing.GoTestFinder;
+import com.goide.runconfig.testing.GoTestFunctionType;
+import com.intellij.codeInspection.LocalInspectionToolSession;
+import com.intellij.codeInspection.ProblemHighlightType;
+import com.intellij.codeInspection.ProblemsHolder;
+import com.intellij.openapi.progress.ProgressManager;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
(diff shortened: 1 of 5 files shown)


src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
```diff
@@ -0,0 +1,93 @@
...
+      @Override
+      public void visitMethodDeclaration(@NotNull GoMethodDeclaration o) {
+        super.visitMethodDeclaration(o);
+        visitDeclaration(holder, o.getSignature(), "Method");
+      }
+
+      @Override
+      public void visitFunctionDeclaration(@NotNull GoFunctionDeclaration o) {
+        super.visitFunctionDeclaration(o);
+        if (GoTestFinder.isTestFile(o.getContainingFile()) && GoTestFunctionType.fromName(o.getName()) != null) {
```
zolotov: Why?

zolotov: Test for this?

src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
```diff
@@ -0,0 +1,93 @@
...
+    for (GoParameterDeclaration parameterDeclaration : parameters.getParameterDeclarationList()) {
+      if (parameterDeclaration.getParamDefinitionList().isEmpty()) {
+        hasUnnamed = true;
+      }
+      else {
+        hasNamed = true;
+      }
+    }
+
+    if (hasNamed && hasUnnamed) {
```
zolotov: Might be inside `for`

src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
dlsniper: I've done a stupid copy paste and forgot to think for that, will remove, thanks.

src/com/goide/inspections/GoMixedNamedUnnamedParametersInspection.java
dlsniper: What I've noticed while working on this was that there's a difference between the Go parser and the plugin parser for the following Parameters ` a, b int, string, int `:
- the plugin parser sees it as: ` a, b int // ParamDecl1, string // unamed ParamDecl2, int // unamed ParamDecl3 ` 
- the Go parser sees it as: ` a, b int // ParamDecl1, string, int // ParamDecl2 + missing type info `

I can move it in the for loop but it also depends on how the reporting should be done for this. Should all the parameter/return list be flagged?

The quick fix for this one is not immediately clear, maybe the quickfix should be postponed until the change signature capability lands (if ever done at all)?

zolotov:
merged manually

PR #2508, opened 2016-05-08
Add inspection for assignment to receiver (fixes #2429)

### LINKED ISSUE #2429: Add inspection for assignment to method receiver
The value assigned in a method receiver is only visible in the scope of that method. As such, it would be good to warn the user about this issue.

Sample code:
```go
package main

import "fmt"

type demo struct{}

func (d *demo) change() {
	d = nil
	d.myVal()
}

func (d *demo) myVal() {
	fmt.Printf("my val: %#v\n", d)
}

func (d demo) change2() {
	d = 
}

func main() {
	d := &demo{}
	d.myVal()
	d.change()
	d.myVal()
}
```

dlsniper:


zolotov:
merged manually

PR #2509, opened 2016-05-08
Allow unused function inspection to omit exported functions (fixes #2374)

### LINKED ISSUE #2374: Exported functions marked as unused
- Plugin version (or commit hash): 0.11.1162
- IDE name and version: IDEA 16 EAP

Currently, alI exported functions (starting with a capital letter) that are not used within a library itself, are marked as unused. This seems odd to me. Most exported functions in a library are never used within the library itself, but I think it is wrong to mark them as unused since they are not primarily meant to be used withing the library.

dlsniper:


src/com/goide/inspections/unresolved/GoUnusedFunctionInspection.java
```diff
@@ -28,21 +28,30 @@
...
+import com.intellij.openapi.util.text.StringUtil;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.search.searches.ReferencesSearch;
 import org.jetbrains.annotations.NotNull;
 
+import javax.swing.*;
+
 public class GoUnusedFunctionInspection extends GoInspectionBase {
+  @SuppressWarnings("WeakerAccess")
+  public boolean ignoreExportedFunctions;
```
zolotov: What's the point of suppression?

zolotov:
1. all unused inspection should behave in the same way
2. the option is not discoverable at all
3. unneeded suppressions are not welcome

src/com/goide/inspections/unresolved/GoUnusedFunctionInspection.java
dlsniper: The inspection that IntelliJ has for this will flag it as unused. I'll try and see I can use the field as a private one. In the example I had to make the option work the field was public (but maybe it was used in a different place?).

dlsniper:
> all unused inspection should behave in the same way

Should I add this in the plugin option then to allow for any exported symbol to be flagged as unused or not?

> the option is not discoverable at all

I thought that's the pattern the IDE has so I've tried to follow it. Maybe it would be useful to have the option appear in the intention submenu? Not sure if this is possible or not (or maybe I should file an issue in YouTrack for this?)

Or would the solution I've mentioned above be good enough?

zolotov:
> Should I add this in the plugin option then to allow for any exported symbol to be flagged as unused or not?

No. Actually it's better to separate each unused inspection in two instead of using options, in this case user will be able to set up validating exported function for some limited scope

dlsniper:
> No. Actually it's better to separate each unused inspection in two instead of using options, in this case user will be able to set up validating exported function for some limited scope

Agreed but how should this be made more visible for the users then? If you have any examples, I would happily work on it. Thanks.

zolotov:
Inspections are in editor already and can be disabled via alt+enter

dlsniper:
In this case should we close https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2374 as won't fix?

zolotov:
> In this case should we close #2374 as won't fix?

no, why?

dlsniper:
I don't see how #2374 could be solved then. Better docs?  If we want the user to rely on the mechanisms IntelliJ already provides to allow for exported symbols to be flagged as used and not introduce anything else then the other option is detecting if the project is a library or not, which is next to impossible to do it reliable. Or what am I missing?

zolotov:
> Actually it's better to separate each unused inspection in two instead of using options

dlsniper:
Ok, I'll do that, thanks.

PR #2515, opened 2016-05-11
Split unused variables inspections into independent inspections

dlsniper:
Hi,

This is the result of the discussion here: https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2509#issuecomment-217804372 This is the first step to having separate inspections for each of the unused types. This PR addresses the variable / global variables.
If the way I've done this is ok, I'll continue with the other types as well.


zolotov:
Unused variable inspection already was splitted in two. As far as I can see the PR changes nothing

testData/inspections/suppression/fix/innerVariableDeclarationSuppressionFix2-after-highlighting.go
```diff
@@ -1,6 +1,6 @@
 package newmath
 
 func TestFunction() {
-	//noinspection GoUnusedVariable
```
zolotov: such changes breaks existing user's suppressions. avoid them

dlsniper:
It was split into global and non-global variables. If I understood correctly the comment https://github.com/go-lang-plugin-org/go-lang-idea-plugin/pull/2509#issuecomment-217804372 then we should also split them by visibility (exported and not exported).

testData/inspections/suppression/fix/innerVariableDeclarationSuppressionFix2-after-highlighting.go
dlsniper: Is there a way to maintain this while changing the inspection name? I know this is a breaking change.

zolotov: There is a way, but this is not the case, I don't see any reason to change inspection name

dlsniper: Maybe I've approached this the wrong way. I've split the GoUnusedVariable into two inspections but I think it can remain as a single inspection (since it doesn't matter if the variable is exported or not for a non-package level variable).

The problem would still happen for package level variables since we would still have two inspections (one for exported and one for unexported) in which case I don't know which one should be considered as a default for the existing `noinspection GoUnusedGlobalVariable`.

zolotov:
> It was split into global and non-global variables.

Yes, because it copies behavior of compiler, take a look at language spec, compiler code or at least at severity levels of these inspections.

So, the PR:
1. breaks IDE behavior 
2. unnecessarily breaks users suppressions
3. works badly on local variables with capitalized names hence fixes the initial issue in a wrong way

Please leave variables inspection and make similar changes in other unused inspections

dlsniper:
Will do, thanks.

PR #2521, opened 2016-05-13
Add empty declaration inspection with tests

grenki:
New request instead of old closed request

### DIFF
--- a/resources/META-INF/plugin.xml
+++ b/resources/META-INF/plugin.xml
@@ -198,6 +198,10 @@
     <localInspection language="go" displayName="Unused parameter inspection" groupPath="Go"
                      groupName="Declaration redundancy" enabledByDefault="true" level="WARNING"
                      implementationClass="com.goide.inspections.unresolved.GoUnusedParameterInspection"/>
+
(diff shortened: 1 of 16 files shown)


zolotov:
Merged.

Please add tests for `Cleanup` quickfix in separate PR

resources/META-INF/plugin.xml
```diff
@@ -198,6 +198,10 @@
     <localInspection language="go" displayName="Unused parameter inspection" groupPath="Go"
                      groupName="Declaration redundancy" enabledByDefault="true" level="WARNING"
                      implementationClass="com.goide.inspections.unresolved.GoUnusedParameterInspection"/>
+
+    <localInspection language="go" displayName="Empty declaration inspection" groupPath="Go"
+                     groupName="Declaration redundancy" enabledByDefault="true" level="WARNING"
```
dlsniper: I think this should be `WEAK_WARNING` since the code compiles and runs ok with empty declarations. Should I submit a PR to change ti?

zolotov: > since the code compiles and runs ok with empty declarations. 

unused functions don't lead to compilation error neither. No do not submit a PR

PR #2530, opened 2016-05-19
Redundant second index in slices

grenki:


src/com/goide/inspections/GoRedundantSecondIndexInSlicesInspection.java
```diff
@@ -0,0 +1,61 @@
...
+      @Override
+      public void visitIndexOrSliceExpr(@NotNull GoIndexOrSliceExpr o) {
+        Trinity<GoExpression, GoExpression, GoExpression> indexes = o.getIndices();
+        if (indexes.third == null && indexes.second instanceof GoCallExpr && o.getExpression() != null &&
+            o.getExpression().getReference() != null) {
+          if (o.getExpression() instanceof GoReferenceExpression && ((GoReferenceExpression)o.getExpression()).getQualifier() != null) {
+            return;
+          }
+          GoCallExpr callExpr = (GoCallExpr)indexes.second;
+          if (callExpr.getExpression().textMatches("len") &&
```
zolotov: you should check that call expr is resolved to builtin `len` function. See `GoPsiImplUtil.builtin(psiElement)` and `GoReferenceExpression#resolve`.

resources/META-INF/plugin.xml
```diff
@@ -223,6 +223,10 @@
     <!-- /redeclare inspections -->
 
     <!-- code style issues -->
+    <localInspection language="go" displayName="Redundant second index in slices inspection" groupPath="Go"
```
zolotov: move to redundant declaracy

zolotov:
merged manually

PR #2534, opened 2016-05-21
Added inspection for tagged fields (fixes #2365)

### LINKED ISSUE #2365: Add unexported field tag inspection
Given the following example
```go
package demo

type hello struct {
	world `json:"nope"`
}
```

The plugin should report the following: ` The struct field "world" is not exported but has a tag attached to it. Reflection will not work from other packages. ` and offer to capitilize the field as a quickfix.

This would help issues like this: https://www.reddit.com/r/golang/comments/47zfry/jsonunmarshal_not_working

dlsniper:


dlsniper:
To finish this I'd need to make the rename inspection work by also renaming the instances of the fields (which I don't know how to do yet). Any help is greatly appreciated. Thanks.

src/com/goide/psi/impl/GoElementFactory.java
```diff
@@ -214,4 +216,38 @@ private static String prepareVarOrConstDeclarationText(@NotNull String name, @Nu
     value = StringUtil.isNotEmpty(value) ? " = " + value : "";
     return name + type + value;
   }
+
+  public static GoStructType createMultipleFieldsFromSingleDeclaration(@NotNull Project project,
```
dlsniper: Another thing that I'm not sure of is this. Should this be done for all the structure, regardless of the FieldDeclaration on which the quickfix was invoked on?

PR #2548, opened 2016-05-30
Rework placeholder count inspection (fixes #2188)

### LINKED ISSUE #2188: Rework the placeholder detection in the placeholder count detection

dlsniper:
This matches the inspection with what `go vet` does. The problem is that `go vet` currently doesn't match the language specs / what the Go compiler does, thus the failing tests.
I'll work into getting this back in working order without changing the tests as the error is in the inspection logic not the test code.

### DIFF
--- a/src/com/goide/inspections/GoPlaceholderChecker.java
+++ b/src/com/goide/inspections/GoPlaceholderChecker.java
@@ -1,0 +1,386 @@
+/*
+ * Copyright 2013-2016 Sergey Ignatov, Alexander Zolotov, Florin Patan
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.goide.inspections;
+
+import com.goide.psi.GoFunctionOrMethodDeclaration;
+import com.intellij.openapi.util.Pair;
+import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.util.containers.ContainerUtil;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.*;
+
+import static com.goide.GoConstants.TESTING_PATH;
+import static com.goide.inspections.GoPlaceholderChecker.PrintfArgumentType.*;
+
+public class GoPlaceholderChecker {
+
+  // This holds the name of the known formatting functions and position of the string to be formatted
+  private static final Map<String, Integer> FORMATTING_FUNCTIONS = ContainerUtil.newHashMap(
+    Pair.pair("errorf", 0),
+    Pair.pair("fatalf", 0),
+    Pair.pair("fprintf", 1),
+    Pair.pair("fscanf", 1),
+    Pair.pair("logf", 0),
+    Pair.pair("panicf", 0),
+    Pair.pair("printf", 0),
+    Pair.pair("scanf", 0),
+    Pair.pair("skipf", 0),
+    Pair.pair("sprintf", 0),
+    Pair.pair("sscanf", 1));
+
+  private static final Set<String> PRINTING_FUNCTIONS = ContainerUtil.newHashSet(
+    "error",
+    "error",
+    "fatal",
+    "fprint",
+    "fprintln",
+    "log",
+    "panic",
+    "panicln",
+    "print",
+    "println",
+    "sprint",
+    "sprintln"
+  );
+
+  protected enum PrintfArgumentType {
+    ANY(-1),
+    BOOL(1),
+    INT(2),
+    RUNE(3),
+    STRING(4),
+    FLOAT(5),
+    COMPLEX(6),
+    POINTER(7);
+
+    private int myMask;
+
+    PrintfArgumentType(int mask) {
+      myMask = mask;
+    }
+
+    public int getValue() {
+      return myMask;
+    }
+  }
+
+  enum PrintVerb {
+    Percent('%', "", 0),
+    b('b', " -+.0", INT.getValue() | FLOAT.getValue() | COMPLEX.getValue()),
+    c('c', "-", RUNE.getValue() | INT.getValue()),
+    d('d', " -+.0", INT.getValue()),
+    e('e', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    E('E', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    f('f', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    F('F', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    g('g', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    G('G', " -+.0", FLOAT.getValue() | COMPLEX.getValue()),
+    o('o', " -+.0#", INT.getValue()),
+    p('p', "-#", POINTER.getValue()),
+    q('q', " -+.0#", RUNE.getValue() | INT.getValue() | STRING.getValue()),
+    s('s', " -+.0", STRING.getValue()),
+    t('t', "-", BOOL.getValue()),
+    T('T', "-", ANY.getValue()),
+    U('U', "-#", RUNE.getValue() | INT.getValue()),
+    V('v', " -+.0#", ANY.getValue()),
+    x('x', " -+.0#", RUNE.getValue() | INT.getValue() | STRING.getValue()),
+    X('X', " -+.0#", RUNE.getValue() | INT.getValue() | STRING.getValue());
+
+    private char myVerb;
+    private String myFlags;
+    private int myMask;
+
+    PrintVerb(char verb, String flags, int mask) {
+      myVerb = verb;
+      myFlags = flags;
+      myMask = mask;
+    }
+
+    public char getVerb() {
+      return myVerb;
+    }
+
+    @NotNull
+    public String getFlags() {
+      return myFlags;
+    }
+
+    public int getMask() {
+      return myMask;
+    }
+
+    @Nullable
+    public static PrintVerb getByVerb(char verb) {
+      for (PrintVerb v : values()) {
+        if (verb == v.getVerb()) return v;
+      }
+      return null;
+    }
+  }
+
+  public static boolean isFormattingFunction(String functionName) {
+    return FORMATTING_FUNCTIONS.containsKey(functionName);
+  }
+
+  public static boolean isPrintingFunction(String functionName) {
+    return PRINTING_FUNCTIONS.contains(functionName);
+  }
+
+  static class Placeholder {
+    private final String placeholder;
+    private final int startPos;
+    private final State state;
+    private final PrintVerb verb;
+    private final List<Integer> arguments;
+    private final String flags;
+
+    enum State {
+      OK,
+      MISSING_VERB_AT_END,
+      ARGUMENT_INDEX_NOT_NUMERIC
+    }
+
+    Placeholder(State state, int startPos, String placeholder, String flags, List<Integer> arguments, PrintVerb verb) {
+      this.placeholder = placeholder;
+      this.startPos = startPos;
+      this.verb = verb;
+      this.state = state;
+      this.arguments = arguments;
+      this.flags = flags;
+    }
+
+    public String getPlaceholder() {
+      return placeholder;
+    }
+
+    public int getPosition() {
+      return arguments.get(arguments.size() - 1);
+    }
+
+    public State getState() {
+      return state;
+    }
+
+    public PrintVerb getVerb() {
+      return verb;
+    }
+
+    public List<Integer> getArguments() {
+      return arguments;
+    }
+
+    public String getFlags() {
+      return flags;
+    }
+
+    public int getStartPos() {
+      return startPos;
+    }
+  }
+
+  @NotNull
+  public static List<Placeholder> parsePrintf(@NotNull String placeholderText) {
+    List<Placeholder> placeholders = new ArrayList<Placeholder>();
+    int argNum = 1;
+    int w;
+    for (int i = 0; i < placeholderText.length(); i += w) {
+      w = 1;
+      if (placeholderText.charAt(i) == '%') {
+        FormatState state = parsePrintfVerb(placeholderText.substring(i), i, argNum);
+        w = state.format.length();
+
+        // We are not interested in %% which prints %
+        if (state.state == Placeholder.State.OK && state.verb == PrintVerb.Percent) {
+          // Special magic case for allowing things like %*% to pass (which are sadly valid expressions)
+          if (state.format.length() == 2) continue;
+        }
+
+        placeholders.add(state.toPlaceholder());
+
+        // We only consider ok states as valid to increase the number of arguments. Should we?
+        if (state.state != Placeholder.State.OK) continue;
+
+        if (!state.indexed) {
+          if (!state.argNums.isEmpty()) {
+            int maxArgNum = Collections.max(state.argNums);
+            if (argNum < maxArgNum) {
+              argNum = maxArgNum;
+            }
+          }
+        }
+        else {
+          argNum = state.argNums.get(state.argNums.size() - 1);
+        }
+        argNum++;
+      }
+    }
+
+    return placeholders;
+  }
+
+  protected static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration function) {
+    Integer position = FORMATTING_FUNCTIONS.get(StringUtil.toLowerCase(function.getName()));
+    if (position != null) {
+      String importPath = function.getContainingFile().getImportPath(false);
+      if ("fmt".equals(importPath) || "log".equals(importPath) || TESTING_PATH.equals(importPath)) {
+        return position;
+      }
+    }
+    return -1;
+  }
+
+  private static class FormatState {
+    @Nullable private PrintVerb verb;   // the format verb: 'd' for "%d"
+    private String format;              // the full format directive from % through verb, "%.3d"
+    @NotNull private String flags = ""; // the list of # + etc
+    private boolean indexed;            // whether an indexing expression appears: %[1]d
+    private final int startPos;         // index of the first character of the placeholder in the formatting string
+
+    // the successive argument numbers that are consumed, adjusted to refer to actual arg in call
+    private final List<Integer> argNums = new ArrayList<Integer>();
+
+    // Keep track of the parser state
+    private Placeholder.State state;
+
+    // Used only during parse.
+    private int argNum;           // Which argument we're expecting to format now
+    private int nBytes = 1;       // number of bytes of the format string consumed
+
+    FormatState(String format, int startPos, int argNum) {
+      this.format = format;
+      this.startPos = startPos;
+      this.argNum = argNum;
+    }
+
+    @NotNull
+    private Placeholder toPlaceholder() {
+      return new Placeholder(state, startPos, format, flags, argNums, verb);
+    }
+  }
+
+  @NotNull
+  private static FormatState parsePrintfVerb(@NotNull String format, int startPos, int argNum) {
+    FormatState state = new FormatState(format, startPos, argNum);
+
+    parseFlags(state);
+
+    if (!parseIndex(state)) return state;
+
+    // There may be a width
+    if (!parseNum(state)) return state;
+
+    if (!parsePrecision(state)) return state;
+
+    // Now a verb, possibly prefixed by an index (which we may already have)
+    if (!parseIndex(state)) return state;
+
+    if (state.nBytes == format.length()) {
+      state.state = Placeholder.State.MISSING_VERB_AT_END;
+      return state;
+    }
+
+    state.verb = PrintVerb.getByVerb(state.format.charAt(state.nBytes));
+    if (state.verb != null && state.verb != PrintVerb.Percent) state.argNums.add(state.argNum);
+    state.nBytes++;
+    state.format = state.format.substring(0, state.nBytes);
+    state.state = Placeholder.State.OK;
+
+    return state;
+  }
+
+  public static boolean hasPlaceholder(@NotNull String formatString) {
+    return StringUtil.containsChar(formatString, '%');
+  }
+
+  private static void parseFlags(@NotNull FormatState state) {
+    String knownFlags = "#0+- ";
+    StringBuilder flags = new StringBuilder(state.flags);
+    while (state.nBytes < state.format.length()) {
+      if (StringUtil.containsChar(knownFlags, state.format.charAt(state.nBytes))) {
+        flags.append(state.format.charAt(state.nBytes));
+      }
+      else {
+        state.flags = flags.toString();
+        return;
+      }
+      state.nBytes++;
+    }
+    state.flags = flags.toString();
+  }
+
+  private static void scanNum(@NotNull FormatState state) {
+    while (state.nBytes < state.format.length()) {
+      if (!StringUtil.isDecimalDigit(state.format.charAt(state.nBytes))) {
+        return;
+      }
+      state.nBytes++;
+    }
+  }
+
+  private static boolean parseIndex(@NotNull FormatState state) {
+    if (state.nBytes == state.format.length() || state.format.charAt(state.nBytes) != '[') return true;
+
+    state.indexed = true;
+    state.nBytes++;
+    int start = state.nBytes;
+    scanNum(state);
+    if (state.nBytes == state.format.length() || state.nBytes == start || state.format.charAt(state.nBytes) != ']') {
+      state.state = Placeholder.State.ARGUMENT_INDEX_NOT_NUMERIC;
+      return false;
+    }
+
+    int arg;
+    try {
+      arg = Integer.parseInt(state.format.substring(start, state.nBytes));
+    }
+    catch (NumberFormatException ignored) {
+      state.state = Placeholder.State.ARGUMENT_INDEX_NOT_NUMERIC;
+      return false;
+    }
+
+    state.nBytes++;
+    state.argNum = arg;
+
+    return true;
+  }
+
+  private static boolean parseNum(@NotNull FormatState state) {
+    if (state.nBytes < state.format.length() && state.format.charAt(state.nBytes) == '*') {
+      state.nBytes++;
+      state.argNums.add(state.argNum);
+      state.argNum++;
+    }
+    else {
+      scanNum(state);
+    }
+
+    return true;
+  }
+
+  private static boolean parsePrecision(@NotNull FormatState state) {
+    if (state.nBytes < state.format.length() && state.format.charAt(state.nBytes) == '.') {
+      state.flags += '.';
+      state.nBytes++;
+      if (!parseIndex(state)) return false;
+      if (!parseNum(state)) return false;
+    }
+
+    return true;
+  }
+}
--- a/src/com/goide/inspections/GoPlaceholderCountInspection.java
+++ b/src/com/goide/inspections/GoPlaceholderCountInspection.java
@@ -16,142 +16,26 @@
 
 package com.goide.inspections;
 
+import com.goide.inspections.GoPlaceholderChecker.Placeholder;
+import com.goide.inspections.GoPlaceholderChecker.PrintVerb;
 import com.goide.psi.*;
+import com.goide.psi.impl.GoPsiImplUtil;
+import com.goide.psi.impl.GoTypeUtil;
 import com.intellij.codeInspection.LocalInspectionToolSession;
 import com.intellij.codeInspection.ProblemHighlightType;
 import com.intellij.codeInspection.ProblemsHolder;
-import com.intellij.openapi.util.Pair;
+import com.intellij.openapi.util.TextRange;
 import com.intellij.openapi.util.text.StringUtil;
+import com.intellij.psi.ElementManipulators;
 import com.intellij.psi.PsiElement;
 import com.intellij.psi.PsiReference;
-import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Collections;
 import java.util.List;
-import java.util.Map;
-import java.util.regex.Matcher;
-import java.util.regex.Pattern;
-
-import static com.goide.GoConstants.TESTING_PATH;
 
 public class GoPlaceholderCountInspection extends GoInspectionBase {
-  private static final Pattern PLACEHOLDER_PATTERN =
-    Pattern.compile("((?<!(?:[^%]%|%%%))%(?:#|\\+|-)?((\\[\\d+\\]|\\*?)?\\.?)*(\\d*\\.?\\d*)*\\s?(v|T|t|b|c|d|o|q|x|X|U|b|e|E|f|F|g|G|s|q|x|X|p))");
-
-  private static final Pattern INDEXED_PLACEHOLDER_PATTERN = Pattern.compile("(?:\\[(\\d+)\\])");
-
-  // This holds the name of the known formatting functions and position of the string to be formatted
-  private static final Map<String, Integer> FORMATTING_FUNCTIONS = ContainerUtil.newHashMap(
-    Pair.pair("printf", 0),
-    Pair.pair("sprintf", 0),
-    Pair.pair("errorf", 0),
-    Pair.pair("fprintf", 1),
-    Pair.pair("fscanf", 1),
-    Pair.pair("scanf", 0),
-    Pair.pair("sscanf", 1),
-    Pair.pair("fatalf", 0),
-    Pair.pair("panicf", 0),
-    Pair.pair("logf", 0),
-    Pair.pair("skipf", 0));
-
-  private static int getPlaceholderPosition(@NotNull GoFunctionOrMethodDeclaration function) {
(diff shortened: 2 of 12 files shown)


zolotov:
@dlsniper thanks for PR. This is really great that you repeated whole `govet` logic. The only thing is that it's better to write Java in Java style, not in Go. I'll comment some snippets with inappropriate go-style in java.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+    public boolean parseIndex() {
+      if (nBytes == format.length() || format.charAt(nBytes) != '[') return true;
+
+      indexed = true;
+      nBytes++;
+      int start = nBytes;
+      scanNum();
+      if (nBytes == format.length() || nBytes == start || format.charAt(nBytes) != ']') {
+        String message = "Illegal syntax for <code>#ref</code> argument index";
+        myHolder.registerProblem(myPlaceholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
```
zolotov: This is strange that `state`-class reports something to problems holder. Actually it shouldn't do anything besides storing data. But ok, at least it should not interact to outer world.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+        else {
+          this.flags = flags.toString();
+          return;
+        }
+      }
+      this.flags = flags.toString();
+    }
+
+    private void scanNum() {
+      for (; nBytes < format.length(); nBytes++) {
```
zolotov: we have `while` loop in java

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+        if (format.charAt(nBytes) < '0' || '9' < format.charAt(nBytes)) {
+          return;
```
zolotov: Character#isDigit?

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import static com.goide.GoConstants.TESTING_PATH;
+import static com.goide.inspections.GoPlaceholderChecker.PrintfArgType.*;
+
+public class GoPlaceholderChecker {
```
zolotov: Fill nullability

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+      String importPath = function.getContainingFile().getImportPath(false);
+      if ("fmt".equals(importPath) || "log".equals(importPath) || TESTING_PATH.equals(importPath)) {
+        return position;
+      }
+    }
+    return -1;
+  }
+
+  @Nullable
+  private static String resolve(@NotNull GoExpression argument) {
```
zolotov: I don't understand the ordering of methods. I read three methods and didn't understand why I did this

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+    state.verb = state.format.charAt(state.nBytes);
+    state.nBytes++;
+    if (state.verb != '%') state.argNums.add(state.argNum);
+    state.format = state.format.substring(0, state.nBytes);
+
+    return state;
+  }
+
+  private static boolean isStringPlaceholder(@NotNull GoExpression argument) {
+    GoType goType = argument.getGoType(null);
```
zolotov: see GoTypeUtil

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+  public void checkPrint() {
+    // TODO florin: Implement the print function checking
+    // Ref code: https://github.com/golang/go/blob/79f7ccf2c3931745aeb97c5c985b6ac7b44befb4/src/cmd/vet/print.go#L587
+  }
+
+  private class FormatState {
+    public char verb;             // the format verb: 'd' for "%d"
+    public String format;         // the full format directive from % through verb, "%.3d".
+    public String flags;          // the list of # + etc.
+    public List<Integer> argNums; // the successive argument numbers that are consumed, adjusted to refer to actual arg in call
```
zolotov: it's always initialized in `parsePrintfVerb` and it's always `new ArrayList`. Do you have any reason to not to declare it as final

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+    public int type;
+
+    public PrintVerb(char verb, String flags, int type) {
+      this.verb = verb;
+      this.flags = flags;
+      this.type = type;
+    }
+  }
+
+  @NotNull private static List<PrintVerb> printVerbs = ContainerUtil.newArrayList(
```
zolotov: why not final? are you going to reassign it?

why not enum? are you going to change inner values?

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+    return null;
+  }
+
+  public void checkPrint() {
+    // TODO florin: Implement the print function checking
+    // Ref code: https://github.com/golang/go/blob/79f7ccf2c3931745aeb97c5c985b6ac7b44befb4/src/cmd/vet/print.go#L587
+  }
+
+  private class FormatState {
+    public char verb;             // the format verb: 'd' for "%d"
```
zolotov: Why all fields are public? Who is responsible for changing state of `state`? What `FormatState` is it exactly, parser or state? If parser, it shouldn't expose all it's state to anyone and shouldn't allow to change it. If state, then it shouldn't parse anything.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,497 @@
...
+
+  private static boolean isFunctionValue(GoExpression expression) {
+    return expression instanceof GoCallExpr;
+  }
+
+  private boolean okPrintfArg(@NotNull GoExpression placeholder, @NotNull List<GoExpression> arguments, @NotNull FormatState state) {
+    PrintVerb v = null;
+    boolean found = false;
+
+    for (int i = 0; i < printVerbs.size(); i++) {
```
zolotov: foreach

dlsniper:
@zolotov thank you so much for the review. I was fir focused on getting it to run then passing the tests then make it correct Java. Your comments will help me out a lot, I'll try to fix everything today. Thank you!

src/com/goide/inspections/GoPlaceholderChecker.java
zolotov: `found` variable is completely redundant: if you cannot find `verb` the `v` variable should be null.
whole loop looks silly. just replace printVerbs with any map.

dlsniper:
I'll also add the test data from `go vet` as soon as our tests are fixed and the code is written according to Java style. For now, `go vet` still has inferior matching of our current solution so I'll also contribute back to Go the fix for that once it's working.

src/com/goide/inspections/GoPlaceholderChecker.java
dlsniper: From what I understand in the method documentation, Character#isDigit will match also non-Latin digits. Go is explicit at supporting only ` 0-9 ` digits for everything here so we could even add a "Number is not a valid Latin-1 digit" error here. What do you think?

zolotov: You're right. `StringUtil#isDecimalDigit` then

dlsniper:
@zolotov I've updated this. I've changed the order of methods as well, but not sure if that has improved the readability or not. All the other items have been done.
I'm not sure if this is the best construct to solve the the print verbs definition 
```
private enum PrintVerb {
    Percent(Couple.of('%', Couple.of("", 0))),
...
```
so maybe that's something to look after. It could be simplified for now as we don't use the type mask yet (needed in the type assertion between parameter type and placeholder type).

Once this passes your review, I'll proceed with adding the tests from `go vet` and checking if any other changes need to be done.

Thank you.

zolotov:
Actually I don't understand why do you need Couple there. 

1. `Couple` is shorthand for `Pair` and should be used when both object have the same type
2. It's supposed to use static types anywhere you can in Java
3. Why don't just 
```
private enum PrintVerb {
PERCENT('%', "", 0),
 b('b', " -+.0", argInt.getValue() | argFloat.getValue() | argComplex.getValue())
...
```

with descriptive argument names, types and so on.

dlsniper:
> 3. Why don't just

Oh, mostly because I didn't knew I can do that in Java and the enum docs I've seen weren't mentioning that. I'll change it right now. Thank you.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,502 @@
...
+    argBool(1),
+    argInt(2),
+    argRune(3),
+    argString(4),
+    argFloat(5),
+    argComplex(6),
+    argPointer(7),
+    anyType(-1);
+
+    private int myVal;
```
zolotov: The comment says
> // The types of expressions a printf verb accepts. It is a bitmask.

The first part is presented in enum's name. The second you can show by renaming `myVal` variable and `getValue` method. Comment is redundant.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,502 @@
...
+    Pair.pair("fscanf", 1),
+    Pair.pair("logf", 0),
+    Pair.pair("panicf", 0),
+    Pair.pair("printf", 0),
+    Pair.pair("scanf", 0),
+    Pair.pair("skipf", 0),
+    Pair.pair("sprintf", 0),
+    Pair.pair("sscanf", 1));
+
+  public static final List<String> PRINT_FUNCTIONS = ContainerUtil.newArrayList(
```
zolotov: 1. the field is used only in `GoPlaceholderCountInspection`, so why it's here and public?
2. if the list is used for `contains` operation only it shouldn't be a list.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,502 @@
...
+      myVal = val;
+    }
+
+    public int getValue() {
+      return myVal;
+    }
+  }
+
+  private enum PrintVerb {
+    Percent(Couple.of('%', Couple.of("", 0))),
```
zolotov: Since you don't depend on enum's fields names it makes sense to make them more descriptive

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,502 @@
...
+      }
+    }
+
+    return result.toString();
+  }
+
+  private FormatState parsePrintfVerb(@NotNull GoExpression placeholder, @NotNull String format, int firstArg, int argNum) {
+    FormatState state = new FormatState(format, argNum, firstArg);
+
+    // There may be flags
```
zolotov: comments about flags before invoking `parseFlags` method looks redundant

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,502 @@
...
+
+    String placeholderText = resolve(placeholder);
+    if (placeholderText == null) return;
+
+    // We always receive the text with double quotes at the beginning and at the end so remove them
+    placeholderText = placeholderText.substring(1, placeholderText.length() - 1);
+
+    if (!hasPlaceholder(placeholderText) && arguments.size() > placeholderPosition) {
+      String message = "Value used for formatting text does not appear to contain a placeholder";
+      myProblemsHolder.registerProblem(placeholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
```
zolotov: Still strange dependency on problems holder here. You have a parser which just parses some string.
First of all it's weird that the parser knows about problemsHolder. In the second, it means that you can parse printf placeholder only in inspections, so no completion support in future. In the third, what is the real responsibility of GoPlaceholderChecker? If it's a Checker it should check and return the result (success or not). If it's parser it should parse and return parsed representation of printf argument (e.g. a Map). If it's inspection, it should visit elements and create problem descriptors for some of them. You're trying to create a class that knows everything and does all these three things. It's hard to read, hard to modify and hard to reuse. Try to keep single responsibility for each class and method.

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -162,23 +40,18 @@ public void visitCallExpr(@NotNull GoCallExpr o) {
...
-          String message = "Value used for formatting text does not appear to be a string";
-          holder.registerProblem(placeholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
-          return;
-        }
 
-        int parameterCount = arguments.size() - placeholderPosition - 1;
-        int placeholdersCount = getPlaceholders(placeholder);
-        if (placeholdersCount == -1 || placeholdersCount == parameterCount) return;
+        String functionName = ((GoFunctionOrMethodDeclaration)resolved).getName();
+        if (functionName == null) return;
```
zolotov: toLowerCase return null for null argument, no needed to check function name here, check it later and avoid variable reassignments

src/com/goide/inspections/GoPlaceholderChecker.java
dlsniper: I have the same opinion on this. I'll try to organize the code better and split its concerns but this might take a bit. Thank you.

dlsniper:
I've split up all the things and our tests are passing (in the process I was able to add a few more cases previously unsupported). I hope the layout it's better now.
The next step is to fix the inconsistency in handling indexes as there's currently an off-by-one error in some cases (highlighted by the go vet tests).
I'll continue in the coming days with fixing this.
Finally, I've noticed that the performance tests are failing but I don't know what could be the cause for this (I am receiving emails from TeamCity about them).

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,456 @@
...
+      return null;
+    }
+  }
+
+  public static boolean isFormattingFunction(String functionName) {
+    return FORMATTING_FUNCTIONS.containsKey(functionName);
+  }
+
+  public static boolean isPrintingFunction(String functionName) {
+    return PRINTING_FUNCTIONS.contains(functionName);
```
zolotov: avoid `contains` on lists

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,456 @@
...
+    private final char verb;
+    private final List<Integer> arguments;
+    private final String flags;
+
+    enum State {
+      OK(0),
+      MISSING_VERB_AT_END(1),
+      ARGUMENT_INDEX_NOT_NUMERIC(2);
+
+      private int state;
```
zolotov: is this used anywhere?

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,456 @@
...
+      else if (expr instanceof GoAddExpr) {
+        result.append(getValue(expr));
+      }
+    }
+
+    return result.toString();
+  }
+
+  private static class FormatState {
+    private char verb;                  // the format verb: 'd' for "%d"
```
zolotov: As I can see this field is not used much. It's compared to `%` and it's converted to Verb. So why it's not nullable Verb field?

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,263 @@
...
-    if (expression == null) return "";
-    StringBuilder result = new StringBuilder();
-    for (GoExpression expr : expression.getExpressionList()) {
-      if (expr instanceof GoStringLiteral) {
-        result.append(expr.getText());
+  private static int computeMaxArgsNum(@NotNull List<Placeholder> placeholders) {
+    int maxArgsNum = 0;
+    for (Placeholder placeholder : placeholders) {
+      List<Integer> arguments = placeholder.getArguments();
+      if (arguments == null || arguments.isEmpty()) continue;
```
zolotov: is arguments nullable?

testData/highlighting/struct.go
```diff
@@ -58,7 +58,7 @@ func main() {
 
     c := Car{4}
     Println("A Car has this many wheels:", c.wheelCount)
-    Println("A Car has this many wheels:", c.numberOfWheels)
+    Println("A Car has this many wheels:", c.numberOfWheels())
```
zolotov: both lines are identical, is this intentionally?

src/com/goide/psi/impl/GoTypeUtil.java
```diff
@@ -41,6 +41,13 @@ private static boolean isArray(@Nullable GoType type) {
 
   public static boolean isString(@Nullable GoType type) {
     GoType underlyingType = type != null ? type.getUnderlyingType() : null;
-    return underlyingType != null && underlyingType.textMatches("string") && GoPsiImplUtil.builtin(underlyingType);
+    return underlyingType != null &&
+           underlyingType.getText() != null &&
```
zolotov: ?

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,263 @@
...
+    @NotNull Placeholder fmtPlaceholder,
+    int callArgsNum,
+    int firstArg) {
+
+    int argNum = fmtPlaceholder.getPosition();
+    if (argNum < 0) return false;
+
+    if (argNum == 0) {
+      String message = String.format("Index value [0] for <code>#ref</code>(\"%s\"); indexes start at 1", fmtPlaceholder.getPlaceholder());
+      holder.registerProblem(placeholder, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
```
zolotov: the function looks like it answers to the question, not reports any problems

src/com/goide/psi/impl/GoTypeUtil.java
dlsniper: That's to prevent a NPE with cgo, I've got it from the `cgotest.go` file.

zolotov: maybe it makes sense to implement `textMatches` or `getText` methods for CTypes here. Or maybe check GoCType explicitly like `underlyingType instanceof GoCType`. Anyway test is required.

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,456 @@
...
+    if (resolved instanceof GoVarDefinition) {
+      value = getValue(((GoVarDefinition)resolved).getValue());
+    }
+    else if (resolved instanceof GoConstDefinition) {
+      value = getValue(((GoConstDefinition)resolved).getValue());
+    }
+
+    // We always receive the text with double quotes at the beginning and at the end so remove them
+    if (value != null) {
+      value = value.substring(1, value.length() - 1);
```
dlsniper: Is there a way to avoid getting the string value double quoted? There'd at least one more place where I had to take this into account to do the check.

zolotov: there is a method StringUtil for this. As for `GoStringLiteral` it's better to use `com.intellij.psi.ElementManipulators#getValueText`

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,456 @@
...
+
+  // todo: implement ConstEvaluator
+  @Nullable
+  private static String getValue(@Nullable GoExpression expression) {
+    if (expression instanceof GoStringLiteral) {
+      return expression.getText();
+    }
+    if (expression instanceof GoAddExpr) {
+      String sum = getValue((GoAddExpr)expression);
+      return sum.isEmpty() ? null : sum;
```
zolotov: StringUtil.nullize

src/com/goide/inspections/GoPlaceholderChecker.java
dlsniper: I've tried `indexOf()` but the IDEA will say I should use `contains()`. Should I use a simple `for` instead?

src/com/goide/psi/impl/GoTypeUtil.java
dlsniper: I've changed this to check if the type is not `instanceof GoCType` and added a test for this but it's still in the placeholder test suite, it isn't contained on its own suite. Should I do that instead?

dlsniper:
Sorry I've disappeared, I've fixed this and now tests pass again.

There are a few todos left in this (some mostly because we don't have type checking just yet). A few others could be done later on as an improvement over both the existing solution and `go vet` itself.

The other main difference between what we'll support now and what `go vet` does is that we won't support arbitrary functions which match the name and signature with the any of the printing functions (e.g. we support `fmt.Printf()` but don't support `dlsniper.Printf()` whereas would support it).

As for the unused methods, I've left them for now to be able to easily use them for various quickfixes which could be applied to this inspection.

Finally I should squash this before merging so let me know if you think it's ok to do so.

src/com/goide/inspections/GoPlaceholderChecker.java
zolotov: No, use Set instead

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,440 @@
...
+    state.verb = PrintVerb.getByVerb(state.format.charAt(state.nBytes));
+    if (state.verb != null && state.verb != PrintVerb.Percent) state.argNums.add(state.argNum);
+    state.nBytes++;
+    state.format = state.format.substring(0, state.nBytes);
+    state.state = Placeholder.State.OK;
+
+    return state;
+  }
+
+  protected static boolean hasPlaceholder(@NotNull String formatString) {
```
zolotov: inspection is the only client of this method, move the method to inspection

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,440 @@
...
+    return StringUtil.contains(formatString, "%");
```
zolotov: Use `containsChar`

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,440 @@
...
+      String importPath = function.getContainingFile().getImportPath(false);
+      if ("fmt".equals(importPath) || "log".equals(importPath) || TESTING_PATH.equals(importPath)) {
+        return position;
+      }
+    }
+    return -1;
+  }
+
+  @Nullable
+  protected static String resolve(@NotNull GoExpression argument) {
```
zolotov: inspection is the only client of this method, move the method to inspection

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,261 @@
...
-      countPlaceholders = INDEXED_PLACEHOLDER_PATTERN.matcher(placeholders.group());
-      while (countPlaceholders.find()) {
-        int index = StringUtil.parseInt(countPlaceholders.group().replaceAll("\\[|\\]", ""), -1);
-        if (index > 0 && index != indexCount) {
-          indexCount = index;
+    GoExpression firstArg = arguments.get(0);
+    if (GoTypeUtil.isString(firstArg.getGoType(null))) {
+      String firstArgText = GoPlaceholderChecker.resolve(firstArg);
+      if (firstArgText != null) {
+        if (GoPlaceholderChecker.hasPlaceholder(firstArgText)) {
```
zolotov: Any string with `%` will be highlighted? Even `fmt.Println("10%")`?

dlsniper: Unfortunately yes. At least that's the current behavior of `go vet`. Maybe we should make this smarter and call for this if the string actually contains a formatting directive?

zolotov: Not sure but it looks strange. Well let's keep it in this way until somebody complain

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,261 @@
...
+        String argText = GoPlaceholderChecker.resolve(argument);
+        if (argText != null && argText.endsWith("\\n")) {
+          String message = "Function already ends with new line";
+          holder.registerProblem(argument, message, ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
         }
       }
-      if (maxIndexCount < indexCount) {
-        maxIndexCount = indexCount;
+      else if (GoTypeUtil.isFunction(goType)) {
+        String message = "Argument <code>#ref</code> is not a function call";
```
zolotov: add `#loc` at the end, try to run `Inspect code` action to see how it looks

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,261 @@
...
-  private static boolean isStringPlaceholder(GoExpression argument) {
-    GoType goType = argument.getGoType(null);
-    GoTypeReferenceExpression typeReferenceExpression = goType != null ? goType.getTypeReferenceExpression() : null;
-    return typeReferenceExpression != null && typeReferenceExpression.textMatches("string");
   }
 
-  @Nullable
-  private static String resolve(@NotNull GoExpression argument) {
-    if (argument instanceof GoStringLiteral) return argument.getText();
+  private static void checkPrintf(
```
zolotov: Usually we format it in this way:

```
private static void checkPrint(@NotNull ProblemsHolder holder,
                                 @NotNull GoCallExpr callExpr,
                                 @NotNull GoFunctionOrMethodDeclaration declaration) {
```

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,261 @@
...
+    if (argNum < 0) return false;
+
+    if (argNum == 0) {
+      holder.registerProblem(placeholder, "Index value [0] is not allowed", ProblemHighlightType.GENERIC_ERROR_OR_WARNING);
+      return false;
+    }
+
+    if (argNum <= callArgsNum) return true;
+    if (GoPsiImplUtil.hasVariadic(callExpr.getArgumentList())) return false;
+    if (argNum <= callArgsNum) return true;
```
zolotov: typo

zolotov:
Some cleanup is needed, but the PR looks really good to me so far, thank you.

src/com/goide/inspections/GoPlaceholderCountInspection.java
dlsniper: Unfortunately it appears as an empty quoted string, ` '' `.

src/com/goide/inspections/GoPlaceholderChecker.java
dlsniper: I was thinking this would also be useful in other use cases of this without the refactoring to move it here then. But sure I can move it.

dlsniper:
I'll let TeamCity check for anything broken and fix it tomorrow morning.

As for the two methods you've suggested moving, I've left them to be used in other places if needed (completion?) and avoid later refactoring and moving things around. If you think I should move them however, I'll do so.

Finally, I'll create issues for the TODOs present in the code once this is fixed.

 Thank you for your patience with this PR.

zolotov:
there is no reason to avoid move-methods refactoring. Actually, you won't avoid it. `resolve` method has nothing to do with placeholder checking and it has a bad name to be public, so as soon as you'll reuse it you'll have to move it in more appropriate place and likely rename. Until then it should be `private`.

Ok, let's keep `hasPlaceholder` in checker. Make it public then

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,440 @@
...
+
+  protected static boolean hasPlaceholder(@NotNull String formatString) {
+    return StringUtil.containsChar(formatString, '%');
+  }
+
+  private static void parseFlags(@NotNull FormatState state) {
+    String knownFlags = "#0+- ";
+    StringBuilder flags = new StringBuilder(state.flags);
+    while (state.nBytes < state.format.length()) {
+      if (knownFlags.indexOf(state.format.charAt(state.nBytes)) != -1) {
```
zolotov: StringUtil.containsChar

zolotov:
Also please fill nullability annotations.

Btw, what about unused methods in checker?

src/com/goide/inspections/GoPlaceholderCountInspection.java
zolotov: It doesn't.

screenshot 2016-06-15 15 00 53

zolotov:
Please consider reporting about particular text range instead of whole string literal

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,256 @@
...
+    }
+
+    return maxArgsNum + firstArg;
+  }
+
+  private static boolean hasErrors(@NotNull ProblemsHolder holder,
+                                   @NotNull GoExpression placeholder,
+                                   @NotNull List<Placeholder> placeholders) {
+
+    for (Placeholder p : placeholders) {
```
zolotov: it's really hard to understand why while traversing `placeholders` you have `placeholder` variable which is not instance of `Placeholder` class. Unreadable, please rename it properly

dlsniper:
> Also please fill nullability annotations.

GoPsiImplUtil.java needs this but it will touch too many things and I'd rather to this outside of this PR otherwise I've ran `Infer nullity` from IDEA on all the other files. Does it matter if I'm still using IDEA 15 for this?

> Please consider reporting about particular text range instead of whole string literal

Oh wow, I've missed that feature, that's really cool! :heart: it!

I've moved and followed the rest of the feedback with regards to visibility and location for methods.

Finally, I've added a bunch more todos and a couple of missing tests.

Thank you.

zolotov:
> GoPsiImplUtil.java needs this but it will touch too many things 

Fill nullability annotations only for code you've added

src/com/goide/inspections/GoPlaceholderCountInspection.java
```diff
@@ -16,170 +16,321 @@
...
-          indexCount = index;
+    String declarationName = declaration.getName();
+    boolean isLn = declarationName != null && declarationName.endsWith("ln");
+    for (GoExpression argument : arguments) {
+      GoType goType = argument.getGoType(null);
+      if (isLn && GoTypeUtil.isString(goType)) {
+        String argText = resolve(argument);
+        if (argText != null && argText.endsWith("\\n")) {
+          String message = "Function already ends with new line #loc";
+          TextRange range = TextRange.create(argText.length() - 1, argText.length() + 1);
```
zolotov: TextRange.from(argText.length() - 1, 2)

dlsniper: I think I'm doing something wrong, if I do that, then the test will fail: `ERROR: Invalid range specified: (5,2); `
What's strange is that TextRange seems to be 1 indexed not 0 indexed but I might be wrong.

dlsniper:
> Fill nullability annotations only for code you've added 

For the single function I've added there it's already correct as for the other files in the PR I've already done it (or I'm missing something as `Infer nullity` doesn't do any further changes in the code base).

Thank you.

src/com/goide/inspections/GoPlaceholderCountInspection.java
zolotov: TextRange is not indexed at all, it's just container for two integers :)

dlsniper:
I'll rebase it when it's ready to merge / reviewed if needed. Thanks.

zolotov:
@dlsniper I'm ready to merge this with minor fixes. What about todos you've placed in code?

zolotov:
@dlsniper also what's the mask in PrintVerb stands for? It's still unused

dlsniper:
For the todos I'll create the according issues, I plan to work on adding the issues after Thursday if that's ok for you. As for PrintVerb, that might be useful in when solving the various todos such as checking the type and see if we can suggest a better one than the current, if it doesn't make sense to have it there for the future (as it's easier to keep it than figure out later how to get it) we can remove it from now, up to you. Should I rebase this first as it has quite a few commits that might not be relevant? Thank you.

dlsniper:
To be clear, I'll on on all the existing inspections as well and if if there's a todo or a missing quickfix, determine what's needed and create the according ticket, not just for this particular issue.

dlsniper:
and rebese means squash, sorry

zolotov:
Ok, thanks, will merge it then.

> Should I rebase this first as it has quite a few commits that might not be relevant? 

As you wish, I can squash all of them into a single commit while merging if you want.

dlsniper:
Done, thank you!

dlsniper:
I'll add the todos / missing quick fixes for this when it's merged, to be consistent with the other inspections.

zolotov:
merged manually

src/com/goide/inspections/GoPlaceholderChecker.java
```diff
@@ -0,0 +1,386 @@
...
+      return false;
+    }
+
+    state.nBytes++;
+    state.argNum = arg;
+
+    return true;
+  }
+
+  private static boolean parseNum(@NotNull FormatState state) {
```
zolotov: why is this always return `true`? and why other code supposes the function can return `false`?

dlsniper: I'll have to look at gofmt again, I thought it's strange as well.

PR #2555, opened 2016-06-06
Add getText() method to GoLightType and GoLightType.LightTypeList

grenki:


src/com/goide/psi/impl/GoLightType.java
```diff
@@ -67,6 +67,11 @@ public GoType getUnderlyingType() {
     return GoPsiImplUtil.getUnderlyingType(this);
   }
 
+  @Override
+  public String getText() {
+    return "";
```
zolotov: ?

src/com/goide/psi/impl/GoLightType.java
```diff
@@ -108,6 +113,15 @@ public LightTypeList(@NotNull GoCompositeElement o, @NotNull List<GoType> types)
     public String toString() {
       return "MyGoTypeList{myTypes=" + myTypes + '}';
     }
+
+    @Override
+    public String getText() {
+      CharSequence[] typeNames = new String[getTypeList().size()];
```
zolotov: test?

src/com/goide/psi/impl/GoLightType.java
```diff
@@ -108,6 +108,15 @@ public LightTypeList(@NotNull GoCompositeElement o, @NotNull List<GoType> types)
...
+      for (int i = 0; i < getTypeList().size(); i++) {
+        typeNames[i] = getTypeList().get(i).getText();
+      }
+      return String.join(", ", typeNames);
```
zolotov: avoid using java 1.8 methods, reimplement with StringUtil

src/com/goide/psi/impl/GoLightType.java
zolotov: see StringUtil#join(java.util.Collection<? extends T>, com.intellij.util.Function<? super T,java.lang.String>, java.lang.String)

PR #2583, opened 2016-06-27
Add receiver names inspection (don't use generics names: me, this, self) (fixes #2158)

### LINKED ISSUE #2158: Don't use me, this, self as receiver names
From https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

```
Don't use generic names such as "me", "this" or "self", identifiers typical of object-oriented languages that
place more emphasis on methods as opposed to functions. 
```

grenki:
Fixes #2158


ignatov:
And the second part
> The name of a method's receiver should be a reflection of its identity; often a one or two letter abbreviation of its type suffices (such as "c" or "cl" for "Client").

zolotov:
Merged manually

PR #2593, opened 2016-07-01
Change GoAnonymousFieldDefinition bnf

grenki:
Add getPointerType method
Remove getTextOffsetMethod


grenki:
wrong destination branch

PR #2594, opened 2016-07-01
Expected type

grenki:
Change getExpectedType
- add is variadic boolean into method result
- fix bug for variadic function
- add tests


grenki:
wrong branch

Issue #2159, opened 2015-12-24
Use consistent receiver names across the methods


dlsniper:
From: https://github.com/golang/go/wiki/CodeReviewComments#receiver-names

```
 Be consistent, too: if you call the receiver "c" in one method, don't call it "cl" in another.
```

PR #2605, opened 2016-07-04
add rename receiver names quickfix

grenki:
fixes issue https://github.com/go-lang-plugin-org/go-lang-idea-plugin/issues/2159

### DIFF
--- a/src/com/goide/inspections/GoReceiverNamesInspection.java
+++ b/src/com/goide/inspections/GoReceiverNamesInspection.java
@@ -16,15 +16,16 @@
 
 package com.goide.inspections;
 
-import com.goide.psi.GoReceiver;
-import com.goide.psi.GoVisitor;
-import com.goide.quickfix.GoRenameQuickFix;
+import com.goide.psi.*;
+import com.goide.quickfix.GoRenameReceiverQuickFix;
 import com.intellij.codeInspection.LocalInspectionToolSession;
 import com.intellij.codeInspection.ProblemsHolder;
 import com.intellij.psi.PsiElement;
+import com.intellij.util.ObjectUtils;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
 
+import java.util.List;
 import java.util.Set;
 
 public class GoReceiverNamesInspection extends GoInspectionBase {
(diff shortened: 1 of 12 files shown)


dlsniper:
A few small things:
- can we split this into two options? Run for all, run for current receiver?
- is it possible to suggest from the existing list of receivers?
- I would not enforce consistency across receiver, especially since it could be tricky to get the name the same in other parts of the code due to the "rather use single letters" recommendations
- finally, since this changes a bit the detection, do you think we should test against the not exported types as well? mainly because if we have this:
```go
type demo struct{}
type myDemo struct

func (demo<caret> demo) Hello() {}
func (myDemo myDemo) World() {}
```
that will look really confusing for users and I'd rather have that renamed as ` d ` or ` md ` in the second case.

Thank you.

src/com/goide/inspections/GoReceiverNamesInspection.java
```diff
@@ -39,7 +40,37 @@ public void visitReceiver(@NotNull GoReceiver o) {
...
           if (identifier == null) return;
-          holder.registerProblem(identifier, "Receiver has generic name", new GoRenameQuickFix(o));
+          holder.registerProblem(identifier, "Receiver has generic name", new GoRenameReceiverQuickFix(identifier, false),
+                                 new GoRenameReceiverQuickFix(identifier, true));
+        }
+      }
+
+      @Override
+
+      public void visitMethodDeclaration(@NotNull GoMethodDeclaration o) {
```
ignatov: I suppose we should visit TypeSpecs, not methods, otherwise that'll be O(n^2).

src/com/goide/inspections/GoReceiverNamesInspection.java
```diff
@@ -39,7 +40,37 @@ public void visitReceiver(@NotNull GoReceiver o) {
...
+        GoType type = methodReceiver.getType();
+        type = type instanceof GoPointerType ? ((GoPointerType)type).getType() : type;
+        if (type == null) return;
+        GoTypeReferenceExpression reference = type.getTypeReferenceExpression();
+        if (reference == null) return;
+        GoTypeSpec typeSpec = ObjectUtils.tryCast(reference.resolve(), GoTypeSpec.class);
+        if (typeSpec == null) return;
+        List<GoMethodDeclaration> methods = typeSpec.getMethods();
+        for (GoMethodDeclaration method : methods) {
+          if (!file.isEquivalentTo(method.getContainingFile())) continue;
```
ignatov: We should check all methods for the type and rename their receivers if needed.

zolotov:
any updates on this?

PR #2641, opened 2016-07-10
Added types, consts and methods check for reserved word inspection

dlsniper:


tests/com/goide/inspections/GoHighlightingTest.java
```diff
@@ -295,18 +296,6 @@ public void testMainInMainPackage() {
     myFixture.checkHighlighting();
   }
 
-  public void testDuplicateBuiltinFunction() {
```
zolotov: ?

dlsniper: I've removed the two separate tests and created a file for it, I don't know why I haven't done that the first time around.

src/com/goide/inspections/GoReservedWordUsedAsNameInspection.java
```diff
@@ -34,57 +35,56 @@
...
-    GoFile builtin = GoSdkUtil.findBuiltinFile(variable);
+  private static void check(@NotNull GoNamedElement element, @NotNull ProblemsHolder holder) {
+    ProgressManager.checkCanceled();
+    GoFile builtin = GoSdkUtil.findBuiltinFile(element);
     if (builtin == null) return;
 
-    String name = variable.getName();
-    if (name == null) return;
+    String name = element.getName();
+    if (name == null || "Type".equals(name)) return;
```
zolotov: Where does `Type` name come from? I also don't see test for it

testData/highlighting/struct.go
```diff
@@ -31,7 +31,7 @@ func (e E) foo() {
 
 type Type int
 
-func <warning descr="Function 'new' collides with builtin function">new</warning>(Type) *Type
+func new(Type) *Type
```
zolotov: Why is this not a warning anymore?

src/com/goide/inspections/GoReservedWordUsedAsNameInspection.java
dlsniper: `Type` is there because it's a rather special case in the `builtin.go` file. However, I'm starting to to think a different approach should be considered since there are other cases as well.

See: https://github.com/golang/go/blob/master/src/builtin/builtin.go#L105

testData/highlighting/struct.go
dlsniper: I'll need to rework it, it should still be.

zolotov:
Also please check what will happen if you open `builtin` file

zolotov:
And clarify in commit message what's the improvement you've actually done

src/com/goide/inspections/GoReservedWordUsedAsNameInspection.java
zolotov: `Type` is not the only case. See `com.goide.psi.impl.GoTypeReference#BUILTIN_TYPE`

dlsniper:
I've updated this as requested in the feedback.
I've checked the `builtin.go` file a few times, even edited it from the IDE, the inspection does not seem to be triggered on that file at all and I'm not sure why.

zolotov:
merged manually

Issue #2147, opened 2015-12-24
Non-interface type types don't inherit the existing methods


dlsniper:
```go
package main

import "sync"

type myMutex sync.Mutex

func main() {
	var mtx myMutex
	mtx.Lock()   // mtx.Lock undefined (type myMutex has no field or method Lock)
	mtx.Unlock() // mtx.Unlock undefined (type myMutex has no field or method Unlock)
}
```

pbennett:
I see the exact same problem even with direct use of sync.Mutex:
var (
   pathLock sync.RWMutex
...
)

func xxxx() {
   pathLock.RLock()    <-- marked as undefined
   defer PathLock.RUnlock()  <--  same
...

dlsniper:
This was resolved (at least it works in the nightly) but the message is: Unresolved reference rather than the one I've posted. I've checked the comment from @pbennett and that's fixed as well.

@zolotov / @ignatov should we close this?

ignatov:
@dlsniper Please add a test for this issue.

dlsniper:
@ignatov sure, but should I leave the text of "Unresolved reference" or should I send them as the compiler does now with more explicit output? If the later I could try and see if I can add the text change not just send the text. Thanks.

ignatov:
"Unresolved reference" is fine.

PR #2644, opened 2016-07-10
Add tests for #2147

dlsniper:
As discussed in the issue, there should be some tests to cover it.

### DIFF
--- a/testData/highlighting/gh2147.go
+++ b/testData/highlighting/gh2147.go
@@ -1,0 +1,20 @@
+package main
+
+import "sync"
+
+type myMutex sync.Mutex
+
+var (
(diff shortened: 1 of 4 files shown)


zolotov:
@dlsniper please give sensible names to tests. If this test failed I wouldn't know why it shouldn't because it name tells nothing.
@ignatov please do not merge PRs which isn't passed tests

dlsniper:
This was green on my machine, I don't know why it failed, from the error message it looks like a problem with Gradle. I've seen that a few times so I'm not sure what's the best to do in these cases. Should I retry the build?

zolotov:
@dlsniper no, it was gradle problem, fixed in master

PR #2674, opened 2016-07-16
Added package updating functionality (fixes #2671)

### LINKED ISSUE #2671: Update package intentention
Will be cool to have an intention to update already existing package, like go get -u.
Like download missing package.

dlsniper:


dlsniper:
The only item that I'm not sure how to best do it: how to determine if an import is from an SDK or not. Should I generate a list of all packages in GOROOT? Should I try to see if $GOROOT/src/packgeImportPath exists?

dlsniper:
Also, should I include an "Update all" functionality? as it would need to extend the current one mostly.

src/com/goide/intentions/GoUpdatePackage.java
```diff
@@ -0,0 +1,95 @@
...
+import com.intellij.openapi.vfs.VirtualFileManager;
+import com.intellij.psi.PsiElement;
+import com.intellij.psi.PsiFile;
+import com.intellij.util.Consumer;
+import com.intellij.util.IncorrectOperationException;
+import org.jetbrains.annotations.Nls;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class GoUpdatePackage implements IntentionAction{
```
zolotov: code style

src/com/goide/psi/impl/GoPsiImplUtil.java
```diff
@@ -1130,6 +1130,15 @@ public static boolean isCImport(@NotNull GoImportSpec importSpec) {
     return GoConstants.C_PATH.equals(importSpec.getPath());
   }
 
+  public static boolean isRelativeImport(@NotNull GoImportSpec importSpec) {
+    return GoImportReferenceSet.isRelativeImport(importSpec.getPath());
```
zolotov: method looks sense
less

src/com/goide/intentions/GoUpdatePackage.java
```diff
@@ -0,0 +1,95 @@
...
+    if (!(elementAt.getParent() instanceof GoImportString)) return null;
+
+    GoImportString importString = (GoImportString)elementAt.getParent();
+    if (!(importString.getParent() instanceof GoImportSpec)) return null;
+
+    GoImportSpec importSpec = (GoImportSpec)importString.getParent();
+    if (importSpec.isCImport()) return null;
+    if (importSpec.isRelativeImport() || importSpec.isSDKImport()) return null;
+
+    return importString.getPath();
```
zolotov: should not be available for unresolved imports

src/com/goide/intentions/GoUpdatePackage.java
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  @Override
+  public boolean startInWriteAction() {
+    return false;
+  }
+
+  @Nullable
+  private static String getPackageName(Editor editor, PsiFile file) {
+    PsiElement elementAt = file.findElementAt(editor.getCaretModel().getOffset());
```
zolotov: PsiElementBaseIntentionAction

zolotov:
> Should I generate a list of all packages in GOROOT?
> Should I try to see if $GOROOT/src/packgeImportPath exists?

No, this path/package can be shadowed by vendor package. You must resolve import and check that directory is under GOROOT for current module

src/com/goide/psi/impl/GoPsiImplUtil.java
dlsniper: The idea was to add the method to the ImportSpec so that it's usable in that context (and to ask if I should move the `GoImportReferenceSet.isRelativeImport` to `GoPsiUtil`. I guess it's not a good idea? :D

zolotov: Well, creating method for PsiElement which just invoke some other reference-specific method is strange.

dlsniper:
I've updated and made this work, thank you for the guidance on this.
Final question, should I write some tests for this? And if so, can you please point me to an example? Thank you.

zolotov:
> Final question, should I write some tests for this?

Yes, test availability of intention in different cases

gen/com/goide/psi/GoImportSpec.java
```diff
@@ -51,4 +51,6 @@
 
   boolean isCImport();
 
+  boolean isSDKImport();
```
zolotov: We don't use abbreviations in tests

screenshot 2016-07-17 20 10 35

zolotov: 1. isSdkImport is not a property of import, so it shouldn't be instance method of ImportSpec. isCImport is a property because it depends on import spec and nothing more. Sdk- or not-sdk- import depends on module, settings of module, project, current context and many other things.

2. it's not good to introduce some method based on copy-paste from `GoPathScopeHelper` and do not patch `GoPathScopeHelper` itself.

dlsniper: I'll need to reread the GoPathScopeHelper again as I don't understand how it can be used to check if an import refers to the Sdk or not.

zolotov: It shouldn't be used, but it check import too. It means that you duplicate code by inplementing this method, this is unacceptable

zolotov:
There are plenty tests which look for available intention actions + obviously it should be part of `fixture`-classes

PR #2698, opened 2016-07-30
Rework struct initialization inspection (fixes #2687)

### LINKED ISSUE #2678: Test status is reported incorrectly if output contains "--- FAIL"
- Plugin version (or commit hash): 0.10.1679
- IDE name and version: IntelliJ IDEA 15 Community Edition 15.0.6
- Java version: 1.8.0_65
- OS name and version: OS X El Capitan 10.11.5

- What are you trying to do?
Run tests that contain a `fmt.Printf` or `t.Logf` statement that prints `--- FAIL:`

- What would you expect to happen?
Tests run, print the specified output, complete, and are marked as passing.

- What happens?
Lines that start with `--- FAIL:` are not shown as part of the test output, and the string that comes after `FAIL:` is reported as a failing test, which causes the tests to show as failed.

When the same test is run using `go test -v`, the output and reporting is correct.

```go
package badtest

import (
	"fmt"
	"testing"
)

func TestPrintfOkay(t *testing.T) {
	// this line is okay -- test passes and prints output
	fmt.Printf("--- FAIL: TestRun (from fmt.Printf)")
}

func TestPrintfBad(t *testing.T) {
	// this causes IntelliJ to report that "PhantomTest" has failed, and output is not available anywhere
	fmt.Printf("--- FAIL: PhantomTest (from fmt.Printf)\n")
}

func TestLogf(t *testing.T) {
	// this causes IntelliJ to report that "NonexistentTest" has failed, and output is not available anywhere
	t.Logf("--- FAIL: NonexistentTest (from t.Logf)")
}
```

Tests reported as failing in IntelliJ:
intellij_fail

Tests passing with all output using `go test -v`:
gotest_pass

### LINKED ISSUE #2687: Incorrect warning "Unnamed field initialization" on nested slices with structs
- Plugin version (or commit hash): 0.12.1724
- IDE name and version: IntelliJ IDEA 2016.2 Build #IC-162.1121.32, built on July 9, 2016
- Java version: 1.8.0_77-b03 amd64
- OS name and version: Windows 10

- What are you trying to do?
Initialize a slice of type struct, which contains another slice with type struct. I use named parameters but I still get a warning.

Example code
https://play.golang.org/p/CxBRjv9c90

- What would you expect to happen?
No warning.

- What happens?
There is a warning. When I use the quickfix he updates l18 to
`foo: {foo: "foo"}, // This one gives the "Unnamed field initialization"`
which gives on running
> prog.go:18: undefined: foo

dlsniper:


src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -35,60 +35,50 @@
 
 public class GoStructInitializationInspection extends GoInspectionBase {
   public static final String REPLACE_WITH_NAMED_STRUCT_FIELD_FIX_NAME = "Replace with named struct field";
-  public boolean reportImportedStructs;
+  public boolean reportLocalStructs;
```
dlsniper: I can do the inspection without changing this if it's a breaking change for the users.

zolotov: yes please, also fix commit message – it references wrong issue

dlsniper: Done, good catch on issue ID, sorry.

dlsniper:
I don't understand why this failed, seems unrelated to the current changes. Should I look into it?

zolotov:
Not sure what happened but it compiles now

zolotov:
merged manually

PR #2699, opened 2016-07-30
Introduce inspection to detect imported package names as identifier names

dlsniper:


zolotov:
merged manually, thanks

zolotov:
@dlsniper could you remind me what was the motivation of this inspection please? When is this useful?

dlsniper:
Accidentally shadow a package name as then you can't use that for the rest of the scope you do it. There are users / code which does this and then users are baffled by why they can't use certain package functions or other exported identifiers.

zolotov:
Ok, thanks

PR #2734, opened 2016-08-12
Fix keywords repeated(#fixes 2733)

bingohuang:
Fix `KEYWORDS` attribute(`TokenSet` type) in `GoParserDefinition.java` file has 27 keywords which have two repeating keywords `PACKAGE` and `IMPORT`.


zolotov:
these two keywords were placed in alphabetical order, so it's quite logical to remove two others instead in order to preserve the order.

bingohuang:
Get it and I fix it. Thanks.

PR #2738, opened 2016-08-14
Improve receiver assignment inspection (fixes #2719)

### LINKED ISSUE #2719: Incorrect inspection for assignment to method receiver
Issue #2429 introduced new type of inspection, "Assignment to method receiver propagates only to callees but not to callers". However it seems that it doesn't work correctly when dereferencing pointer receiver to reassign underlying value.

Consider the following example:

``` go
package main

import "fmt"

type Flags []string

func (f *Flags) Add(flag string) {
    *f = append(*f, flag)
}

func main() {
    var f Flags
    f.Add("a")
    fmt.Println(f)
}
```

It seems to be valid use of pointer receiver. But the inspection is triggered here.

dlsniper:
Currently the only annoyance this produces is when the receiver is not returned from the method.

For example:

``` go
type Flagz []string

func (f Flagz) Add(flag string) Flagz {
    f = append(f, flag) // still a warning, shouldn't be
    return f
}
```

Short from walking the PSI tree and checking if the receiver is returned, how could this be done? Or should we disable this by default?

I've also moved the inspection from ReferenceExpression to Assignment in hopes this will be faster to use.

### DIFF
--- a/src/com/goide/inspections/GoAssignmentToReceiverInspection.java
+++ b/src/com/goide/inspections/GoAssignmentToReceiverInspection.java
@@ -16,15 +16,12 @@
 
 package com.goide.inspections;
 
-import com.goide.psi.GoPointerType;
-import com.goide.psi.GoReceiver;
-import com.goide.psi.GoReferenceExpression;
-import com.goide.psi.GoVisitor;
-import com.intellij.codeInsight.highlighting.ReadWriteAccessDetector;
+import com.goide.psi.*;
 import com.intellij.codeInspection.LocalInspectionToolSession;
 import com.intellij.codeInspection.ProblemsHolder;
 import com.intellij.psi.PsiElement;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 import static com.intellij.codeInspection.ProblemHighlightType.WEAK_WARNING;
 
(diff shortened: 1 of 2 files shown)


dlsniper:
There might be something wrong with the build agent as the build error does not appear on my setup and it's not related the sources I've changed.

src/com/goide/inspections/GoAssignmentToReceiverInspection.java
```diff
@@ -34,19 +31,45 @@
...
+    if (resolveType instanceof GoPointerType) {
+      if (parent instanceof GoUnaryExpr) {
+        if (((GoUnaryExpr)parent).getMul() != null) return false;
+      }
+    }
+
+    GoType underlyingType = resolveType.getUnderlyingType();
+    if (underlyingType instanceof GoChannelType ||
+        underlyingType instanceof GoArrayOrSliceType //||
+        //underlyingType instanceof GoMapType
```
zolotov: ?

src/com/goide/inspections/GoAssignmentToReceiverInspection.java
```diff
@@ -34,19 +31,45 @@
...
+    PsiElement parent = o.getParent();
+
+    if (resolveType instanceof GoPointerType) {
+      if (parent instanceof GoUnaryExpr) {
+        if (((GoUnaryExpr)parent).getMul() != null) return false;
+      }
+    }
+
+    GoType underlyingType = resolveType.getUnderlyingType();
+    if (underlyingType instanceof GoChannelType ||
```
zolotov: `if () { return false } else { return true }` looks strange

src/com/goide/inspections/GoAssignmentToReceiverInspection.java
dlsniper: Ooops, forgot to uncomment it, it is covered by a test.

src/com/goide/inspections/GoAssignmentToReceiverInspection.java
dlsniper: I've replaced it with `return !(() || () || ())` but not sure if it's preferred to `return (!() && !() && !())`

PR #2774, opened 2016-09-09
163.3983 EAP support

dlsniper:


zolotov:
Thank you. I completely forgot about it

dlsniper:
Can you also please push a nightly after TC finishes building it?
I've tried to build it on my machine but I get a:

```
FAILURE: Build failed with an exception.

* What went wrong:
Could not read entry ':prepareSandbox' from cache taskArtifacts.bin (/home/florin/IdeaProjects/google-go-lang-idea-plugin/.gradle/3.0-milestone-2/taskArtifacts/taskArtifacts.bin).
> org.jetbrains.intellij.dependency.PluginDependency; local class incompatible: stream classdesc serialVersionUID = 2952204855539725202, local class serialVersionUID = 560590560477762714
```

zolotov:
> Can you also please push a nightly after TC finishes building it?

sure

> Could not read entry ':prepareSandbox' from cache taskArtifacts.bin (/home/florin/IdeaProjects/google-go-lang-idea-plugin/.gradle/3.0-milestone-2/taskArtifacts/taskArtifacts.bin).

So remove the cache

dlsniper:
Thanks, I was doing that just now

PR #2826, opened 2016-11-13
Rework GoStructInitializationInspection

### LINKED ISSUE #2819: Highlight the whole struct literal when unnamed element is present
I think that current logic of `GoStructInitializationInspection` could be better: inspection suggests to replace single unnamed element by named, but we have to name all element in the struct literal to get valid code.

So i propose highlight the whole struct literal when at least one unnamed element is present. Quickfix will add names to all unnamed elements.

P.S. Current implementation misses case when several fields in struct declared by one type (see `image.Point`) as example.

wbars:
fixes #2819


zolotov:
It's not really good idea to rename an inspection at least for two reasons:

1. I'd prefer to save vcs-history and git-blame data
2. The inspection is quite old and some users might have suppressed it with a comment `//noinspection GoStructInitialization`. The thing is that you changed the suppression id by renaming the class and old suppressions won't work anymore. The same thing for settings – some users could disable the inspection but since its ID was changed – the settings became invalid.

So please avoid inspections renaming until you have a really good reason

zolotov:
@grenki please review the changes

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,131 @@
...
+      List<GoElement> elements = structType != null ? literal.getElementList() : emptyList();
+      List<String> fieldDefinitionNames = structType != null ? getFieldDefinitionsNames(structType) : emptyList();
+      if (!areElementsNamesMatchesDefinitions(getNames(elements), fieldDefinitionNames)) return;
+      replaceElementsByNamed(elements, fieldDefinitionNames, project);
+    }
+  }
+
+  private static void replaceElementsByNamed(@NotNull List<GoElement> elements,
+                                             @NotNull List<String> fieldDefinitionNames,
+                                             @NotNull Project project) {
```
grenki: Place project as first parameter, please

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,131 @@
...
-      myStructField = structField;
     }
 
     @Override
     public void applyFix(@NotNull Project project, @NotNull ProblemDescriptor descriptor) {
-      PsiElement startElement = descriptor.getStartElement();
-      if (startElement instanceof GoElement) {
-        startElement.replace(GoElementFactory.createLiteralValueElement(project, myStructField, startElement.getText()));
-      }
+      GoLiteralValue literal = ObjectUtils.tryCast(descriptor.getStartElement(), GoLiteralValue.class);
```
grenki: startElement can be null and invalid. You could check it

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,131 @@
...
-        if (refType instanceof GoStructType) {
-          processStructType(holder, o, (GoStructType)refType);
+      public void visitCompositeLit(@NotNull GoCompositeLit compositeLit) {
+        GoLiteralValue literalValue = compositeLit.getLiteralValue();
+        GoStructType structType = getStructType(literalValue);
+        if (structType == null || !isStructImportedOrLocalAllowed(structType, literalValue)) return;
+
+        List<String> elementsNames = getNames(literalValue.getElementList());
+        if (hasNull(elementsNames.toArray()) && areElementsNamesMatchesDefinitions(elementsNames, getFieldDefinitionsNames(structType))) {
+          holder.registerProblem(literalValue, "Unnamed field initializations", ProblemHighlightType.GENERIC_ERROR_OR_WARNING, QUICK_FIX);
```
grenki: I think better report problem for each element and it could be weak warning, because it can be valid code.

wbars: Actually, reporting the literal was one of the main point. Why do you think that report each element is better? We have to name all elements in literal to get valid code anyway, so what is the point in renaming one element.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,131 @@
...
+  private static List<String> getNames(@NotNull List<GoElement> elements) {
+    return map(elements, element -> {
+      GoKey key = element.getKey();
+      return key != null ? key.getText() : null;
+    });
+  }
+
+  private static boolean areElementsNamesMatchesDefinitions(@NotNull List<String> elementsNames,
+                                                            @NotNull List<String> fieldDefinitionsNames) {
+    return range(0, elementsNames.size()).allMatch(i -> isNullOrEqual(elementsNames.get(i), getByIndex(fieldDefinitionsNames, i)));
```
wbars: Method is private and i don't see it as part of GoPsiImpl interface cause it's not related to psi operations.

src/com/goide/inspections/GoStructInitializationInspection.java
wbars: And i believe that if we have mixed named and unnamed elements in the literal code [will not compile at all](https://golang.org/ref/spec#Composite_literals).

src/com/goide/inspections/GoStructInitializationInspection.java
grenki: I think we must have public util method like that. But while we haven't suitable util class it can be in `GoPsiImplUtli`

src/com/goide/inspections/GoStructInitializationInspection.java
grenki: Reporting each element without keys 
 - shows user only error element,
 - it doesn't mean that quickfix must rename only this element 
 -  big part of code which highlights as error looks bad :(
Anyway we mustn't highlight as error valid code like this: https://play.golang.org/p/4cP6PcSWCZ

And if you want to change problem highlighting type you could change level of inspection in gogland.xml

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,131 @@
...
       @Override
-      public void visitLiteralValue(@NotNull GoLiteralValue o) {
-        if (PsiTreeUtil.getParentOfType(o, GoReturnStatement.class, GoShortVarDeclaration.class, GoAssignmentStatement.class) == null) {
-          return;
-        }
-        PsiElement parent = o.getParent();
-        GoType refType = GoPsiImplUtil.getLiteralType(parent, false);
-        if (refType instanceof GoStructType) {
-          processStructType(holder, o, (GoStructType)refType);
+      public void visitCompositeLit(@NotNull GoCompositeLit compositeLit) {
```
grenki: And you could check literal value. If you look to go.bnf you cam see that literal value not always has composite literal parent.

For example it's valid code 
```
type S struct{int}
_ = []S{ {1}, {2}, {3}} // parent of literal value {1} is GoValue {1}
```

src/com/goide/inspections/GoStructInitializationInspection.java
wbars: For me as user it would be a little strange, that invoke fix on element affect the parent, but if you fine with it, I don't mind.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,148 @@
...
+           : ObjectUtils.tryCast(GoPsiImplUtil.getLiteralType(lit, false), GoStructType.class);
   }
 
-  private void processStructType(@NotNull ProblemsHolder holder, @NotNull GoLiteralValue element, @NotNull GoStructType structType) {
-    if (reportLocalStructs || !GoUtil.inSamePackage(structType.getContainingFile(), element.getContainingFile())) {
-      processLiteralValue(holder, element, structType.getFieldDeclarationList());
-    }
+  @Contract("_, null -> false; null, _ -> false")
+  private static boolean isSliceElementValue(@Nullable GoLiteralValue literalValue, @Nullable GoType parentLitType) {
+    return parentLitType instanceof GoArrayOrSliceType && literalValue != null && literalValue.getParent() instanceof GoValue;
```
wbars: For this operation (getting parent from nullable) I would like to move `GoStructLiteralCompletion#parent` to `GoPsiTreeUtil` since there are already plenty occurrences of this code in the plugin project, but I have some doubts. 
There are already numerous methods for getting parent and, in particular, `PsiTreeUtil#findFirstParent` which, as I think, can become confusing with the new one. What do you think @grenki?

wbars:
Ping :)

zolotov:
@wbars sorry for delay. Review is in progress. @grenki right?

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,148 @@
...
-  private static String getFieldName(@NotNull GoFieldDeclaration declaration) {
-    List<GoFieldDefinition> list = declaration.getFieldDefinitionList();
-    GoFieldDefinition fieldDefinition = ContainerUtil.getFirstItem(list);
-    return fieldDefinition != null ? fieldDefinition.getIdentifier().getText() : null;
+  private boolean isStructImportedOrLocalAllowed(@NotNull GoStructType structType, @NotNull GoLiteralValue literalValue) {
+    return reportLocalStructs || !GoUtil.inSamePackage(structType.getContainingFile(), literalValue.getContainingFile());
+  }
+
+  @NotNull
+  private static List<String> getNames(@NotNull List<GoElement> elements) {
```
grenki: it isn't names :)

src/com/goide/inspections/GoStructInitializationInspection.java
grenki: It's great idea! It very useful and convenient method.

`PsiTreeUtil#findFirstParent` and `PsiTreeUtil#getParentOfType` are hard to use in this case

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,148 @@
...
+  private static void replaceElementsByNamed(@NotNull Project project,
+                                             @NotNull List<GoElement> elements,
+                                             @NotNull List<String> fieldDefinitionNames) {
+    for (int i = 0; i < elements.size(); i++) {
+      GoElement element = elements.get(i);
+      String fieldDefinitionName = GoPsiImplUtil.getByIndex(fieldDefinitionNames, i);
+      GoValue value = fieldDefinitionName != null && element.getKey() == null ? element.getValue() : null;
+      if (value == null) continue;
+
+      GoElement namedElement = GoElementFactory.createLiteralValueElement(project, fieldDefinitionName, value.getText());
```
grenki: you can inline `namedElement` 
it is element with key, not named

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -55,23 +56,42 @@ protected GoVisitor buildGoVisitor(@NotNull ProblemsHolder holder, @NotNull Loca
...
   }
 
   @Nullable
   @Contract("null -> null")
-  private static GoStructType getStructType(@Nullable GoLiteralValue literal) {
-    return literal != null ? ObjectUtils.tryCast(GoPsiImplUtil.getLiteralType(literal.getParent(), false), GoStructType.class) : null;
+  private static GoStructType getStructType(@Nullable GoLiteralValue literalValue) {
+    GoCompositeLit lit = PsiTreeUtil.getParentOfType(literalValue, GoCompositeLit.class);
+    GoType litType = lit != null ? lit.getType() : null;
+    return isSliceElementValue(literalValue, litType)
```
wbars: Example does not compile (should be `S: S{2, 3<caret>}` I guess), but fixed version works. Maybe we should not highlight such cases at all? (when there is no composite literal for literal value).

grenki: Sorry for invalid example. I mean something like this
```
type S struct {
	r, t int
}
var _ = [][]S{
	{
		{2,   2},
	},
}


var _ = map[string] []S{
	"s": {
		{1, 2},
	},
}
```

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,148 @@
...
+                                             @NotNull List<GoElement> elements,
+                                             @NotNull List<String> fieldDefinitionNames) {
+    for (int i = 0; i < elements.size(); i++) {
+      GoElement element = elements.get(i);
+      String fieldDefinitionName = GoPsiImplUtil.getByIndex(fieldDefinitionNames, i);
+      GoValue value = fieldDefinitionName != null && element.getKey() == null ? element.getValue() : null;
+      if (value == null) continue;
+
+      GoElement namedElement = GoElementFactory.createLiteralValueElement(project, fieldDefinitionName, value.getText());
+      element.replace(namedElement);
```
wbars: Yes, indeed. But I could not find a way to add colon and key to the element without several boilerplate lines of code so I would like to stay with the current solution.

wbars:
PTAL

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,180 @@
...
+        registerProblemsForElementsWithoutKeys(elements, keys, holder);
       }
     };
   }
 
-  @Override
-  public JComponent createOptionsPanel() {
-    return new SingleCheckboxOptionsPanel("Report for local type definitions as well", this, "reportLocalStructs");
+  @Contract("null -> null")
+  private static GoStructType getLiteralStructType(@Nullable GoLiteralValue literalValue) {
```
grenki: this example is invalid
```
type S struct {
	r, t int
}

type B struct{
	S
}

var  _  = []B{  S: {2,  3}}

```

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,180 @@
...
+    String definitionName = getFieldDefinitionName(GoPsiTreeUtil.getDirectParentOfType(literalValue, GoValue.class));
+    return definitionName != null && litType != null ? getFieldDefinitionType(litType, definitionName) : litType;
   }
 
-  private void processStructType(@NotNull ProblemsHolder holder, @NotNull GoLiteralValue element, @NotNull GoStructType structType) {
-    if (reportLocalStructs || !GoUtil.inSamePackage(structType.getContainingFile(), element.getContainingFile())) {
-      processLiteralValue(holder, element, structType.getFieldDeclarationList());
-    }
+  @Nullable
+  private static String getFieldDefinitionName(@Nullable GoValue value) {
```
grenki: better to return named element in which key resolves: 
```
PsiElement field = fieldName != null ? fieldName.resolve() : null;
return field instanceof GoAnonymousFieldDefinition || field instanceof GoFieldDefinition
           ? ObjectUtils.tryCast(field, GoNamedElement.class) : null;
```

if you return field definition you wouldn't use `getDefinition` method in `getFieldDefinitionType`

tests/com/goide/inspections/GoStructInitializationInspectionTest.java
```diff
@@ -38,26 +38,44 @@ protected void tearDown() throws Exception {
...
+  public void testInnerStruct()                     { doQuickfixTest(); }
+  public void testOnelineQuickfix()                 { doQuickfixTest(); }
+  public void testVarDeclaration()                  { doQuickfixTest(); }
+  public void testNestedLiteral()                   { doQuickfixTest(); }
+  public void testInnerLiteral()                    { doQuickfixTest(); }
+  public void testInnerAnonLiteral()                { doQuickfixTest(); }
+  public void testInnerLiteralMap()                 { doQuickfixTest(); }
+  public void testInnerLiteralFieldWithKey()        { doQuickfixTest(); }
+  public void testMultipleInnerLiterals()           { doQuickfixTest(); }
+  public void testInnerLiteralWithoutType()           { doQuickfixTest(); }
```
grenki: alignment

testData/inspections/struct-initialization/literalValueWithoutCompositeLit.go
```diff
@@ -0,0 +1,13 @@
...
+type S struct {
+	t int
+}
+
+type B struct{
+	S
+}
+
+func main() {
+	var  _  = []B{ S: {<weak_warning descr="Unnamed field initializations">1<caret></weak_warning>}}
```
grenki: invalid test:
```
[]B {
 // S is invalid key because in array literals we use integer keys (indicies) 
    S : {  // expected literal type is B (parent literal has []B type)
           1 // expected key is `S` (type is B)
     }, 
}
```

testData/inspections/struct-initialization/uninitializedStructImportedOnly.go
```diff
@@ -114,13 +114,13 @@ func _() {
...
 	_ = io.LimitedReader{
-		<weak_warning descr="Unnamed field initialization">nil</weak_warning>,
+	<weak_warning descr="Unnamed field initializations">nil</weak_warning>,
 	}
 	_ = os.LinkError{
-		<weak_warning descr="Unnamed field initialization">"string"</weak_warning>,
-		<weak_warning descr="Unnamed field initialization">"string"</weak_warning>,
-		<weak_warning descr="Unnamed field initialization">"string"</weak_warning>,
-		<weak_warning descr="Unnamed field initialization">nil</weak_warning>,
+		<weak_warning descr="Unnamed field initializations">"string"</weak_warning>,
```
grenki: `initialization` is better

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,180 @@
...
+    GoType type = parentLit.getGoType(null);
+    return type != null && GoPsiImplUtil.getUnderlyingType(type) instanceof GoStructType;
+  }
+
+  private boolean isStructImportedOrLocalAllowed(@NotNull GoStructType structType, @NotNull GoLiteralValue literalValue) {
+    return reportLocalStructs || !GoUtil.inSamePackage(structType.getContainingFile(), literalValue.getContainingFile());
+  }
+
+  @NotNull
+  private static List<String> getKeys(@NotNull List<GoElement> elements) {
```
grenki: Better to use fields in which keys resolved instead names.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,180 @@
...
+    return range(0, keys.size()).allMatch(i -> isNullOrEqual(keys.get(i), GoPsiImplUtil.getByIndex(fieldDefinitionsNames, i)));
+  }
+
+  @Contract("null, _ -> true")
+  private static boolean isNullOrEqual(@Nullable Object o, @Nullable Object objectToCompare) {
+    return o == null || Comparing.equal(o, objectToCompare);
+  }
+
+  @NotNull
+  private static List<String> getFieldDefinitionsNames(@Nullable GoStructType type) {
```
grenki: better return fields objects instead names

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -22,94 +22,180 @@
...
   }
 
   @Nullable
-  private static String getFieldName(@NotNull GoFieldDeclaration declaration) {
-    List<GoFieldDefinition> list = declaration.getFieldDefinitionList();
-    GoFieldDefinition fieldDefinition = ContainerUtil.getFirstItem(list);
-    return fieldDefinition != null ? fieldDefinition.getIdentifier().getText() : null;
+  private static GoType getUnderlyingType(@NotNull GoFieldDefinition fieldDefinition) {
+    GoType type = fieldDefinition.getGoType(null);
+    return type != null ? GoPsiImplUtil.getUnderlyingType(type) : null;
```
grenki: `type.getUnderlyingType()`

wbars:
Force updated.

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -49,67 +56,114 @@
...
+      .flatMap(declaration -> getFieldDefinitions(declaration).stream())
+      .collect(toList()) : emptyList();
+  }
+
+  @NotNull
+  private static List<GoNamedElement> getFieldDefinitions(@NotNull GoFieldDeclaration declaration) {
+    GoNamedElement anonymousDefinition = ObjectUtils.tryCast(declaration.getAnonymousFieldDefinition(), GoNamedElement.class);
+    return anonymousDefinition != null
+           ? list(anonymousDefinition)
+           : map(declaration.getFieldDefinitionList(), definition -> ObjectUtils.tryCast(definition, GoNamedElement.class));
```
grenki: you can return `List<? extends GonamedElement>` and don't cast definittions

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -49,67 +56,114 @@
...
-        holder.registerProblem(element, "Unnamed field initialization", ProblemHighlightType.GENERIC_ERROR_OR_WARNING, fixes);
-      }
-    }
+  @Nullable
+  private static GoNamedElement getFieldDefinition(@Nullable GoValue value) {
+    GoKey key = PsiTreeUtil.getPrevSiblingOfType(value, GoKey.class);
+    GoFieldName fieldName = key != null ? key.getFieldName() : null;
+    PsiElement field = fieldName != null ? fieldName.resolve() : null;
+    return field instanceof GoAnonymousFieldDefinition || field instanceof GoFieldDefinition ? ObjectUtils
+      .tryCast(field, GoNamedElement.class) : null;
```
grenki: bad formatting

src/com/goide/inspections/GoStructInitializationInspection.java
```diff
@@ -49,67 +56,114 @@
...
-                                                        : LocalQuickFix.EMPTY_ARRAY;
-        holder.registerProblem(element, "Unnamed field initialization", ProblemHighlightType.GENERIC_ERROR_OR_WARNING, fixes);
-      }
-    }
+  @Nullable
+  private static GoNamedElement getFieldDefinition(@Nullable GoValue value) {
+    GoKey key = PsiTreeUtil.getPrevSiblingOfType(value, GoKey.class);
+    GoFieldName fieldName = key != null ? key.getFieldName() : null;
+    PsiElement field = fieldName != null ? fieldName.resolve() : null;
+    return field instanceof GoAnonymousFieldDefinition || field instanceof GoFieldDefinition ? ObjectUtils
```
grenki: extract to util class and reuse `GoMoveToStructInitializationIntention#isFieldDefinition`

wbars:
PTAL

PR #2858, opened 2017-01-07
Master

easyihave:


ignatov:
Sorry, but what's this?

dlsniper:
@ignatov I think I might be a mistake like we've had in the past. I'll close it for now as it can't be used as it ia anyway.

PR #2863, opened 2017-01-17
Update SDK in External Libraries tree

### LINKED ISSUE #2861: Can not update External Libraries tree after change project SDK
- Plugin version (or commit hash): 610ff0ef4e8a55690b9a480efe725c5776e341f8
- IDE name and version: IntelliJ IDEA 2016.2.2
- Java version: 1.8.0_111
- OS name and version: OSX 10.10.5

- What are you trying to do?

I want to update the Go SDK after I change the project SDK in `Project Structure` panel.

- What would you expect to happen?

The External Libraries tree can be updated immediately after I change project SDK.

- What happens?

It always displays the same Go SDK no matter how I change the project SDK.

I run the plugin from source using Gradle.

screenshot

Additionally, I think `go-lang-idea-plugin` and `intellij-erlang` have a similar internal design, but I can easily update the External Libraries tree immediately after I change project SDK with `intellij-erlang` even I comment all the configurations [here](https://github.com/ignatov/intellij-erlang/blob/master/plugin/resources/META-INF/plugin.xml#L44-L57) which confuses me a lot.

screenshot

Thanks in advance for any help or advice :)

zjhmale:
Fixes #2861.

When a Go project is created by the plugin, the project SDK is fixed in iml file e.g.
`<orderEntry type="jdk" jdkName="Go 1.3.3" jdkType="Go SDK" />`.

But if we can change this line to `<orderEntry type="inheritedJdk" />` in iml file, the SDK in `External Libraries` tree can be synchronized as we change project SDK in `Project Structure` panel. Got this trick from [intellij-erlang](https://github.com/ignatov/intellij-erlang/blob/master/src/org/intellij/erlang/rebar/importWizard/RebarProjectImportBuilder.java#L235) 😄 

screenshot

### DIFF
--- a/src/com/goide/GoModuleBuilder.java
+++ b/src/com/goide/GoModuleBuilder.java
@@ -21,11 +21,13 @@ import com.intellij.compiler.CompilerWorkspaceConfiguration;
 import com.intellij.ide.util.projectWizard.JavaModuleBuilder;
 import com.intellij.ide.util.projectWizard.ModuleBuilderListener;
 import com.intellij.ide.util.projectWizard.SourcePathsBuilder;
+import com.intellij.openapi.application.ApplicationManager;
 import com.intellij.openapi.module.Module;
 import com.intellij.openapi.module.ModuleType;
 import com.intellij.openapi.options.ConfigurationException;
 import com.intellij.openapi.projectRoots.SdkTypeId;
 import com.intellij.openapi.roots.ModifiableRootModel;
+import com.intellij.openapi.roots.ModuleRootManager;
 import com.intellij.openapi.util.Pair;
 import com.intellij.util.containers.ContainerUtil;
 import org.jetbrains.annotations.NotNull;
@@ -58,5 +60,8 @@ public class GoModuleBuilder extends JavaModuleBuilder implements SourcePathsBui
   @Override
   public void moduleCreated(@NotNull Module module) {
     CompilerWorkspaceConfiguration.getInstance(module.getProject()).CLEAR_OUTPUT_DIRECTORY = false;
+    ModifiableRootModel rootModel = ModuleRootManager.getInstance(module).getModifiableModel();
+    rootModel.inheritSdk();
+    ApplicationManager.getApplication().runWriteAction(rootModel::commit);
   }
 }


ignatov:
Thanks for your PR

zjhmale:
@ignatov Thanks for you great repo 😄

src/com/goide/GoModuleBuilder.java
```diff
@@ -58,5 +60,8 @@ public boolean isSuitableSdkType(SdkTypeId sdkType) {
   @Override
   public void moduleCreated(@NotNull Module module) {
     CompilerWorkspaceConfiguration.getInstance(module.getProject()).CLEAR_OUTPUT_DIRECTORY = false;
+    ModifiableRootModel rootModel = ModuleRootManager.getInstance(module).getModifiableModel();
+    rootModel.inheritSdk();
+    ApplicationManager.getApplication().runWriteAction(rootModel::commit);
```
zolotov: Changing model of module while handling listener's event changing model of modules looks very suspisious I'd rather avoid this.
Setting inheritSdk happens in com.intellij.ide.util.projectWizard.JavaModuleBuilder#setupRootModel, it's better to fingure out why it doesn't work as expected.

zolotov:
Also the question is how you created a module in the first place?

zjhmale:
@zolotov I created a module with choosing a concrete SDK version, I can not find `select Project inherit SDK` option here 😅 

image

src/com/goide/GoModuleBuilder.java
zjhmale: Yeah, it seems a little hack here, `com.intellij.ide.util.projectWizard.JavaModuleBuilder#setupRootModel` need `myJdk` to be null in order to inherit SDK, but without a concrete SDK, a module can not be created.

zolotov: > but without a concrete SDK, a module can not be created.

Then this should be fixed. It is UI problem, not the problem of module builders

zolotov:
It's not a good way to fix a UI problem. Please rethink the PR

PR #2892, opened 2017-05-24
Unresolved struct field inspection

wbars:


src/com/goide/inspections/unresolved/GoAddStructFieldFix.java
```diff
@@ -0,0 +1,65 @@
...
+  public String getText() {
+    return QUICK_FIX_NAME;
+  }
+
+  @Override
+  public void invoke(@NotNull Project project, @NotNull PsiFile file, @NotNull PsiElement startElement, @NotNull PsiElement endElement) {
+    GoStructType structType = ObjectUtils.tryCast(startElement, GoStructType.class);
+    if (structType == null) return;
+    List<GoFieldDeclaration> declarations = structType.getFieldDeclarationList();
+    PsiElement anchor = !declarations.isEmpty() ? ContainerUtil.getLastItem(declarations) : structType.getLbrace();
```
zolotov: = ContainerUtil.getLastItem(declarations, structType.getLbrace)

src/com/goide/inspections/unresolved/GoAddStructFieldFix.java
```diff
@@ -0,0 +1,65 @@
...
+    if (anchor != null) structType.addAfter(GoElementFactory.createFieldDeclaration(project, myFieldText, myTypeText), anchor);
```
zolotov: When can it happen? Can you test this?

src/com/goide/inspections/unresolved/GoUnresolvedReferenceInspection.java
```diff
@@ -74,7 +76,12 @@ public void visitReferenceExpression(@NotNull GoReferenceExpression o) {
         }
         else if (reference.resolve() == null) {
           LocalQuickFix[] fixes = LocalQuickFix.EMPTY_ARRAY;
-          if (isProhibited(o, qualifier)) {
+          GoType type = qualifier != null ? qualifier.getGoType(null) : null;
+          GoStructType structType = type != null ? ObjectUtils.tryCast(type.getUnderlyingType(), GoStructType.class) : null;
+          if (!"_".equals(reference.getCanonicalText()) && structType != null) {
+            fixes = new LocalQuickFix[]{new GoAddStructFieldFix(reference.getCanonicalText(), getTypeName(o), structType)};
```
zolotov: Usually, psi element that is passed to QuickFix is an inspection holder. In this case you pass there the element that should be changed, not the unresolved element. It's bad, because the struct type or unresolved reference could completely changed on the moment of quickfix invocation and should not be modified,

